PR #350, opened 2020-09-29
Improve x86 bitmap() on 16 bits integers

jfalcou:
Using the code by @DenisYaroshevskiy


DenisYaroshevskiy:
Do you want to maybe write this as `convert`? This is effectively what you do.

jfalcou:
> Do you want to maybe write this as `convert`? This is effectively what you do.
> And fix the convert.

convert do a lot fo stuff and bitmap is an "internal function", so I think we should keep the dependency flow in a single direction.

DenisYaroshevskiy (approved):
LGTM

PR #379, opened 2020-10-22
Fix #354 - expected_cardinal should be restricted to native_abi

### LINKED ISSUE #354: [BUG]`expected_cardinal` should be restricted to `native_abi`
It's not clear what `exepcted_cardinal` is for `aggregated_abi` or `emulated_abi`.

jfalcou:


DenisYaroshevskiy:
Looks good.

A few questions:
1) Don't you already have a `is_power_of_2` function?
2) should the `first_true` example be updated with `regular_abi`
3) shouldn't `regular_abi` exclude emulated one as well?

jfalcou:
> Looks good.
:tada: 

> 1. Don't you already have a `is_power_of_2` function?
Yes but I am wary to bring this include dependency for such a small test

> 2. should the `first_true` example be updated with `regular_abi`

> 3. shouldn't `regular_abi` exclude emulated one as well?
No, as emulated_abi is able to return you a proper meaningful expected cardinal by design.
I suspect emumated_ will later become emulated_<NBits> to accommodate for silly arch like Blue Gene Q but
it always can provide the info expected_cardinal requires.

I pondered calling it native_abi but the fact it's native+emulated, I shot for another term.

DenisYaroshevskiy:
LGTM. Do you want me to merge?

Issue #356, opened 2020-10-03
add docs: example extending eve with custom functions for non-native register size: reverse


DenisYaroshevskiy:
issue to have something for pull request.

DenisYaroshevskiy:
I still didn't write text

DenisYaroshevskiy:
Won't fix, reverse is too complex.

Issue #348, opened 2020-09-27
[BUG] Conversion from short to char to on avx2 (and probably avx) generates too much code


DenisYaroshevskiy:
C++ code:
```
auto test(eve::logical<eve::wide<T>> x) {
  return eve::convert(x, eve::as_<char>{});
}
```

ASM:
```
...
	vextracti128	xmm1, ymm0, 1
	vpinsrw	xmm1, xmm1, word ptr [rsp + 28], 6
	vpinsrw	xmm1, xmm1, word ptr [rsp + 30], 7
	vinserti128	ymm0, ymm0, xmm1, 1
	vpxor	xmm1, xmm1, xmm1
	vpcmpeqw	ymm0, ymm0, ymm1
	vpcmpeqd	ymm1, ymm1, ymm1
	vpxor	ymm0, ymm0, ymm1
	vextracti128	xmm1, ymm0, 1
	vpacksswb	xmm0, xmm0, xmm1
	vpand	xmm0, xmm0, xmmword ptr [rip + .LCPI0_0]
..
```

I believe you can do entire conversion on avx like this:

```
    __m128i top = _mm256_extractf128_si256(x, 1);
    __m128i bottom = _mm256_castsi256_si128(x);
    __m128i packed = _mm_packs_epi16(bottom, top);
```

Haven't  actually tested but a similar technic is used here: https://github.com/jfalcou/eve/issues/347 and I tested that.

jfalcou:
Yes, this is a repeatable pattern of unpack+conversion across all integers and architectures. 
This needs to be in, but I wish we can fix it when the swizzles are there so we can have a generic, architecture agnostic implementation

DenisYaroshevskiy:
How are swizzles going to help?
If you fix it, you can use it in `bitmap` fix.

jfalcou:
Superceded by #534

PR #381, opened 2020-10-24
Issue #356: reverse example: writing a test

DenisYaroshevskiy:
OK - so this is a test for reverse example - which is suppose to illustrate how to work with non-native ABI sizes.

The test successfully reverses everything on x86 from sse2 to AVX2.

Please have a look - see how much does it look like example you are looking for

I have questions:
1) I don't believe I can compile with `EVE_NO_SIMD` in a meaningful way since `current_api` is still going to be SSE2, correct?
https://github.com/jfalcou/eve/blob/e749a9ce6f029d362825ff16961e7dc59ff190cf/include/eve/arch/spec.hpp#L27

2) I don't know how to correctly disable this test for anything but the ones I want. Like - avx512 is still going to try to compile my test with current macro?

3) How do I include parts of the test in the doc and not the entire code? There is a bit of boiler plate that is probably not necessary in the doc.

### DIFF
--- a/test/doc/developer/reverse.cpp
+++ b/test/doc/developer/reverse.cpp
@@ -1,233 +1,56 @@
+#include "test.hpp"
+#include "reverse.hpp"
 
-#include <eve/wide.hpp>
-#include <eve/function/bit_cast.hpp>
-#include <eve/function/store.hpp>
-
-#include <algorithm>
-#include <bit>
-#include <concepts>
-#include <cstdint>
-
-#include <type_traits>
-
-#if defined(EVE_HW_X86) && !defined(SPY_SIMD_IS_X86_AVX512)
-
-namespace eve_extra {
-namespace _reverse {
-
-template <typename T, typename N, eve::x86_abi ABI>
-eve::wide<T, N, ABI> actual_reverse(eve::wide<T, N, ABI> x) requires(sizeof(T) == 8)
+TTS_CASE("Check for reverse behavior")
 {
-       if constexpr(std::same_as<ABI, eve::x86_128_>) return {_mm_shuffle_epi32(x.storage(), _MM_SHUFFLE(1, 0, 3, 2))};
-  else if constexpr(std::same_as<ABI, eve::x86_256_>) return {_mm256_permutevar8x32_epi32(x.storage(), _MM_SHUFFLE(3, 2, 1, 0))};
+  #if defined(EVE_HW_X86) && !defined(SPY_SIMD_IS_X86_AVX512)
+
+  #endif
 }
 
-template <typename N>
-constexpr std::uint32_t mm_shuffle() {
-  if constexpr(N() == 4) return _MM_SHUFFLE(0, 1, 2, 3);
-  if constexpr(N() == 2) return _MM_SHUFFLE(3, 2, 0, 1);
-}
+// template<typename T, typename Test>
+// void for_all_test_pack_sizes(Test test)
+// {
+//   constexpr int number_of_packs = std::countr_zero(256u >> sizeof(T)) + 2;
 
-template <typename T, typename N, eve::x86_abi ABI>
-eve::wide<T, N, ABI> actual_reverse(eve::wide<T, N, ABI> x) requires(sizeof(T) == 4)
-{
-  auto raw = x.storage();
+//   [&]<int... idx>(std::integer_sequence<int, idx...>)
+//   {
+//     (test(eve::wide<T, eve::fixed<static_cast<size_t>(1 << idx)>> {}), ...);
+//   } (std::make_integer_sequence<int, number_of_packs>());
+// }
 
-       if constexpr(std::same_as<ABI, eve::x86_128_>) return {_mm_shuffle_epi32(raw, mm_shuffle<N>())};
-  else if constexpr(std::same_as<ABI, eve::x86_256_>)
-  {
-    using mask_type = eve::wide<int, eve::fixed<8>>;
-    auto mask = mask_type([](int i, int) { return 7 - i; }).storage();
+// template <typename Test>
+// void for_all_test_packs(Test test)
+// {
+//   [&]<typename...T>(T...) { (for_all_test_pack_sizes<T>(test), ...); }
+//   (std::int8_t{},  std::uint8_t{},
+//    std::int16_t{}, std::uint16_t{},
+//    std::int32_t{}, std::uint32_t{},
+//    std::int64_t{}, std::uint64_t{},
+//    float{},        double{});
+// }
 
-    return {_mm256_permutevar8x32_epi32(raw, mask)};
-  }
-}
+// int main()
+// {
+//   auto test = []<typename wide>(wide) {
+//     wide input([](int i, int) { return i ; });
+//     input.set(0, std::numeric_limits<typename wide::value_type>::min());
+//     input.set(wide::static_size - 1, std::numeric_limits<typename wide::value_type>::max());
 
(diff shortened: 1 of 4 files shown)


jfalcou:
> 1. I don't believe I can compile with `EVE_NO_SIMD` in a meaningful way since `current_api` is still going to be SSE2, correct?

It's indeed an oversight. SPY provides a undefined_simd_ marker so I guess I need to use this whenever the macro is defined. I'm fixing that in turbo mode today

> 2\. I don't know how to correctly disable this test for anything but the ones I want. Like - avx512 is still going to try to compile my test with current macro?

One way to do so is to wrap your test around a `#if !defined(SPY_SIMD_IS_X86_AVX512)`.

> 3\. How do I include parts of the test in the doc and not the entire code?

Currently, it's complicated. I am sending a message to Morgan from markdeep to see if we could have some support for that.
Meanwhile, I guess you can put your code in a .hpp then include it in the cpp. We'll then write somethign to wrap loose .hpp into markdeep.

I'm going over the code and filling small comments manwhile.

test/doc/developer/reverse.cpp
```diff
@@ -0,0 +1,194 @@
...
+
+#include <type_traits>
+
+#if defined(EVE_HW_X86)
+
+namespace eve_extra {
+namespace _reverse {
+
+template <typename T, typename N, typename ABI>
+eve::wide<T, N, ABI> actual_reverse(eve::wide<T, N, ABI> x) requires(sizeof(T) == 8)
```
jfalcou, line 18: traditionally, we prefer when tests like this `requires` are in fact `if constexpr` inside.

test/doc/developer/reverse.cpp
```diff
@@ -0,0 +1,194 @@
...
+#include <concepts>
+
+#include <type_traits>
+
+#if defined(EVE_HW_X86)
+
+namespace eve_extra {
+namespace _reverse {
+
+template <typename T, typename N, typename ABI>
```
jfalcou, line 17: If you know your ABI is x86 related, you should use the `x86_abi` concept instead of `typename`

test/doc/developer/reverse.cpp
```diff
@@ -0,0 +1,194 @@
...
+    [](int i, int) { return i >= N() ? i : N() - 1 - i; }
+  ).storage();
+}
+
+template <typename T, typename N, typename ABI>
+eve::wide<T, N, ABI> actual_reverse(eve::wide<T, N, ABI> x) requires(sizeof(T) == 1)
+{
+  auto raw = x.storage();
+
+       if constexpr (eve::current_api <= eve::sse3)
```
jfalcou, line 66: [style] - spurious blank

test/doc/developer/reverse.cpp
```diff
@@ -0,0 +1,194 @@
...
+  {
+    static constexpr std::size_t size = static_cast<std::size_t>(N());
+
+    alignas(sizeof(T) * size) std::array<T, size> buffer;
```
jfalcou, line 70: I think it'll be better to just use the lambda constructor to reconstruct a wide in reverse instead of going through memory.

`wide<T> result( [&x](int i, int c) { return x[c-1-i]; }`

?

test/doc/developer/reverse.cpp
```diff
@@ -0,0 +1,194 @@
...
+     raw = _mm_shuffle_epi32(raw, _MM_SHUFFLE(1, 0, 3, 2));
+
+     return {raw};
+  }
+  else if constexpr (std::same_as<ABI, eve::x86_128_>) return {_mm_shuffle_epi8(raw, reverse_8_shorts_mask<N>())};
+  else if constexpr (std::same_as<ABI, eve::x86_256_>) return {reverse_256_using_128_mask(raw, reverse_8_shorts_mask<N>())};
+}
+
+template <typename Wide>
+constexpr bool treat_like_aggregated() {
```
jfalcou, line 109: what are you testing here that `has_aggregated_abi_v` doesn't do ?

test/doc/developer/reverse.cpp
```diff
@@ -0,0 +1,194 @@
...
+    as_ints = eve_extra::reverse(as_ints);
+    return eve::bit_cast(as_ints, eve::as_<Wide>{});
+  }
+  else return _reverse::actual_reverse(x);
+}
+
+}  // namespace eve_extra
+
+
+template<typename T, typename Test>
```
jfalcou, line 152: Would you like if I pass over this section so you can use TTS instead of manual tests ?

jfalcou (changes requested):

test/doc/developer/reverse.cpp
DenisYaroshevskiy: I thought about it - but this makes the function 15 miles long. If you want me to - I can do it.

test/doc/developer/reverse.cpp
DenisYaroshevskiy: OK

test/doc/developer/reverse.cpp
DenisYaroshevskiy: OK

test/doc/developer/reverse.cpp
DenisYaroshevskiy: Ok - can do. Will check the codegen though - I suspect we are going to get much bigger binary code. Will have to see.

test/doc/developer/reverse.cpp
DenisYaroshevskiy: `x86_256_` on AVX does not have an aggregated abi, but we still need to treat it like one.

test/doc/developer/reverse.cpp
DenisYaroshevskiy: I would appreciate it.

test/doc/developer/reverse.cpp
jfalcou: Normally no has it usually use the mmset like intrinsic.

test/doc/developer/reverse.cpp
jfalcou: Oh I see. I think we never had this.
I think we can add a splittable concept to handle that.

test/doc/developer/reverse.cpp
jfalcou: As we discussed on Slack, better keep those separated for the clarity of the tutorial.
We'll later do a "you want to write something FOR EVE"  tutorial.

DenisYaroshevskiy:
As we discussed on slack, the reverse for chars on sse should happen via shorts.

However at the moment `eve::convert(wide<uint16_t>, eve::as_<uint8_t>{})` generates horrandous code, see #348 

I can fix that but I am confused about where that should happen, can you point me please?

DenisYaroshevskiy:
Updated the pr.

Fixing conversions in general is a bit more tricky, so I have a separate implementation here.

DenisYaroshevskiy:
Fixed the test. However, I don't like that we only test the native size, when the whole point of the test is to show all emulation.

+ the types coverage is insufficient.

jfalcou:
> Fixed the test. However, I don't like that we only test the native size, when the whole point of the test is to show all emulation.
> 
> * the types coverage is insufficient.

I adapted it as a regular tests, i.e variations of size is handled by the CI.
We can change to a model where developer tests are more manually driven, not a big deal.

DenisYaroshevskiy:
@jfalcou - can we merge this?

jfalcou (approved):

PR #386, opened 2020-11-12
top_bits optimization

jfalcou:
top_bits allow for proper x86 handling of movmask like operations

### DIFF
--- a/include/eve/detail/top_bits.hpp
+++ b/include/eve/detail/top_bits.hpp
@@ -11,20 +11,21 @@
 //==================================================================================================
 #pragma once
 
+#include <eve/as.hpp>
 #include <eve/detail/has_abi.hpp>
-#include <eve/traits/element_type.hpp>
+#include <eve/detail/meta.hpp>
 #include <eve/traits/as_wide.hpp>
 #include <eve/traits/cardinal.hpp>
-#include <eve/function/bit_cast.hpp>
-#include <eve/as.hpp>
-#include <optional>
+#include <eve/traits/element_type.hpp>
+
 #include <array>
 #include <bit>
+#include <optional>
 
 namespace eve::detail
 {
   //================================================================================================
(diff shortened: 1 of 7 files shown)


include/eve/detail/top_bits.hpp
```diff
@@ -0,0 +1,273 @@
...
+      else
+      {
+        if constexpr( std::same_as<typename Pack::abi_type,x86_128_>)
+        {
+          raw = _mm_movemask_epi8((__m128i)(x.storage()));
+        }
+        else if constexpr( std::same_as<typename Pack::abi_type,x86_256_>)
+        {
+          raw = _mm256_movemask_epi8((__m256i)(x.storage()));
+        }
```
DenisYaroshevskiy, line 93: I thought that for 4/8 bit numbers you gonna use different instructions.

DenisYaroshevskiy:
Suggestion for an interface:
we want to support ignore and not, right?

How about `get_top_bits` that can support ignore and not? Probably multiple level of not?

Not sure how that'd look like exactly though. Probably depends on how you want not to work.

Like `eve::first_true[eve::ignore_first_{3}](eve::not_(eve::not_(mask))`?

include/eve/detail/top_bits.hpp
```diff
@@ -0,0 +1,309 @@
...
+  template <typename Pack>
+  top_bits<Pack> combine_ignore(const top_bits<Pack>& x, const top_bits<Pack>& ignore) {
+    return x & ignore;
+  }
+*/
+
+  //------------------------------------------------------------------------------------------------
+  // Find the index of the first true bit if any
+  template <typename Pack>
+  std::optional<std::size_t> first_true(const top_bits<Pack>& x) noexcept
```
DenisYaroshevskiy, line 245: Maybe return a `std::ptrdiff_t`? You use signed everywhere, can be here too.

PR #406, opened 2020-11-29
Make load & store more flexible

jfalcou:
- load is now a callable supporting conditional expressions.
 - conditional expressions has been refined by separating if_ from ignore_* and making ignore supports alternative
 - extensive tests for load


DenisYaroshevskiy:
You don't do intel masked load magic yet, right?

include/eve/module/core/function/generic/load.hpp
```diff
@@ -0,0 +1,154 @@
...
+  template<relative_conditional_expr C, scalar_pointer Ptr>
+  EVE_FORCEINLINE auto load_(EVE_SUPPORTS(cpu_), C const &cond, Ptr ptr) noexcept
+                  -> as_wide_t<std::remove_cvref_t<decltype(*std::declval<Ptr>())>>
+  {
+    using e_t = std::remove_cvref_t<decltype(*ptr)>;
+    using r_t = as_wide_t< e_t >;
+
+    if constexpr( !std::is_pointer_v<Ptr> )
+    {
+      return eve::load[cond](ptr.get());
```
DenisYaroshevskiy, line 55: We will replace this `load[cond]` with `load_unsafe` when we have it, right?

jfalcou: Yes

jfalcou:
> You don't do intel masked load magic yet, right?

Yes this is being done atm.

include/eve/module/core/function/generic/load.hpp
```diff
@@ -0,0 +1,154 @@
...
+      }
+      else
+      {
+        auto offset = cond.offset( as_<r_t>{} );
+
+        if constexpr(C::has_alternative)
+        {
+          [[maybe_unused]] r_t that(cond.alternative);
+          auto* dst   = (e_t*)(&that.storage());
+          std::memcpy( dst + offset, ptr + offset, cond.count( as_<r_t>{} ) );
```
DenisYaroshevskiy, line 77: This is broken for smaller wides I believe.

DenisYaroshevskiy (changes requested):

include/eve/module/core/function/generic/load.hpp
DenisYaroshevskiy: @jfalcou says I'm wrong. If there is a test for it, I'm good.

include/eve/conditional.hpp
```diff
@@ -315,14 +361,28 @@ namespace eve
     static constexpr bool is_inverted     = false;
     static constexpr bool is_complete     = false;
 
-    constexpr keep_between_(std::ptrdiff_t b, std::ptrdiff_t e) noexcept : begin_(b), end_(e) {}
+    constexpr keep_between_(std::ptrdiff_t b, std::ptrdiff_t e) noexcept
+    : begin_(std::min(b,e)), end_(std::max(b,e))
```
DenisYaroshevskiy, line 365: Why do you need min here?
These couple of instructions can be an issue when the whole thing takes 5 ns

include/eve/conditional.hpp
```diff
@@ -315,14 +361,28 @@ namespace eve
...
 
+    template<typename T> EVE_FORCEINLINE std::ptrdiff_t offset(eve::as_<T> const&) const
+    {
+      return begin_;
+    }
+
+    template<typename T> EVE_FORCEINLINE auto count(eve::as_<T> const&) const
+    {
+      return sizeof(element_type_t<T>) * (end_ - begin_ + 1);
+    }
```
DenisYaroshevskiy, line 384: +1? Why do you have a +1? Ignoring all elements is fine? What I can't keep_between 0 elements?

include/eve/conditional.hpp
```diff
@@ -349,11 +409,22 @@ namespace eve
...
+
+    template<typename T> EVE_FORCEINLINE std::ptrdiff_t offset(eve::as_<T> const&) const
+    {
+      return first_count_;
+    }
+
+    template<typename T> EVE_FORCEINLINE auto count(eve::as_<T> const&) const
+    {
+      return sizeof(element_type_t<T>) * (cardinal_v<T> - last_count_ - first_count_);
```
DenisYaroshevskiy, line 427: See - you don't have a +1 here

include/eve/detail/function/simd/x86/load.hpp
```diff
@@ -11,125 +11,110 @@
...
-      }
-      else if constexpr( std::is_same_v<T, float> )
-      {
-        return _mm_loadu_ps(p);
-      }
-      else if constexpr( std::is_integral_v<T> )
-      {
-        return _mm_loadu_si128((__m128i *)p);
-      }
+      return load(eve::as_<wide<T, N>>{}, ABI{}, p.get());
```
DenisYaroshevskiy, line 30: But this will generate an unaligned load, no? Since this is `x86` it's not OK

include/eve/detail/function/simd/x86/load.hpp
```diff
@@ -11,125 +11,110 @@
...
-        return _mm_load_ps(p.get());
+        return _mm_loadu_si128((__m128i *)p);
       }
-      else if constexpr( std::is_integral_v<T> )
+      else
       {
-        return _mm_load_si128((__m128i *)p.get());
+        typename wide<T, N>::storage_type that;
+        std::memcpy(&that, p, N::value * sizeof(T));
+        return that;
```
DenisYaroshevskiy, line 52: Do you really need this `else` in `x86`? I think this should give you a compiler error, which it will if you do nothing.

include/eve/detail/function/simd/x86/load.hpp
```diff
@@ -11,125 +11,110 @@
...
+                if constexpr( cat == category::float64x4 )  return _mm256_blendv_pd(a,b,mm);
+          else  if constexpr( cat == category::float32x8 )  return _mm256_blendv_ps(a,b,mm);
+          else  if constexpr( cat == category::float64x2 )  return _mm_blendv_pd(a,b,mm);
+          else  if constexpr( cat == category::float32x4 )  return _mm_blendv_ps(a,b,mm);
+          else  if constexpr( sizeof(a) == 32)              return _mm256_blendv_epi8(a,b,mm);
+          else  if constexpr( sizeof(a) == 16)              return _mm_blendv_epi8(a,b,mm);
+        };
+
+        r_t alt(cond.alternative);
+        that = blender(alt,that,mask);
```
DenisYaroshevskiy, line 110: Why not just call `if_else`?  Trying to save on compile time?

include/eve/module/core/function/generic/load.hpp
```diff
@@ -0,0 +1,146 @@
...
+          else                              return r_t{};
+        }
+      }
+      else
+      {
+        auto offset = cond.offset( as_<r_t>{} );
+
+        if constexpr(C::has_alternative)
+        {
+          [[maybe_unused]] r_t that(cond.alternative);
```
DenisYaroshevskiy, line 67: I am really confused anout these `maybe_unused` It's clearly used

include/eve/module/core/function/generic/load.hpp
```diff
@@ -0,0 +1,146 @@
...
+          std::memcpy( dst + offset, ptr + offset, cond.count( as_<r_t>{} ) );
+          return that;
+        }
+        else
+        {
+          [[maybe_unused]] r_t that;
+          auto* dst   = (e_t*)(&that.storage());
+          std::memcpy( dst + offset, ptr + offset, cond.count( as_<r_t>{} ) );
+          return that;
+        }
```
DenisYaroshevskiy, line 125: Aren't your `if/else` branches exactly the same? Why do you need them?

include/eve/detail/function/simd/x86/load.hpp
jfalcou: Oh this is an error yes, was too sloppy with my copy pasta. Will fix

include/eve/detail/function/simd/x86/load.hpp
jfalcou: this was meant to be the case for wide smaller than the expected cardinal. But now that I look at it, if you have a small real, it calls loadu. This maybe need to be generalized to other cases no or can I loadu past my data ?

include/eve/detail/function/simd/x86/load.hpp
jfalcou: if_else is in core an I want to minimize the internals details dependency on outside function.
It also saves compiles time as we don't need the full power of if_else special cases handling.

include/eve/module/core/function/generic/load.hpp
jfalcou: this is indeed unneeded here.

include/eve/module/core/function/generic/load.hpp
jfalcou: I'll take care of forcing the tests but it works for N == 1 so ...

include/eve/module/core/function/generic/load.hpp
jfalcou: the first one fill the return value with the alternative default value befor loading. 
However, this can be factorized into simpler code.

include/eve/conditional.hpp
jfalcou: I wanted to support the case where begin is greater than end.
I may just assert if this error occurs?

include/eve/conditional.hpp
jfalcou: Indeed, this is an error.

include/eve/conditional.hpp
jfalcou: OK

include/eve/conditional.hpp
```diff
@@ -126,11 +130,23 @@ namespace eve
...
 
+    template<typename T> EVE_FORCEINLINE std::ptrdiff_t offset(eve::as_<T> const&) const
+    {
+      return 0;
+    }
+
+    template<typename T> EVE_FORCEINLINE auto count(eve::as_<T> const&) const
+    {
+      return 0ULL;
+    }
```
DenisYaroshevskiy, line 148: Why is this unsigned but offset is signed?

include/eve/detail/function/simd/x86/load.hpp
```diff
@@ -11,125 +11,129 @@
...
-      if constexpr( std::is_same_v<T, double> )
+      if constexpr( isfull256 )
       {
-        return _mm_loadu_pd(p);
+              if constexpr( cat == category::float64x4 )  return _mm256_load_pd(p.get());
+        else  if constexpr( cat == category::float32x8 )  return _mm256_load_ps(p.get());
+        else return _mm256_load_si256((__m256i *)p.get());
       }
-      else if constexpr( std::is_same_v<T, float> )
+      else  if constexpr( isfull128 )
```
DenisYaroshevskiy, line 40: Why 2 spaces?

include/eve/detail/function/simd/x86/load.hpp
```diff
@@ -11,125 +11,129 @@
...
+  template<scalar_pointer Ptr, typename T, typename N, x86_abi ABI>
+  EVE_FORCEINLINE auto load(eve::as_<wide<T, N>> const &, ABI const &, Ptr p)
+  requires( std::same_as<T, std::remove_cvref_t<decltype(*p)>> )
   {
-    if constexpr( N::value * sizeof(T) == x86_128_::bytes )
+    constexpr auto cat = categorize<wide<T, N>>();
+    constexpr bool isfull256 = N::value*sizeof(T) == x86_256_::bytes;
+    constexpr bool isfull128 = N::value*sizeof(T) == x86_128_::bytes;
+
+    if constexpr( !std::is_pointer_v<Ptr> )
```
DenisYaroshevskiy, line 32: I think this should probably be a proper check for `aligned_ptr<, sizeof(T) * N>`
Otherwise this will break.

Also - I believe that this will eat a `unique_ptr` or some other nonsense.

include/eve/detail/function/simd/x86/load.hpp
DenisYaroshevskiy: Ok. You could put it in some common place though. But I don't care much.

include/eve/module/core/function/generic/load.hpp
```diff
@@ -0,0 +1,146 @@
...
+  }
+
+  template<relative_conditional_expr C, scalar_pointer Ptr>
+  EVE_FORCEINLINE auto load_(EVE_SUPPORTS(cpu_), C const &cond, Ptr ptr) noexcept
+                  -> as_wide_t<std::remove_cvref_t<decltype(*std::declval<Ptr>())>>
+  {
+    using e_t = std::remove_cvref_t<decltype(*ptr)>;
+    using r_t = as_wide_t< e_t >;
+
+    if constexpr( !std::is_pointer_v<Ptr> )
```
DenisYaroshevskiy, line 45: Again - this is not a very good check.

include/eve/module/core/function/generic/load.hpp
jfalcou: Removed

include/eve/conditional.hpp
jfalcou: this is a size and is usually computed from sizeof in other case, hence the 0ULL.
Offset is maybe a good candidate to become a size_t too.

include/eve/detail/function/simd/x86/load.hpp
jfalcou: Fixed

include/eve/conditional.hpp
DenisYaroshevskiy: Not sure I agree, but it doesn't really matter.

include/eve/detail/function/simd/x86/load.hpp
DenisYaroshevskiy: Doesn't seem like it

include/eve/detail/function/simd/x86/load.hpp
```diff
@@ -11,125 +11,129 @@
...
-      if constexpr( std::is_same_v<T, double> )
+      if constexpr( isfull256 )
       {
-        return _mm_load_pd(p.get());
+              if constexpr( cat == category::float64x4 )  return _mm256_loadu_pd(p);
+        else  if constexpr( cat == category::float32x8 )  return _mm256_loadu_ps(p);
+        else return _mm256_loadu_si256((__m256i *)p);
       }
-      else if constexpr( std::is_same_v<T, float> )
+      else  if constexpr( isfull128 )
```
DenisYaroshevskiy, line 61: 2 spaces

include/eve/detail/function/simd/x86/load.hpp
```diff
@@ -11,125 +11,129 @@
...
+      r_t that;
+      auto mask = cond.mask( as_<r_t>{} ).bits();
+
+      constexpr auto cat = categorize<r_t>();
+
+            if constexpr( cat == category::float64x4 )  that = _mm256_maskload_pd (ptr,mask);
+      else  if constexpr( cat == category::float64x2 )  that = _mm_maskload_pd    (ptr,mask);
+      else  if constexpr( cat == category::float32x8 )  that = _mm256_maskload_ps (ptr,mask);
+      else  if constexpr( cat == category::float32x4 )  that = _mm_maskload_ps    (ptr,mask);
+      else  if constexpr( current_api >= avx2 )
```
DenisYaroshevskiy, line 97: Again 2 spaces? Do you want them?

test/unit/api/wide/load/arithmetic/load.hpp
```diff
@@ -26,83 +26,236 @@ auto data_block()
...
+  auto const* ref_const_ptr = &data[idx];
+
+  TTS_WHEN("For some given relative masks")
+  {
+    // Conditional selectors
+    auto il = eve::ignore_last(EVE_CARDINAL/4);
+    auto i1 = eve::ignore_first(EVE_CARDINAL/4);
+    auto kf = eve::keep_first(EVE_CARDINAL/4);
+    auto kl = eve::keep_last(EVE_CARDINAL/4);
+    auto kb = eve::keep_between ( std::min(0L,EVE_CARDINAL)
```
DenisYaroshevskiy, line 119: This min doesn't make sense

test/unit/api/wide/load/arithmetic/load.hpp
```diff
@@ -26,83 +26,236 @@ auto data_block()
...
+      if constexpr(EVE_CARDINAL == eve::expected_cardinal_v<EVE_VALUE>)
+      {
+      TTS_EQUAL(eve::load[eve::ignore_none](ref_ptr)              , full_ref        );
+      TTS_EQUAL((eve::load[il](ref_ptr) & ml.mask() )             , ignore_last_ref );
+      TTS_EQUAL((eve::load[i1](ref_ptr) & m1.mask() )             , ignore_first_ref);
+      TTS_EQUAL((eve::load[kf](ref_ptr) & mkf.mask())             , keep_first_ref  );
+      TTS_EQUAL((eve::load[kl](ref_ptr) & mkl.mask())             , keep_last_ref   );
+      TTS_EQUAL((eve::load[kb](ref_ptr) & mkb.mask())             , keep_between_ref);
+      TTS_EQUAL((eve::load[ie](ref_ptr) & mie.mask())             , ignore_ext_ref  );
+      TTS_EQUAL((eve::load[eve::ignore_all](ref_ptr) & mia.mask()), T{0}            );
```
DenisYaroshevskiy, line 156: That's not guaranteed to be 0 probably, right?

I would also suggest to load from really less data then a register.

Finally - I am not sure it's OK to compare for full equality - again - we don't populate garbage values

test/unit/api/wide/load/arithmetic/load.hpp
```diff
@@ -26,83 +26,236 @@ auto data_block()
...
+      TTS_EQUAL((eve::load(eve::as_aligned<algt>(ref_const_ptr), eve::lane<EVE_CARDINAL>)) , ref);
+    }
+  }
+}
+
+TTS_CASE_TPL("Check conditional load from unaligned pointer for wide", EVE_TYPE)
+{
+  constexpr std::ptrdiff_t algt = eve::alignment_v<T>;
+  auto [data,idx] = data_block<EVE_VALUE, eve::fixed<EVE_CARDINAL>>();
+  auto* ref_ptr             = &data[idx];
```
DenisYaroshevskiy, line 109: Is this pointer really unaligned? Otherwise you might mess up unaligned for aligned instruction and not notice

test/unit/api/wide/load/arithmetic/load.hpp
```diff
@@ -26,83 +26,236 @@ auto data_block()
...
+      TTS_EQUAL((eve::load[i1](ref_const_ptr, lanes) & m1.mask() )              , ignore_first_ref);
+      TTS_EQUAL((eve::load[kf](ref_const_ptr, lanes) & mkf.mask())              , keep_first_ref  );
+      TTS_EQUAL((eve::load[kl](ref_const_ptr, lanes) & mkl.mask())              , keep_last_ref   );
+      TTS_EQUAL((eve::load[kb](ref_const_ptr, lanes) & mkb.mask())              , keep_between_ref);
+      TTS_EQUAL((eve::load[ie](ref_const_ptr, lanes) & mie.mask())              , ignore_ext_ref  );
+      TTS_EQUAL((eve::load[eve::ignore_all](ref_const_ptr, lanes) & mia.mask()) , T{0}            );
+    }
+
+    // Aligned pointer
+    auto algt_ptr = eve::as_aligned<algt>(ref_ptr);
```
DenisYaroshevskiy, line 200: Yeah, doesn't seem like ref_ptr is unaligned. It should be

test/unit/api/wide/load/logical/load.hpp
```diff
@@ -37,84 +37,241 @@ auto data_block()
   return std::pair{ref,start};
 }
 
-TTS_CASE_TPL( "Check ctor from unaligned pointer for wide", EVE_TYPE )
+TTS_CASE_TPL("Check load for scalar", EVE_TYPE)
```
DenisYaroshevskiy, line 40: This is the same test, right? Maybe don't copypaste?

include/eve/detail/function/simd/x86/load.hpp
jfalcou: I thought of the `unique_ptr` situation. I can refine the concept to be T* or aligned_ptr<T,N>.
The concept makes the code far easier to work with and more compact.

test/unit/api/wide/load/arithmetic/load.hpp
jfalcou: will replace it with a proper sub-space of 0->cardinal

include/eve/module/core/function/generic/load.hpp
jfalcou: Except it doesn't work. I was trying to merge the constexpr if into a local lambda but I am forced to initialize if no alternative are present, thus making a useless pxor.

include/eve/detail/function/simd/x86/load.hpp
jfalcou: Fixed to specifically test for is_pointer + aligned_ptr

include/eve/concept/memory.hpp
```diff
@@ -0,0 +1,28 @@
...
+{
+  namespace detail
+  {
+    template<typename T> struct is_scalar_pointer : std::is_pointer<T> {};
+
+    template<typename T, std::size_t A>
+    struct is_scalar_pointer<aligned_ptr<T,A>> : std::true_type {};
+  }
+
+  template<typename T> concept scalar_pointer = detail::is_scalar_pointer<T>::value;
```
DenisYaroshevskiy, line 27: I thing this should be that concept. And it should also accept `simd value`

include/eve/detail/function/simd/x86/load.hpp
```diff
@@ -11,125 +11,67 @@
...
 namespace eve::detail
 {
   //================================================================================================
-  // 128 bits loads
+  // Regular loads
   //================================================================================================
-  template<real_scalar_value T, typename N>
-  EVE_FORCEINLINE auto load(eve::as_<wide<T, N>> const &, eve::x86_128_ const &, T const* p)
+  template<scalar_pointer Ptr, typename T, typename N, x86_abi ABI>
+  EVE_FORCEINLINE auto load(eve::as_<wide<T, N>> const &, ABI const &, Ptr p)
```
DenisYaroshevskiy, line 25: `as_<T, N, ABI>` ? Why didn't it fail tests?

jfalcou: the fact the ABI is constrainted to x86_abi make it that wide<T,N> is the proper ABI.
I can add it for clarity though

include/eve/arch/cardinals.hpp
```diff
@@ -0,0 +1,52 @@
...
+
+namespace eve
+{
+  template<std::ptrdiff_t Cardinal>
+  struct fixed : std::integral_constant<std::ptrdiff_t, Cardinal>
+  {
+    static constexpr bool is_pow2(std::ptrdiff_t v) { return !v || (!(v & (v - 1)) && v); }
+    static_assert(is_pow2(Cardinal), "Cardinal must be a power of 2");
+
+    static constexpr bool is_default = false;
```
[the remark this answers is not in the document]
jfalcou: It's an old piece of code from long before that I forgot to remove. I'll remove it

include/eve/arch/cardinals.hpp
```diff
@@ -0,0 +1,52 @@
...
+#pragma once
+
+#include <cstddef>
+
+namespace eve
+{
+  template<std::ptrdiff_t Cardinal>
+  struct fixed : std::integral_constant<std::ptrdiff_t, Cardinal>
+  {
+    static constexpr bool is_pow2(std::ptrdiff_t v) { return !v || (!(v & (v - 1)) && v); }
```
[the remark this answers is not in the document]
jfalcou: True

include/eve/detail/function/simd/x86/load.hpp
DenisYaroshevskiy: Ah - the test doesn't fail, because no-one ever specifies api themselves explicitly

DenisYaroshevskiy (approved):
LGTM

PR #411, opened 2020-12-03
Fix # 353 - Add abi_compatible_with concept

jfalcou:


test/unit/meta/abi_compatible_with.cpp
```diff
@@ -0,0 +1,60 @@
...
+#elif defined(SPY_SIMD_IS_ARM)
+TTS_CASE_TPL( "Check that wide<T,N,x86*> satisfies abi_compatible_with with any ARM ABI"
+            , TTS_NUMERIC_TYPES
+            )
+{
+  using eve::wide;
+  using eve::logical;
+
+  TTS_EXPECT((eve::abi_compatible_with<wide<T>          , eve::arm_64_ , eve::arm_128_> ));
+  TTS_EXPECT((eve::abi_compatible_with<logical<wide<T>> , eve::arm_64_ , eve::arm_128_> ));
```
DenisYaroshevskiy, line 58: Where are the false tests?

include/eve/concept/abi.hpp
```diff
@@ -0,0 +1,56 @@
...
+                                        >
+                              >
+              || ...
+              )
+            >
+    {};
+  }
+
+  template<typename Wide, typename... ABI>
+  concept abi_compatible_with = detail::is_abi_compatible<Wide,ABI...>::value;
```
DenisYaroshevskiy, line 55: This name doesn't quite make sence.
Why compatible? This checks that the cardinal is the expected one? That it's not smaller or larger?
`expected` `natural` `default`?

test/unit/meta/abi_compatible_with.cpp
jfalcou: Good point

include/eve/concept/abi.hpp
```diff
@@ -0,0 +1,56 @@
...
+#include <eve/traits/as_wide.hpp>
+#include <eve/forward.hpp>
+#include <concepts>
+
+namespace eve
+{
+  namespace detail
+  {
+    template<typename Wide, typename... ABI>
+    struct  is_abi_compatible
```
DenisYaroshevskiy, line 23: This should probably be renamed as well?

test/unit/meta/native_simd_for_abi.cpp
```diff
@@ -0,0 +1,99 @@
...
+#if defined(SPY_SIMD_IS_X86)
+TTS_CASE_TPL( "Check that wide<T,N,x86*> satisfies native_simd_for_abi with any X86 ABI"
+            , TTS_NUMERIC_TYPES
+            )
+{
+  using eve::wide;
+  using eve::logical;
+
+  TTS_EXPECT((eve::native_simd_for_abi<wide<T>          , eve::x86_128_ , eve::x86_256_> ));
+  TTS_EXPECT((eve::native_simd_for_abi<logical<wide<T>> , eve::x86_128_ , eve::x86_256_> ));
```
DenisYaroshevskiy, line 36: I would like to see also tests for bigger and smaller wide.

include/eve/concept/abi.hpp
jfalcou: Sure

test/unit/meta/native_simd_for_abi.cpp
jfalcou: Yes, also the test need to move from concept to arch as currently is is only tested once in scalar mode :/

DenisYaroshevskiy (approved):

DenisYaroshevskiy (approved):

Issue #412, opened 2020-12-05
[BUG] unaligned store for chars for __m128 on avx is busted


DenisYaroshevskiy:
```{asm}
store(eve::wide<signed char, eve::fixed<8l>, eve::x86_128_>, eve::aligned_ptr<signed char, 8ul>): 
    vpextrb    byte ptr [rdi], xmm0, 0
    vpextrb    byte ptr [rdi + 1], xmm0, 1
    vpextrb    byte ptr [rdi + 2], xmm0, 2
    vpextrb    byte ptr [rdi + 3], xmm0, 3
    vpextrb    byte ptr [rdi + 4], xmm0, 4
    vpextrb    byte ptr [rdi + 5], xmm0, 5
    vpextrb    byte ptr [rdi + 6], xmm0, 6
    vpextrb    byte ptr [rdi + 7], xmm0, 7
    ret
```

jfalcou:
Yup, sounds like we need a proper memcpy there instead of manually forcing scalar storage.

PR #413, opened 2020-12-05
Fix #393: basic asm testing functionality

### LINKED ISSUE #393: [BUG] Codegen testing
At the moment all of the testing is about correctness.
However, the many `if constexpr` branches can be not written correctly (and we saw a few examples like that).

We need codegen testing: smth a long the lines of: this function will produce the following instruction.
We discussed that one compiler should be enough and it's hard to do more than that, since the codegen will vary between the compilers.

DenisYaroshevskiy:
**_NOTE:_** I didn't write tests for python yet. Let's first talk more about what we want and then I'll write python tests.

## What is this?

Because the next generations of architecture is backwards compatible to the previous ones,
we often encounter bugs when we fallback to a weaker generation.
This results in performance penalties, sometimes drastic (especially if we fallback to scalar code).

Testing this in C++ is painful and would require and extra layer of super complicated meta-programming.
We decided to test the assembly generated by the compiler, with approval testing.

We are going to generate asm once, confirm that it's what's expected and commit it.
During future builds we will verify that it didn't change.

This is a first iteration of a framework to do it.

## Fundamental Issues

-  We have to do this with one compiler. I would not expect gcc, clang and for sure msvc to produce the same asm.
-  If you changed asm intentionally but have no way to build for one of the platforms, you are stuck and cannot merge. Is there some easy access to the build machine where I can copy the files? I could dump the mismatched files somewhere.
- This is a lot of asm to review if you add a new function. And if some big change happens, we will have a bad day. We need to carefully decide what we test and what we don't test.

## What does the error look like?

Let's say we mess up and change aligned load to unaligned for 4 floats (128) on avx2.
The output will be:

```
Failure!!!===========

Mismatch for load(eve::aligned_ptr<float const, 16ul>, eve::fixed<4l>): for avx2:
Expected:
    vmovaps    xmm0, xmmword ptr [rdi]
    ret

Actual:
    vmovups    xmm0, xmmword ptr [rdi]
    ret
```

## How is it implemented?

I call compiler directly, not abusing CMAKE for it. And I also don't have to init multiple configurations to just test them.

The compilation process is the same for both generation and validation:

* find all relevant sources in 'test/asm/sources'
* for all tested architectures, batch those sources (by default 40 in a batch, no idea about the correct number atm).
* do a unity build those batches in parallel. (You can specify -j for the number of processes, by default - 4). This is important even now when we don't have that much code, since for separate architectures, we have to run separate compiler instances.
* send resulting asm through 'c++filt', so we don't have to deal with mangled names. (Hope we have it in CI, we should).
* filter out directives, comments and what not (based on what compiler explorer does + some extra)
* aggregate by overload in files (so - if you have multiple functions tested in one source file, they'd be split in separate files). Then aggregate by assmebly => by function name => by arch, so that we have to review multiple times the same asm.

## What if I need to iterate?

The problem with not using CMAKE - no incremental build. I still don't think it's worth the hustle to use CMAKE here, but what if you want to work on a specific file and rebuilding is annoying?

Solution is to pass '--arch <arch> --sources <file>'.
Smth like:

```{shell}
time CXX_FLAGS='--stdlib=libc++' python3.9 test/asm/scripts/validate_asm.py -j 8 --arch sse4 --sources load.cc
```

Or to generate

```{shell}
time CXX_FLAGS='--stdlib=libc++' python3.9 test/asm/scripts/generate_asm.py -j 8 --arch sse4 --sources load.cc
```

You can pass multiple sources and arches.

## Running the script:

This is how I generate ASM in expected:

```{shell}
time CXX_FLAGS='--stdlib=libc++' python3.9 test/asm/scripts/generate_asm.py -j 8
```

For just load it takes 1s. For load and store still about 1s.

This is how I run validation
```{shell}
time CXX_FLAGS='--stdlib=libc++' python3.9 test/asm/scripts/validate_asm.py -j 8
```

Still about 1s.

### PS

Already found 1 bug https://github.com/jfalcou/eve/issues/412
This probably messed up at least some of my measurements :)
Removed store for now.

### PSPS:

My aggregation logic is somewhat slow, might become an issue at scale,


DenisYaroshevskiy (commented):
_TODO_: 
* if the results for 2 architectures are identical: merge them.
  Maybe on.a function level?

UPD: done

test/asm/expected/load.s
```diff
@@ -0,0 +1,97 @@
...
+load(eve::aligned_ptr<short const, 8ul>, eve::fixed<4l>):        # avx, avx2
+load(eve::aligned_ptr<int const, 8ul>, eve::fixed<2l>):          # avx, avx2
+load(eve::aligned_ptr<long const, 8ul>, eve::fixed<1l>):         # avx, avx2
+load(eve::aligned_ptr<float const, 8ul>, eve::fixed<2l>):        # avx, avx2
+load(eve::aligned_ptr<double const, 8ul>, eve::fixed<1l>):       # avx, avx2
+    vmovsd    xmm0, qword ptr [rdi]           
+    ret
+
+
+load(eve::aligned_ptr<unsigned char const, 16ul>, eve::fixed<16l>):  # avx, avx2
```
jfalcou, line 63: I like this format

test/asm/expected/loadu.s
```diff
@@ -0,0 +1,97 @@
...
+
+
+loadu(signed char const*, eve::fixed<8l>):  # avx, avx2
+loadu(short const*, eve::fixed<4l>):        # avx, avx2
+loadu(int const*, eve::fixed<2l>):          # avx, avx2
+loadu(long const*, eve::fixed<1l>):         # avx, avx2
+loadu(float const*, eve::fixed<2l>):        # avx, avx2
+loadu(double const*, eve::fixed<1l>):       # avx, avx2
+    vmovsd    xmm0, qword ptr [rdi]           
+    ret
```
jfalcou, line 60: is the `ret` required to be kept ? It is obviously always there no ?

test/asm/sources/common.hpp
```diff
@@ -0,0 +1,34 @@
+#pragma once
+
+#include <cstdint>
+
+#define DELARE_ALL_FUNCTIONS              \
```
jfalcou, line 5: should be `DECLARE_ALL_FUNCTIONS` ?

test/asm/sources/load.cc
```diff
@@ -0,0 +1,21 @@
+#include <eve/eve.hpp>
+
+#include "common.hpp"
+
+#define DECLARE_TEST_FUNCTION(T, N) \
+  auto loadu(const T* p, N) { \
+    return eve::wide<T, N>{p}; \
+  }
+
+DELARE_ALL_FUNCTIONS
```
jfalcou, line 10: should be `DECLARE_ALL_FUNCTIONS` ?

test/asm/sources/load.cc
```diff
@@ -0,0 +1,21 @@
+#include <eve/eve.hpp>
+
+#include "common.hpp"
+
```
jfalcou, line 4: Do we need one file per function are can we aggregate them ?

jfalcou (commented):
Looks like a good start. The output format is great.
You may wish @jtlap to take a look at the python as he is more knowledgeable than me there.

test/asm/expected/loadu.s
DenisYaroshevskiy: Do you want me to remove it? OK.

test/asm/sources/load.cc
DenisYaroshevskiy: We don't. You can put them in whatever files you want (file names don't matter as well. The resulting asm will be split in files by function name, 
like I put `load` and `loadu` in one file.

I would like for a better way to generate all variations. If you habe suggestions, would be cool.
The question is to make compiler instantiate them

test/asm/expected/loadu.s
jfalcou: It's some byte we don't need, if we ever have huge files, maybe it helps to save those.
However, if it's too complex, just leave it

test/asm/sources/load.cc
jfalcou: It's tricky cause it depends on a lot of out-of-code knowledge.

DenisYaroshevskiy:
From slack:

matt  4:44 PM
x86-64 micro-architecture levels may be the closest to what you're looking for, https://sourceware.org/pipermail/libc-alpha/2020-July/116135.html (edited) 

Thiago Macieira  6:45 PM
# Architecture  Based on        New features
arch=Core2      x86_64  sse3,ssse3,cx16
arch=NHM        Core2   sse4.1,sse4.2,popcnt
arch=WSM        NHM
arch=SNB        WSM     avx
arch=IVB        SNB     f16c,rdrnd,fsgsbase
arch=HSW        IVB     avx2,fma,bmi,bmi2,lzcnt,movbe
arch=BDW        HSW     adx,rdseed
arch=BDX        BDW
arch=SKL        BDW     xsavec,xsaves
arch=SKX        SKL     avx512f,avx512dq,avx512cd,avx512bw,avx512vl,clwb
arch=CLX        SKX     avx512vnni
arch=CPX        CLX     avx512bf16
arch=CNL        SKL     avx512f,avx512dq,avx512cd,avx512bw,avx512vl,avx512ifma,avx512vbmi
arch=ICL        SKX     avx512ifma,avx512vbmi,avx512vbmi2,gfni,vaes,vpclmulqdq,avx512vnni,avx512bitalg,avx512vpopcntdq,fsrm
arch=ICX        ICL     pconfig
arch=TGL        ICL     avx512vp2intersect,shstk,movdiri,movdir64b,keylocker
arch=ADL        TGL     avxvnni,serialize,cldemote,waitpkg
arch=SPR        TGL     avx512bf16,amxtile,amxbf16,amxint8,avxvnni,cldemote,pconfig,waitpkg,serialize,tsxldtrk,uintr
:+1:
1

6:46
  # Atom line
arch=SLM        WSM     rdrnd,movbe
arch=GLM        SLM     fsgsbase,rdseed,lzcnt,xsavec,xsaves
arch=TNT        GLM     clwb,gfni,cldemote,waitpkg,movdiri,movdir64b


dyaroshev  7:03 PM
Thanks a lot!

Thiago Macieira  7:06 PM
note that I think the ADL is wrong.

Issue #35, opened 2019-02-24
Support AVX512


jfalcou:
New plan of actions:

- [x] Add support for detecting AVX512* ABI and related register type
- [x] Adapt logical to AVX512 types
- [x] Adapt internal operators
- [ ] Optimize other functions

jfalcou:
Meta-issue is supercede by https://github.com/jfalcou/eve/projects/5

PR #419, opened 2020-12-09
Incremental fix for #35 - MERGEABLE CONDITIONNALY

jfalcou:
This PR is failing on purpose up to the point everything will be set as per #35 steps


include/eve/arch/cpu/spec.hpp
```diff
@@ -22,9 +22,23 @@
...
+  inline constexpr bool supports_avx512popcntdq_    = spy::supports::avx512::popcntdq_;
+  inline constexpr bool supports_avx512_4fmaps_     = spy::supports::avx512::_4fmaps_;
+  inline constexpr bool supports_avx512vnniw_       = spy::supports::avx512::vnniw_;
+  inline constexpr bool supports_avx512vbmi_        = spy::supports::avx512::vbmi_;
+  inline constexpr bool supports_avx512bf16_        = spy::supports::avx512::bf16_;
+  inline constexpr bool supports_avx512bitalg_      = spy::supports::avx512::bitalg_;
+  inline constexpr bool supports_avx512vbmi2_       = spy::supports::avx512::vbmi2_;
+  inline constexpr bool supports_avx512vnni_        = spy::supports::avx512::vnni_;
+  inline constexpr bool supports_avx512vpintersect_ = spy::supports::avx512::vpintersect_;
+}
```
DenisYaroshevskiy, line 44: new lines :)
You gotta check your editor set up

include/eve/arch/x86/tags.hpp
```diff
@@ -17,6 +17,15 @@
...
 namespace eve
 {
+  namespace detail
+  {
+# if defined(SPY_SIMD_IS_X86_AVX512)
+    static inline constexpr bool x86_logical_status = false;
+# else
+    static inline constexpr bool x86_logical_status = true;
+# endif
+  }
```
DenisYaroshevskiy, line 27: What's this?

include/eve/arch/cpu/spec.hpp
```diff
@@ -22,9 +22,23 @@
...
+  inline constexpr bool supports_avx512vl           = spy::supports::avx512::vl_;
+  inline constexpr bool supports_avx512popcntdq_    = spy::supports::avx512::popcntdq_;
+  inline constexpr bool supports_avx512_4fmaps_     = spy::supports::avx512::_4fmaps_;
+  inline constexpr bool supports_avx512vnniw_       = spy::supports::avx512::vnniw_;
+  inline constexpr bool supports_avx512vbmi_        = spy::supports::avx512::vbmi_;
+  inline constexpr bool supports_avx512bf16_        = spy::supports::avx512::bf16_;
+  inline constexpr bool supports_avx512bitalg_      = spy::supports::avx512::bitalg_;
+  inline constexpr bool supports_avx512vbmi2_       = spy::supports::avx512::vbmi2_;
+  inline constexpr bool supports_avx512vnni_        = spy::supports::avx512::vnni_;
+  inline constexpr bool supports_avx512vpintersect_ = spy::supports::avx512::vpintersect_;
```
DenisYaroshevskiy, line 43: Ooooffff - this will get asm testing more unpleasant :)

include/eve/arch/x86/tags.hpp
```diff
@@ -42,8 +52,28 @@ namespace eve
...
+    static constexpr std::size_t expected_cardinal = bytes / sizeof(Type);
+  };
+
+  struct avx2_    : avx_ {};
+
+  struct avx512_  : avx2_
+  {
+    static constexpr std::size_t bits               = 512;
+    static constexpr std::size_t bytes              = 64;
+    static constexpr bool        is_bitwise_logical = detail::x86_logical_status;
```
DenisYaroshevskiy, line 72: ? I see one struct with `is_bit_logical` the other with `is_bitwise_logical`.
Not that either of them makes sense to me

include/eve/arch/x86/tags.hpp
DenisYaroshevskiy: This also seems a bit off.

`logical<wide<T, N, x86_256>>`

should not be bits. Or should it be bits anyways? 

Or should it also be ok to have

`logical<wide<char, eve::fixed<32>, _x86_512>`?

include/eve/arch/cpu/spec.hpp
jfalcou: Dang XD

include/eve/arch/x86/tags.hpp
jfalcou: When you activate AVX512, all logical values including the one from SSE/AVX have to be represented using the mmask registers. So I globally detect if AVX512 is up and fix this flag to true/false for later usage. It's a refactoring of not wanting the macro everywhere down

include/eve/arch/cpu/spec.hpp
jfalcou: According to https://en.wikipedia.org/wiki/AVX-512#CPUs_with_AVX-512 , they are usually grouped by blocks of micro architectures. SO Testing for "skylake-like" or "cannonlake-like" is maybe a better idea than testing all of them separately.

include/eve/arch/x86/tags.hpp
jfalcou: Obvious error is obvious.
The name is also maybe to rework. It's the indicator that the logical is either a regular register OR a compact bitmap register like mmask.

jfalcou: `logical_use_regular_register` is maybe better

jfalcou: This one structure was redundant

PR #420, opened 2020-12-11
#35 - AVX512 Support progress - PARTIALLY MERGEABLE 

jfalcou:
Step 2 : Implements basic operations on AVX512 wide/logical


include/eve/arch/cpu/logical_wide.hpp
```diff
@@ -267,11 +256,10 @@ namespace eve
...
     {
-      constexpr auto sz = sizeof(storage_type)/sizeof(logical<Type>);
-      auto that = bit_cast( p, as_<std::array<logical<Type>,sz>>());
-      os << '(' << that[0];
-      for(size_type i = 1; i != p.size(); ++i) os << ", " << that[i];
-      return os << ')';
+      auto that = p.bitmap();
+      os << '(' << std::boolalpha << that[0];
+      for(size_type i = 1; i < p.size(); ++i) os << ", " << that[i];
+      return os << std::noboolalpha << ')';
```
DenisYaroshevskiy, line 262: I dunno how much you care, but this seems to change the global state of things (if you had boolapha true, it will get it to false).

I would just not bother with boolalpha maybe and then you will just support it:if you want true false - you get true false, otherwise - numbers.

include/eve/arch/cpu/tags.hpp
```diff
@@ -33,26 +34,39 @@ namespace eve
...
   //================================================================================================
   // Dispatching tag for emulated SIMD implementation
   struct emulated_ : cpu_
   {
-    static constexpr std::size_t bits           = 128;
-    static constexpr std::size_t bytes          = 16;
-    static constexpr bool        is_bit_logical = true;
+    static constexpr std::size_t bits                     = 128;
+    static constexpr std::size_t bytes                    = 16;
+    static constexpr bool        regular_logical_register = true;
```
DenisYaroshevskiy, line 51: Any awesome bikeshed ideas? `is_wide_logical` for example?

include/eve/arch/cpu/tags.hpp
```diff
@@ -33,26 +34,39 @@ namespace eve
...
+  concept non_native_abi = detail::is_one_of<T>(detail::types<aggregated_, emulated_> {});
+
+  template<typename T>
+  concept native_abi = !detail::is_one_of<T>(detail::types<aggregated_, emulated_> {});
+
+  //================================================================================================
+  // Checks for logical status in ABI
+  template<typename ABI>
+  struct use_regular_logical_register : std::integral_constant<bool,ABI::regular_logical_register>
+  {};
```
DenisYaroshevskiy, line 67: Is this better for error messages then a typedef?

include/eve/detail/function/simd/x86/bitmask.hpp
```diff
@@ -11,74 +11,109 @@
...
+  // Logical to Bits
   //================================================================================================
   template<typename T, typename N, x86_abi ABI>
-  EVE_FORCEINLINE auto to_bitmap( sse2_ const&, logical<wide<T, N, ABI>> const& p ) noexcept
+  EVE_FORCEINLINE auto to_bits( sse2_ const&, logical<wide<T, N, ABI>> const& p ) noexcept
   {
-    using type = std::bitset<N::value>;
+    using type = typename logical<wide<T, N, ABI>>::bits_type;
+    return bit_cast(p.mask(), as_<type>{});
+  }
```
DenisYaroshevskiy, line 26: Do you need a `bit_cast`? Instead of just accessing the member?

include/eve/detail/function/simd/x86/bitmask.hpp
```diff
@@ -11,74 +11,109 @@
...
-        }
+      if constexpr( std::same_as<ABI,x86_128_>)
+      {
+              if constexpr( std::same_as<T,double > ) return _mm_mask_blend_pd(m,z,a);
+        else  if constexpr( std::same_as<T,float  > ) return _mm_mask_blend_ps(m,z,a);
+        else  if constexpr( sizeof(T) == 8          ) return _mm_mask_blend_epi64(m,z,a);
+        else  if constexpr( sizeof(T) == 4          ) return _mm_mask_blend_epi32(m,z,a);
+        else  if constexpr( sizeof(T) == 2          ) return _mm_mask_blend_epi16(m,z,a);
+        else  if constexpr( sizeof(T) == 1          ) return _mm_mask_blend_epi8(m,z,a);
+      }
```
DenisYaroshevskiy, line 48: Why do you need to blend anything - I am confused.
This is not `regular_logical` - so I assume - it's avx512 logical. Why blend?

Maybe I just don't know what `allbits` is though.

include/eve/detail/function/simd/x86/bitmask.hpp
```diff
@@ -11,74 +11,109 @@
...
         if constexpr( current_api >= avx2 )
         {
-          return type(_mm256_movemask_epi8(p.mask()));
+          return _mm256_movemask_epi8(p.mask());
         }
         else
         {
           auto [l, h] = p.slice();
-          return type( (h.bitmap().to_ullong() << h.size()) | l.bitmap().to_ullong() );
+          return  (h.bitmap().to_ullong() << h.size()) | l.bitmap().to_ullong() ;
```
DenisYaroshevskiy, line 116: extra space after return. And in the end.

include/eve/detail/function/simd/x86/load.hpp
```diff
@@ -72,4 +86,36 @@ namespace eve::detail
...
+                            , ABI const &, Iterator b, Iterator e
+                            ) noexcept
+  {
+    auto block = [&]() -> wide<T, N, ABI>
+    {
+      using tgt = eve::as_<wide<T, N, ABI>>;
+      return load(tgt(), ABI{}, b, e);
+    }();
+
+    return is_nez(block).storage();
```
DenisYaroshevskiy, line 119: I am very confused about this `is_nez`.
I also think that you don'y return a logical but rather it's underlying type that can be convertible to logical.

include/eve/detail/function/simd/x86/load.hpp
```diff
@@ -72,4 +86,36 @@ namespace eve::detail
...
+  template<typename T, typename N, typename Ptr, x86_abi ABI>
+  EVE_FORCEINLINE
+  auto load(eve::as_<logical<wide<T, N, ABI>>> const &, ABI const & abi, Ptr p)
+  requires( std::same_as<logical<T>, std::remove_cvref_t<decltype(*p)>> )
+  {
+    auto block = [&]() -> wide<T, N, ABI>
+    {
+      using tgt = eve::as_<wide<T, N, ABI>>;
+      if constexpr( !std::is_pointer_v<Ptr> ) return load(tgt(), abi, (T const*)(p.get()));
+      else                                    return load(tgt(), abi, (T const*)(p));
```
DenisYaroshevskiy, line 102: Why do you need to do the dispatch here? Won't the `load` you call figure out the dispatch?

jfalcou: This load is the internal one called by wide constructor. The overload system is not operating at this point yet.
This is one thing we need to live with.

include/eve/arch/cpu/tags.hpp
jfalcou: I like that. Will amend.

include/eve/arch/cpu/tags.hpp
jfalcou: That's classic meta-function forwarding style.

include/eve/detail/function/simd/x86/bitmask.hpp
jfalcou: `logical<wide<T>>` mask() returns a `wide<T>`. bits() returns the same wide but as a wide of integral values. Hence the bitcast. The one retuning the mask directly is to_bitmap.

include/eve/detail/function/simd/x86/bitmask.hpp
jfalcou: Because this is`to_mask` which purpose is to turn a logical<wide<T>> into a wide<T>. Not to_bitmap

include/eve/detail/function/simd/x86/bitmask.hpp
jfalcou: Fixed

include/eve/detail/function/simd/x86/load.hpp
jfalcou: As per Slack discussion, this is required to load data from scalar logical then compress to mmask.
This is due to be replaced when the typeless masks are implemented.

include/eve/arch/cpu/logical_wide.hpp
jfalcou: Chanegd to not depend on boolalpha state

PR #437, opened 2021-01-04
Make storage of logical<wide> saner

jfalcou:
Previous implementation of logical was using a fake wide<logical<T>> type to compute its register type and factoring out some implementation. This leads to some complicated code path.

`logical<wide>` is now independent and uses its own variant of as_register.


DenisYaroshevskiy:
Why can't it use `wide` as its storage (except for avx512)? Otherwise you'd have to duplicate things, no?

jfalcou:
> Why can't it use `wide` as its storage (except for avx512)? Otherwise you'd have to duplicate things, no?

Things are not that duplicated as most of it are just calls to external functions anyway.
The wide<logical<T>> was weird and prevents proper constraints on T in wide<T>.

This is also an intermediate state as the common parts of both logical and wide will be shipped in a common class to be reused and which will be the one providing the hidden friend implementation.

PR #439, opened 2021-01-08
Fix #435 - More hidden friend operators - INCOMPLETE

### LINKED ISSUE #435: [FEATURE] Make better usage of the 'hidden friend' idiom
As per :

https://www.justsoftwaresolutions.co.uk/cplusplus/hidden-friends.html

and extensive discussion with @DenisYaroshevskiy , we should aim at moving most if not all of our wide/logical operator into the hidden friend model.

Remaining operators to move:
 - [x] `operator&&`
 - [x] `operator||`
 - [x] `operator <<=`
 - [x] `operator <<`
 - [ ] `operator >>=`
 - [ ] `operator >>`
 - [x] `operator==`
 - [x] `operator!=`
 - [x] `operator>`
 - [x] `operator<`
 - [x] `operator>=`
 - [x] `operator<=`

jfalcou:


include/eve/arch/cpu/base.hpp
```diff
@@ -26,6 +28,8 @@ namespace eve::detail
     wide_storage() {}
     wide_storage(storage_type const &r) : data_(r) {}
 
+    wide_storage& operator=(storage_type const &r) { data_ = r; return *this; }
+
     EVE_FORCEINLINE storage_type const& storage() const & noexcept { return data_; }
     EVE_FORCEINLINE storage_type &      storage() &       noexcept { return data_; }
     EVE_FORCEINLINE storage_type        storage() &&      noexcept { return data_; }
```
DenisYaroshevskiy, line 35: I don't suggest to change it at this time but I heard that it's a good idea to have a `const &&` as well for some weird magic.
Not sure why you bother with move at all, I don't think we have a single non trivially copiable type but anyways

include/eve/arch/cpu/wide.hpp
```diff
@@ -121,59 +122,226 @@ namespace eve
...
-    // Compound operators
+    // Bitwise operators
+    //==============================================================================================
+    friend EVE_FORCEINLINE auto operator~(wide const& v) noexcept
+    {
+      return detail::self_bitnot(v);
+    }
+
+    friend  EVE_FORCEINLINE auto operator&=(wide& w, value auto o) noexcept
+                        ->  decltype(detail::self_bitand(w, o))
```
DenisYaroshevskiy, line 133: Why not just auto? We don't deduce type anywhere else?

Also I believe that `operator&=` and such should return a reference

include/eve/arch/cpu/wide.hpp
```diff
@@ -121,59 +122,226 @@ namespace eve
...
+    {
+      return detail::self_bitand(w, o);
+    }
+
+    template<scalar_value U,typename M>
+    friend EVE_FORCEINLINE    auto operator&(wide const& v, wide<U,M> const& w) noexcept
+                          ->  std::decay_t<decltype(std::declval<wide&>() &= w)>
```
DenisYaroshevskiy, line 140: Here too

include/eve/arch/cpu/wide.hpp
```diff
@@ -121,59 +122,226 @@ namespace eve
...
+    {
+      auto    that  = v;
+      return  that &= w;
+    }
+
+    friend EVE_FORCEINLINE auto operator&(wide const& v, scalar_value auto w) noexcept
+    requires (sizeof(w) == sizeof(Type))
+    {
+      auto    that  = v;
+      return  that &= bit_cast(w, as_<Type>{});
```
DenisYaroshevskiy, line 150: Why are you allowing to and with anything of the same size? That's a bit loose.

include/eve/arch/cpu/wide.hpp
```diff
@@ -121,59 +122,226 @@ namespace eve
...
+      return detail::self_bitor(w, o);
+    }
+
+    template<scalar_value U,typename M>
+    friend EVE_FORCEINLINE    auto operator|(wide const& v, wide<U,M> const& w) noexcept
+                          ->  std::decay_t<decltype(std::declval<wide&>() |= w)>
+    {
+      wide    that  = v;
+      return  that |= w;
+    }
```
DenisYaroshevskiy, line 173: Same comments as for and

include/eve/arch/cpu/wide.hpp
```diff
@@ -121,59 +122,226 @@ namespace eve
...
+      return detail::self_bitxor(w, o);
+    }
+
+    template<scalar_value U, typename M>
+    friend EVE_FORCEINLINE    auto operator^(wide const& v, wide<U,M> const& w) noexcept
+                          ->  std::decay_t<decltype(std::declval<wide&>() |= w)>
+    {
+      auto    that  = v;
+      return  that ^= w;
+    }
```
DenisYaroshevskiy, line 202: Why id it a good idea to ^ arbitrary wides

include/eve/arch/cpu/wide.hpp
```diff
@@ -121,59 +122,226 @@ namespace eve
...
+    {
+      return wide(s) - v;
+    }
+
+    friend EVE_FORCEINLINE auto operator-(wide const& v, real_scalar_value auto s) noexcept
+    {
+      return v - wide(s);
+    }
+
     friend  EVE_FORCEINLINE auto operator*=(wide& w, value auto o) noexcept
```
DenisYaroshevskiy, line 273: Should return a &

include/eve/arch/cpu/base.hpp
jfalcou: I wanted to be sure that calling `storage()` on temporary wide didn't fuck up.
I heard this `const&&` too, will look later

include/eve/arch/cpu/wide.hpp
jfalcou: self_bitand has non-trivial constraints that I didn't want to be replicated here. So we use auto -> decltype to SFINAE out the operator if self_bitand can't be called. All the operators are merely trampolining to the proper function, hence the auto -> decltype.

As per &, self_bitand returns a ref, so the decltype is properly computed.

include/eve/arch/cpu/wide.hpp
jfalcou: same

include/eve/arch/cpu/wide.hpp
jfalcou: That's what we do since version 0. We want BITwise operations to works on BIT, no matter the underlying type as what is important here is that each bit has a counterpart.

It helps writing a lot of SIMD math where you do bit and/or/wor between floating point values and masks.

DenisYaroshevskiy:
LGTM

PR #441, opened 2021-01-09
Make ignore_*::count() returns a size in # of elements

jfalcou:


include/eve/conditional.hpp
```diff
@@ -130,12 +130,17 @@ namespace eve
       return mask;
     }
 
-    template<typename T> EVE_FORCEINLINE std::ptrdiff_t offset(eve::as_<T> const&) const
+    template<typename T> EVE_FORCEINLINE constexpr std::ptrdiff_t offset(eve::as_<T> const&) const
     {
       return 0;
```
DenisYaroshevskiy, line 135: This 0 is iffy. But should work

DenisYaroshevskiy (approved):
I don't like that we don't have any tests. Will add some for top_bits that cover this.

include/eve/conditional.hpp
jfalcou: > I don't like that we don't have any tests. Will add some for top_bits that cover this.

Seems better in top_bits IMHO

jfalcou:
> I don't like that we don't have any tests. Will add some for top_bits that cover this.

Seems better in top_bits IMHO

PR #447, opened 2021-01-11
Fix #443 - No [] for logical. Nope.

### LINKED ISSUE #443: [BUG] logical::operator[] is broken
The following code:

```
mask[i] = true;
```

compiles but silently doesn't work.

jfalcou:
```
logical<wide<int>> l;
l[i] = true;
```

now properly fails at compile-time instead of being a silent noop.

Implementation is based on :
https://lesleylai.info/en/type-of-assignment-operators/
http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2009/n2819.html

### DIFF
--- a/include/eve/arch/cpu/logical.hpp
+++ b/include/eve/arch/cpu/logical.hpp
@@ -44,7 +44,13 @@ namespace eve
     //==============================================================================================
     // Assignment
     //==============================================================================================
-    EVE_FORCEINLINE constexpr logical &operator=(bool v) noexcept
+    EVE_FORCEINLINE constexpr logical &operator=(logical v) & noexcept
+    {
+      value_ = v.value_;
+      return *this;
+    }
+
+    EVE_FORCEINLINE constexpr logical &operator=(bool v) & noexcept
     {
       value_ = v ? true_mask : false_mask;
       return *this;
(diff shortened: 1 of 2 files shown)


include/eve/arch/cpu/logical.hpp
```diff
@@ -44,7 +44,13 @@ namespace eve
...
     // Assignment
     //==============================================================================================
-    EVE_FORCEINLINE constexpr logical &operator=(bool v) noexcept
+    EVE_FORCEINLINE constexpr logical &operator=(logical v) & noexcept
+    {
+      value_ = v.value_;
+      return *this;
+    }
+
+    EVE_FORCEINLINE constexpr logical &operator=(bool v) & noexcept
```
DenisYaroshevskiy, line 53: This makes no sense, sorry.

I mean - one should not be able to assign to a temporary in any case.
This is not a good fix for it.

Do you just return a `logical` by value?

I would suggest to drop `operator[]` all together or return a proxy object.

But this is not very good I think.

jfalcou: I added some references. If you're OK as per our Slack discussion, I'll proceed to the merge

DenisYaroshevskiy: Merging

PR #450, opened 2021-01-13
FIX-290: just enable asan

DenisYaroshevskiy:
I want to see just how long it would take.


DenisYaroshevskiy:
Is a bad idea at the moment - way too slow.

PR #453, opened 2021-01-15
FIX-451: char -> std::byte to avoid const char*

DenisYaroshevskiy:
ASAN failure was happening inside TTS, as_string.

as_string calls `operator<<` which treats `const char*` as c-string and not as a pointer value.
For aligned_ptr tests rewrote as `std::byte`


jfalcou (approved):
LGTM

PR #468, opened 2021-01-19
Significants 1

jtlap:
keep n significant digits


include/eve/module/real/core/function/regular/generic/significants.hpp
```diff
@@ -0,0 +1,38 @@
...
+                               , U const &n) noexcept
+  {
+    if constexpr(has_native_abi_v<T> && has_native_abi_v<U>)
+    {
+      using elt_t =  element_type_t<T>;
+      auto factor = exp10(convert(n, as<elt_t>())- ceil(log10(eve::abs(a))));
+      return nearest(a * factor) / factor;
+    }
+    else  return apply_over(significants, a, n);
+  }
```
DenisYaroshevskiy, line 37: Questions: 
1) why do you need a check for `has_native_abi`? Should just work, no?
2) Should you, maybe, have a requirement that T and U have the same cardinality?

test/unit/module/real/core/significants/regular/significants.hpp
```diff
@@ -0,0 +1,28 @@
...
+TTS_CASE_TPL("Check eve::significants behavior", EVE_TYPE)
+{
+  TTS_EQUAL(eve::significants(T(1.234567), 1), T(1.) );
+  TTS_EQUAL(eve::significants(T(1.234567), 2), T(1.2) );
+  TTS_EQUAL(eve::significants(T(1.234567), 3), T(1.23) );
+  TTS_EQUAL(eve::significants(T(1.234567), 4), T(1.235) );
+  TTS_EQUAL(eve::significants(T(0.01234567), 1),T(0.01) );
+  TTS_EQUAL(eve::significants(T(0.01234567), 2),T(0.012) );
+  TTS_EQUAL(eve::significants(T(0.01234567), 3),T(0.0123) );
+  TTS_EQUAL(eve::significants(T(0.01234567), 4),T(0.01235) );
```
DenisYaroshevskiy, line 27: How about negative numbers tests?
I also see some rounding tested, which is great, but maybe: `987654321` as well?
And rounding to 0 maybe?

DenisYaroshevskiy (commented):
Was just passing by.

include/eve/module/real/core/function/regular/generic/significants.hpp
jtlap: the problem is:
1) it is better to split the entries to natives has soon as possible, as it reduces the splits numbers as apply over does so
2) T and U can be wide or scalar: apply_over take also care of that 
I hope this is a correct answer

test/unit/module/real/core/significants/regular/significants.hpp
jtlap: you're right I will enhance tests

include/eve/module/real/core/function/regular/generic/significants.hpp
DenisYaroshevskiy: I didn't think about scalar - that makes sense.

In general - if you want to see the codegen, you can disassemble any cpp file, similar to compiler explorer.

```
clang++ -std=c++20 --stdlib=libc++ -O3 -DEVE_DISABLE_ASSERTS -march=native -I  ../../eve/include -I ../src -S -mllvm --x86-asm-syntax=intel disassemble.cc
```

Is the command I use

jtlap: thank you, I will try

test/unit/module/real/core/significants/regular/significants.hpp
```diff
@@ -0,0 +1,55 @@
...
+
+  TTS_IEEE_EQUAL(eve::significants(T(-1.234567), 0), eve::nan(eve::as<T>()) );
+  TTS_EQUAL(eve::significants(T(-1.234567), 1), T(-1.) );
+  TTS_EQUAL(eve::significants(T(-1.234567), 2), T(-1.2) );
+  TTS_EQUAL(eve::significants(T(-1.234567), 3), T(-1.23) );
+  TTS_EQUAL(eve::significants(T(-1.234567), 4), T(-1.235) );
```
DenisYaroshevskiy, line 35: I was surprised to learn that this is correct 👍

DenisYaroshevskiy (commented):
LGTM.
@jfalcou should still review.

_NOTE_
I doubt that this is expected to be an extremely hot function, is it?

If we were really worried, we could look for some nifty tricks to do rounding without exp and log but I don't know it.

Found this piece of code online for 2 digits
https://www.geeksforgeeks.org/rounding-floating-point-number-two-decimal-places-c-c/

But I don't see how it will help.

docs/reference/functions/rounding/significants.html
```diff
@@ -0,0 +1,72 @@
...
+:   [integral value](../../concepts.html#value).
+
+Return value
+----------------------------------------------------------------------------------------------------
+<span class="smallskip"></span>
+Computes  [element-wise](../../../glossary.html#elment-wise) the rounding to n significants digits of `x`.
+With null n the result is a nan.
+
+!!! WARNING
+   This is an helper function. Floting numbers are not stored in decimal form. So if you try significants with
```
jfalcou, line 54: Floting -> Floating

docs/reference/functions/rounding/significants.html
```diff
@@ -0,0 +1,72 @@
...
+`n`
+:   [integral value](../../concepts.html#value).
+
+Return value
+----------------------------------------------------------------------------------------------------
+<span class="smallskip"></span>
+Computes  [element-wise](../../../glossary.html#elment-wise) the rounding to n significants digits of `x`.
+With null n the result is a nan.
+
+!!! WARNING
```
jfalcou, line 53: The part abotu it beign an helper adds no infos. Just states it has accuracy limitation.

jfalcou (changes requested):

jfalcou (approved):

PR #469, opened 2021-01-19
Fix #446: top bits avx512

### LINKED ISSUE #446: [FEATURE] top bits on AVX-512
Top bits was implemented for everything x86 except for avx512.
Implement for avx512.

DenisYaroshevskiy:
The `operator` looks a bit heavy.
Generally speaking - it's recommended that they all converge to the same one, but I copied what @jfalcou did. 
Up to you - I can rewrite in classic stl style.

Didn't find a `uint_t<N>` implementation, rolled my own. Do you maybe have one? 
I needed it because I need to figure out to what I should cast.

There is also a question if we maybe want to test `mask_n`

### DIFF
--- a/include/eve/arch/x86/as_register.hpp
+++ b/include/eve/arch/x86/as_register.hpp
@@ -101,35 +101,34 @@ namespace eve
 
       // != and reverse are generated.
       // <=> didn't work for some reason for ==
-      friend bool operator==(mask_n, mask_n) = default;
-      friend bool operator==(mask_n m, type n) { return m == mask_n{n}; }
-      friend auto operator<=>(mask_n, mask_n) = default;
-
+      friend bool operator== (mask_n m, type n) { return m == mask_n{n}; }
+      friend bool operator== (mask_n  , mask_n) = default;
+      friend auto operator<=>(mask_n  , mask_n) = default;
 
       friend constexpr mask_n operator~(mask_n m) { return mask_n{type{~m.value}}; }
 
       friend constexpr mask_n& operator&=(mask_n& m, mask_n n) { m.value &= n.value; return m;     }
-      friend constexpr mask_n& operator&=(mask_n& m, type  n) { m.value &= n;       return m;     }
-      friend constexpr mask_n operator& (mask_n  m, mask_n n) { return mask_n{m.value & n.value};  }
-      friend constexpr mask_n operator& (mask_n  m, type  n) { return mask_n{m.value & n};        }
-      friend constexpr mask_n operator& (type   m, mask_n n) { return mask_n{n.value & n.value};  }
+      friend constexpr mask_n& operator&=(mask_n& m, type   n) { m.value &= n;       return m;     }
+      friend constexpr mask_n  operator& (mask_n  m, mask_n n) { mask_n t{m.value}; return t &= n; }
+      friend constexpr mask_n  operator& (mask_n  m, type   n) { mask_n t{n}; return t &= m;       }
+      friend constexpr mask_n  operator& (type    m, mask_n n) { mask_n t{m}; return t &= n;       }
 
       friend constexpr mask_n& operator|=(mask_n& m, mask_n n) { m.value |= n.value; return m;     }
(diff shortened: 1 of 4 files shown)


include/eve/detail/top_bits.hpp
```diff
@@ -34,12 +34,34 @@ namespace eve::detail
 
 // bit utilities ---------------------
 
+template <std::size_t N>
+auto uint_t_impl() {
```
jfalcou, line 38: This is `make_integer_t<N*8,unsigned>`. It is defined in detail/meta.hpp

include/eve/detail/top_bits.hpp
```diff
@@ -50,7 +72,8 @@ EVE_FORCEINLINE auto movemask_raw(Logical p)
   using ABI = abi_type_t<Logical>;
   using e_t = element_type_t<Logical>;
 
-  if constexpr( std::same_as<ABI, x86_128_>)
+       if constexpr( eve::current_api == avx512 ) return p.storage();
```
jfalcou, line 75: This works but we usually check for `!ABI::is_wide_logical` just in case Intel drop us something strange later on

include/eve/detail/top_bits.hpp
```diff
@@ -92,7 +111,7 @@ struct top_bits
   static constexpr bool is_avx512_logical = !abi_type_t<Logical>::is_wide_logical;
   static constexpr std::ptrdiff_t static_size = Logical::static_size;
   static constexpr std::ptrdiff_t bits_per_element =
-    sizeof(element_type_t<Logical>) == 2 ? 2 : 1;
+    sizeof(element_type_t<Logical>) == 2 && !is_avx512_logical ? 2 : 1;
```
jfalcou, line 114: You compute is_avx512_logical as ` !abi_type_t<Logical>::is_wide_logical` then you re apply ! in the ternary condition.
Is it correct or is one of those ! useless ?

jfalcou (changes requested):
LGTM apart from those little detail.

Can you elaborate or show what you mean by making the operator on mmask more stl like ?

DenisYaroshevskiy:
| Can you elaborate or show what you mean by making the operator on mmask more stl like ?

Well Alexander Stepanov was always insisting that operators call their other forms.

Like: `operator&` should call `operator&=` and such.

All our operators on `type` should construct `mask_n` and call that operator.

We can even maybe write some crtp thing to do it for us.

jfalcou:
> | Can you elaborate or show what you mean by making the operator on mmask more stl like ?
> 
> Well Alexander Stepanov was always insisting that operators call their other forms.
> 
> Like: `operator&` should call `operator&=` and such.
> 
> All our operators on `type` should construct `mask_n` and call that operator.
> 
> We can even maybe write some crtp thing to do it for us.

Ok so : 

```
friend constexpr mask_n operator& (mask_n  m, mask_n n)   { mask_n that{m.value}; return that &= n; }
friend constexpr mask_n operator& (mask_n  m, type  n)    { mask_n that{n}; return that &= m;       }
friend constexpr mask_n operator& (type   m, mask_n n)    { mask_n that{m}; return that &= n;       }
```

and so on ?
I'm OK with that. I wrote the old one in a hurry.

jfalcou (approved):

include/eve/arch/x86/as_register.hpp
```diff
@@ -101,35 +101,34 @@ namespace eve
 
       // != and reverse are generated.
       // <=> didn't work for some reason for ==
-      friend bool operator==(mask_n, mask_n) = default;
-      friend bool operator==(mask_n m, type n) { return m == mask_n{n}; }
-      friend auto operator<=>(mask_n, mask_n) = default;
-
+      friend bool operator== (mask_n m, type n) { return m == mask_n{n}; }
+      friend bool operator== (mask_n  , mask_n) = default;
+      friend auto operator<=>(mask_n  , mask_n) = default;
```
DenisYaroshevskiy, line 106: Missed xomparison with type. Not that I care

PR #474, opened 2021-01-22
Fix #472: all (scalar and x86)

### LINKED ISSUE #472: [FEATURE] ignore support in all/any/none.
As part of this feature, not going to touch:
* ppc - I don't yet know how to compile to power pc and we don't have ppc tests on CI yet.
* arm-v8 special implementation. We don't have the tests on CI yet.

DenisYaroshevskiy:
(I was a bit optimistic when I was naming the branch).

For ignore any - just do the top bits. Well. except we need to invert the ignore.
Maybe we could do smth clever with aggregated case and ignore but I didn't figure it out.

Because the logic is quite different I left both ignore and non ignore overloads.

ARM and any/none will do separately at some point.

### DIFF
--- a/include/eve/module/real/algorithm/function/regular/simd/x86/all.hpp
+++ b/include/eve/module/real/algorithm/function/regular/simd/x86/all.hpp
@@ -12,8 +12,8 @@
 
 #include <eve/concept/value.hpp>
 #include <eve/detail/concepts.hpp>
-#include <eve/detail/implementation.hpp>
-#include <eve/function/bit_cast.hpp>
+#include <eve/detail/top_bits.hpp>
+
 
 namespace eve::detail
 {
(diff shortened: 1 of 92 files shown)


include/eve/module/real/algorithm/function/regular/simd/x86/all.hpp
```diff
@@ -23,20 +23,19 @@ namespace eve::detail
...
-    else
-    {
-      return v.bitmap().all();
-    }
+    if constexpr( N::value == 1) return v[0];
+    else                         return eve::detail::all(eve::detail::top_bits{v});
+  }
+
+  template<logical_simd_value T, relative_conditional_expr C>
+    requires (!eve::has_emulated_abi_v<T>)
```
jfalcou, line 31: [Style] I usually put requires on function after the noexcept

jfalcou (commented):
LGTM

PR #477, opened 2021-01-22
Statistic distribs

jtlap:
just cauchy


jtlap:
will be replaced by other implementation

PR #483, opened 2021-01-24
Cleaning operators

jtlap:
cleaning as much as possible

### DIFF
--- a/test/doc/core/is_equal.cpp
+++ b/test/doc/core/is_equal.cpp
@@ -1,4 +1,4 @@
-#include <eve/function/predicate.hpp>
+#include <eve/function/is_equal.hpp>
 #include <eve/wide.hpp>
 #include <iostream>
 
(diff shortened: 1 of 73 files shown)


test/doc/core/is_equal.cpp
```diff
@@ -12,7 +12,7 @@ int main()
   std::cout << "---- simd" << '\n'
             << "<- pf               = " << pf << '\n'
             << "<- qf               = " << qf << '\n'
-            << "-> is_equal(pf, qf) = " << eve::is_equal(pf, qf) << '\n';
+            << "-> is_equal(pf ==  qf) = " << eve::is_equal(pf, qf) << '\n';
```
jfalcou, line 15: why this change ?

jfalcou (changes requested):

test/doc/core/is_equal.cpp
jtlap: an error

jfalcou (approved):

PR #485, opened 2021-01-24
FIX-414: enable asan for avx2

DenisYaroshevskiy:
We really need an asan build to test some of our functionality.

Decided that 'x86_avx2' will have asan enabled.

Also, usually asan is enabled with: '-g -fno-omit-frame-pointer'. THis is important for better output readability. But I am interested, if I omit them - will the builds run faster? After all - building with ASAN locally to find where the issue is, is not a big deal.


DenisYaroshevskiy:
We will look into this after the new containers.

PR #486, opened 2021-01-24
Various warning fixes

jfalcou:
Fixed a bunch of warnings related to g++ with -Wextra being really snappy


DenisYaroshevskiy (approved):
Looks OK, though I maybe'd use [[maybe_unsued]] instead of pragmas but also works.

PR #487, opened 2021-01-24
FIX-414: remove -O3

DenisYaroshevskiy:
removing O3 seems to make local builds faster.


DenisYaroshevskiy:
As agreed with @jfalcou -merging

PR #492, opened 2021-01-26
FIX-290: load unsafe

### LINKED ISSUE #290: Request to add `load_unsafe`
Many simd algorithms, for example `strlen`, rely on the fact that allocation happens in pages.
See, for example: https://stackoverflow.com/a/25574201/5021064

The problem is - ASAN considers this a bug.

The way to work around it is to mark a function as `__attribute__((no_sanitize_address))` - and then ASAN will stop complaining about it.

So far I have been just using:

```
template <typename P, typename Tgt>
__attribute__((no_sanitize_address)) auto load_unsafe(P p, const Tgt& tgt) noexcept {
  return eve::load(p, tgt);
}
```

However, this does not work for emulated `wide`, consisting of segments. This is not doable up the stack, the reason it worked for small and natural wide, is that everything in the middle is marked `FORCEINLINE`. That is not a terribly nice thing to rely upon - can you maybe add `load_unsafe` function in the library?

DenisYaroshevskiy:
FIX #290 : add load_unsafe 

tested locally.

`no_sanitize_address` does not always affect that called functions and when the logical didn't work, it was the last straw.
So I decided that when asan is going to be enabled, I am just going to do this in scalar.
Alternative is to push it through all of the functions. We'd have to not use memcpy and what not.
I don't expect this to massively damage people's asan performance, since unsafe is mostly used for corner cases.

Also added `previous_aligned_address`: as discussed, a parameter is a pointer and a cardinal - number of elements to align to (default - expected_cardinal_t)

### DIFF
--- a/include/eve/detail/function/simd/x86/friends.hpp
+++ b/include/eve/detail/function/simd/x86/friends.hpp
@@ -145,7 +145,7 @@ namespace eve::detail
     if constexpr( !ABI::is_wide_logical )
     {
       using s_t = typename logical<wide<T,N,ABI>>::storage_type;
-      return logical<wide<T,N,ABI>>{ s_t(~(v.storage().value ^ w.storage().value)) };
+      return logical<wide<T,N,ABI>>{ s_t({~(v.storage().value ^ w.storage().value)}) };
     }
     else
     {
--- a/include/eve/detail/function/simd/x86/load.hpp
+++ b/include/eve/detail/function/simd/x86/load.hpp
@@ -90,7 +90,7 @@ namespace eve::detail
   //================================================================================================
   // logical loads require some specific setup
   //================================================================================================
-  template<typename T, typename N, typename Ptr, x86_abi ABI>
+  template<typename T, typename N, typename Ptr, typename ABI>
   EVE_FORCEINLINE
   auto load([[maybe_unused]] eve::as_<logical<wide<T, N, ABI>>> const& tgt, Ptr p)
   requires( std::same_as<logical<T>, std::remove_cvref_t<decltype(*p)>> )
@@ -113,7 +113,7 @@ namespace eve::detail
     else                                  return bit_cast(block, tgt);
   }
 
-  template<typename Iterator, typename T, typename N, x86_abi ABI>
+  template<typename Iterator, typename T, typename N, typename ABI>
   EVE_FORCEINLINE auto load ( eve::as_<logical<wide<T, N, ABI>>> const &
                             , Iterator b, Iterator e
                             ) noexcept
--- a/include/eve/detail/function/simd/x86/load.hpp
+++ b/include/eve/detail/function/simd/x86/load.hpp
@@ -90,7 +90,7 @@ namespace eve::detail
   //================================================================================================
   // logical loads require some specific setup
   //================================================================================================
-  template<typename T, typename N, typename Ptr, typename ABI>
+  template<typename T, typename N, typename Ptr, x86_abi ABI>
   EVE_FORCEINLINE
   auto load([[maybe_unused]] eve::as_<logical<wide<T, N, ABI>>> const& tgt, Ptr p)
   requires( std::same_as<logical<T>, std::remove_cvref_t<decltype(*p)>> )
@@ -113,7 +113,7 @@ namespace eve::detail
     else                                  return bit_cast(block, tgt);
   }
 
-  template<typename Iterator, typename T, typename N, typename ABI>
+  template<typename Iterator, typename T, typename N, x86_abi ABI>
   EVE_FORCEINLINE auto load ( eve::as_<logical<wide<T, N, ABI>>> const &
                             , Iterator b, Iterator e
                             ) noexcept
--- a/include/eve/detail/overload.hpp
+++ b/include/eve/detail/overload.hpp
@@ -49,6 +49,16 @@
                 };                                                                                 \
       }                                                                                            \
                                                                                                    \
+      template<attribute_conditional Condition>                                                    \
+      EVE_FORCEINLINE constexpr auto operator[](Condition const &c) const noexcept                 \
+      {                                                                                            \
+        return  [c](auto const&... args) EVE_LAMBDA_FORCEINLINE                                    \
+                {                                                                                  \
+                  return callable_object{}(c, args...);                                            \
+                };                                                                                 \
+      }                                                                                            \
+                                                                                                   \
+                                                                                                   \
       template<typename Arg, typename... Args>                                                     \
       EVE_FORCEINLINE constexpr auto operator()(Arg&& d, Args &&... args) const noexcept           \
           -> decltype ( TAG(delay_t{}, EVE_CURRENT_API{},                                          \
--- a/include/eve/detail/overload.hpp
+++ b/include/eve/detail/overload.hpp
@@ -49,16 +49,6 @@
                 };                                                                                 \
       }                                                                                            \
                                                                                                    \
-      template<attribute_conditional Condition>                                                    \
-      EVE_FORCEINLINE constexpr auto operator[](Condition const &c) const noexcept                 \
-      {                                                                                            \
-        return  [c](auto const&... args) EVE_LAMBDA_FORCEINLINE                                    \
-                {                                                                                  \
-                  return callable_object{}(c, args...);                                            \
-                };                                                                                 \
-      }                                                                                            \
-                                                                                                   \
-                                                                                                   \
       template<typename Arg, typename... Args>                                                     \
       EVE_FORCEINLINE constexpr auto operator()(Arg&& d, Args &&... args) const noexcept           \
           -> decltype ( TAG(delay_t{}, EVE_CURRENT_API{},                                          \
--- a/include/eve/memory/aligned_ptr.hpp
+++ b/include/eve/memory/aligned_ptr.hpp
@@ -10,12 +10,13 @@
 //==================================================================================================
 #pragma once
 
-#include <eve/memory/is_aligned.hpp>
-#include <eve/traits/element_type.hpp>
+#include <eve/assert.hpp>
 #include <eve/detail/concepts.hpp>
+#include <eve/memory/align.hpp>
+#include <eve/memory/is_aligned.hpp>
 #include <eve/memory/is_aligned.hpp>
 #include <eve/traits/alignment.hpp>
-#include <eve/assert.hpp>
+#include <eve/traits/element_type.hpp>
 #include <compare>
 
 namespace eve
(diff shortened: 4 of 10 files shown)


test/unit/api/wide/load/arithmetic/load.hpp
```diff
@@ -293,3 +295,37 @@ TTS_CASE_TPL("Check load from range for wide", EVE_TYPE )
...
   TTS_EQUAL(from_begin_end, ref);
 }
+
+TTS_CASE_TPL("Check load unsafe, wide", EVE_TYPE)
+{
+  eve::element_type_t<T> x = 123;
+
+  {
+    auto ptr = eve::previous_aligned_address(
+      &x, eve::lane<sizeof(x) * T::static_size>);
```
DenisYaroshevskiy, line 305: This doesn't work for smaller wides, and probably for bigger wides too. A bit annoying that we can't just `load` a wide. oh well.

test/unit/api/memory/aligned_ptr.cpp
```diff
@@ -170,3 +170,20 @@ TTS_CASE("aligned_ptr provides pointer-like interface")
     }
   }
 }
+
+TTS_CASE("previous aligned address")
+{
+  alignas(16) std::array<std::byte, 16> data;
```
DenisYaroshevskiy, line 176: this array should be more than 16 in size and check for every 16 elements. This way the over aligned stack variable won't cause issues.

DenisYaroshevskiy (commented):
* tests that load unsafe actually loads data correctly
* tests for logical

include/eve/detail/overload.hpp
```diff
@@ -49,6 +49,16 @@
...
+      template<attribute_conditional Condition>                                                    \
+      EVE_FORCEINLINE constexpr auto operator[](Condition const &c) const noexcept                 \
+      {                                                                                            \
+        return  [c](auto const&... args) EVE_LAMBDA_FORCEINLINE                                    \
+                {                                                                                  \
+                  return callable_object{}(c, args...);                                            \
+                };                                                                                 \
+      }                                                                                            \
+                                                                                                   \
+                                                                                                   \
```
DenisYaroshevskiy, line 61: remove

include/eve/module/real/core/function/regular/generic/load.hpp
```diff
@@ -120,8 +121,48 @@ namespace eve::detail
...
+  DISABLE_SANITIZERS auto load_(EVE_SUPPORTS(cpu_), unsafe_type, Ptr ptr, Cardinal const & cardinal) noexcept
+    requires requires(Ptr ptr, Cardinal cardinal) { eve::load(ptr, cardinal); }
+  {
+    using e_t = std::remove_cvref_t<decltype(*ptr)>;
+    using r_t = as_wide_t< e_t, typename Cardinal::type >;
+
+    if constexpr ( !has_aggregated_abi_v<r_t> ) return eve::load(ptr, cardinal);
+    else
+    {
+      constexpr eve::fixed< Cardinal() / 2 > half;
```
DenisYaroshevskiy, line 150: _TODO_: this won't work for over aligned ptrs.
Also at the moment, seems like for aggregated we require the same alignment as for the smallest native. Talk to @jfalcou about it in https://github.com/jfalcou/eve/issues/495 and decide what's the right answer is.

test/unit/api/wide/load/logical/load.hpp
```diff
@@ -321,3 +327,61 @@ TTS_CASE_TPL("load for different alignment", EVE_TYPE )
     test(eve::lane<4>);
   }
 }
+
+#if 0
```
DenisYaroshevskiy, line 331: @jfalcou - can I ask you to try it out?
I think it fails  because of https://github.com/jfalcou/eve/issues/507
fails only on avx512.

include/eve/detail/function/simd/x86/friends.hpp
```diff
@@ -145,7 +145,7 @@ namespace eve::detail
     if constexpr( !ABI::is_wide_logical )
     {
       using s_t = typename logical<wide<T,N,ABI>>::storage_type;
-      return logical<wide<T,N,ABI>>{ s_t(~(v.storage().value ^ w.storage().value)) };
+      return logical<wide<T,N,ABI>>{ s_t({~(v.storage().value ^ w.storage().value)}) };
```
jfalcou, line 148: why ?

include/eve/detail/function/simd/x86/load.hpp
```diff
@@ -90,7 +90,7 @@ namespace eve::detail
   //================================================================================================
   // logical loads require some specific setup
   //================================================================================================
-  template<typename T, typename N, typename Ptr, x86_abi ABI>
+  template<typename T, typename N, typename Ptr, typename ABI>
```
jfalcou, line 93: if this is no more x86 related, it should be moved to the detail/function/simd/common/load.hpp file

include/eve/detail/function/simd/x86/load.hpp
```diff
@@ -105,7 +105,7 @@ namespace eve::detail
     return to_logical(block).storage();
   }
 
-  template<typename Iterator, typename T, typename N, x86_abi ABI>
+  template<typename Iterator, typename T, typename N, typename ABI>
```
jfalcou, line 108: Same for this one

include/eve/module/real/core/function/regular/generic/load.hpp
```diff
@@ -120,6 +143,47 @@ namespace eve::detail
...
+  template<typename Ptr, typename Cardinal>
+  DISABLE_SANITIZERS auto load_(EVE_SUPPORTS(cpu_), unsafe_type, Ptr ptr, Cardinal const& N) noexcept
+    requires sanitizers_are_on && (requires(Ptr ptr, Cardinal const& N) { eve::load(ptr, N); })
+  {
+    using e_t = std::remove_cvref_t<decltype(*ptr)>;
+    using r_t = as_wide_t< e_t, typename Cardinal::type >;
+
+    r_t that;
+
+    for (std::ptrdiff_t i = 0; i != N(); ++i) that.set(i, ptr[i]);
```
jfalcou, line 167: This can be replaced by :

```
return []<std::size_t... I>( std::index_sequence<I...>)
{
  return r_t{ ptr[I]... };
}( std::make_index_sequence<N::value>{});
```

it is still filled in a scalar way BUT call the mm_set or similar intrinsic.
Also the loop is garanteed to be compile-time unrolled.

test/unit/api/wide/load/arithmetic/load.hpp
```diff
@@ -315,7 +318,10 @@ TTS_CASE_TPL("load for different alignment", EVE_TYPE )
       if constexpr (A >= T::static_alignment)
       {
         eve::aligned_ptr<const e_t, static_cast<std::size_t>(A)> ptr{f};
-        T actual{ptr};
+        eve::wide actual = T{ptr};
```
jfalcou, line 321: `auto actual = T{ptr}` ?

jfalcou (changes requested):

include/eve/detail/function/simd/x86/friends.hpp
DenisYaroshevskiy: clang compilation error:

```
n file included from /home/dyarosh/space/eve/build_avx512/../include/eve/detail/function/friends.hpp:17:
/home/dyarosh/space/eve/build_avx512/../include/eve/detail/function/simd/x86/friends.hpp:148:38: error: no matching conversion for functional-style cast from 'unsigned int' to 's_t' (aka 'eve::detail::mask_n<32>')
      return logical<wide<T,N,ABI>>{ s_t(~(v.storage().value ^ w.storage().value)) };
                                     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/dyarosh/space/eve/build_avx512/../include/eve/arch/cpu/base.hpp:134:22: note: in instantiation of function template specialization 'eve::detail::self_eq<unsigned short, eve::fixed<32>, eve::x86_512_>' requested here
      return detail::self_eq(v.self(),w.self());
                     ^
/home/dyarosh/space/eve/build_avx512/../include/eve/arch/cpu/base.hpp:139:16: note: in instantiation of member function 'eve::detail::operator==' requested here
      return v == Derived{w};
               ^
/home/dyarosh/space/eve/test/unit/api/wide/load/logical/load.hpp:343:32: note: in instantiation of function template specialization 'eve::detail::operator==<eve::logical<unsigned short>>' requested here
    TTS_EXPECT(eve::any(loaded == x));
                               ^
/home/dyarosh/space/eve/test/unit/api/wide/load/logical/load.hpp:331:1: note: in instantiation of function template specialization 'tts_function17<eve::wide<unsigned short, eve::fixed<32>, eve::x86_512_>>' requested here
TTS_CASE_TPL("Check load unsafe, logical", EVE_TYPE)
```

include/eve/detail/function/simd/x86/load.hpp
DenisYaroshevskiy: But it is x86 specific? I mean on ARM/POWER we should just load an not do any conversions.

include/eve/module/real/core/function/regular/generic/load.hpp
DenisYaroshevskiy: Doens't work. Asan still complains at you for doing this and I don't believe i can easily put an asan attribute on a lambda. Let's not

include/eve/detail/function/simd/x86/friends.hpp
jfalcou: ok

include/eve/detail/function/simd/x86/load.hpp
jfalcou: OK, there is an issue. Fixing outside

include/eve/module/real/core/function/regular/generic/load.hpp
jfalcou: OK

include/eve/memory/aligned_ptr.hpp
```diff
@@ -308,4 +309,17 @@ namespace eve
   struct  pointer_alignment<aligned_ptr<Type, Alignment>>
         : std::integral_constant<std::size_t,Alignment>
   {};
+
+  template <typename T, typename Cardinal>
+  EVE_FORCEINLINE auto previous_aligned_address(T* p, Cardinal /*width*/) noexcept
+  {
+    static constexpr std::size_t alignment = Cardinal() * sizeof(T);
```
jfalcou, line 316: not sure the static is required here.
`constinit` maybe if we really want to hammer down its pure compile-time nature ?
Does clang supports it ?

jfalcou:
LGTM

jfalcou (approved):

include/eve/memory/aligned_ptr.hpp
DenisYaroshevskiy: Oh - sorry missed it. `static constexpr` doesn't do anything. On the other hand plain `constexpr` tends to leave garbage on the stack for whatever reason.

Can't find an example but you can see it does not harm:
https://godbolt.org/z/46ojM9

PR #496, opened 2021-01-28
Move comparisons operators tests to unit.api

jfalcou:
== != < <= > >= are now tested in unit.api

### DIFF
--- a/include/eve/detail/function/simd/x86/make.hpp
+++ b/include/eve/detail/function/simd/x86/make.hpp
@@ -200,8 +200,7 @@ namespace eve::detail
   {
     if constexpr( !ABI::is_wide_logical )
     {
-      detail::as_mask_t<sizeof...(vs)> that{};
-
+      typename logical<wide<T,S,ABI>>::storage_type that{};
       [&]<std::size_t... N>(auto& v, std::index_sequence<N...>){ (( v |= vs?(1ULL<<N):0),...); }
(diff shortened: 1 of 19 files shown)


include/eve/detail/function/simd/x86/make.hpp
```diff
@@ -218,9 +217,9 @@ namespace eve::detail
   {
     if constexpr( !ABI::is_wide_logical )
     {
-      using i_t = typename detail::as_mask_t<ABI::bytes/sizeof(T)>::type;
-      detail::as_mask_t<ABI::bytes/sizeof(T)> that{ !!v ? ~i_t(0) : i_t(0) };
-      return that;
+      using s_t = typename logical<wide<T,S,ABI>>::storage_type;
+      constexpr auto m = typename s_t::type{0};
+      return s_t{ !!v ? ~m : m };
```
DenisYaroshevskiy, line 222: This got to be fixed as we discussed I think. With a test? :)

DenisYaroshevskiy (commented):
Ok other than the one comment.

include/eve/detail/function/simd/x86/make.hpp
jfalcou: Yes. I will do that rn

DenisYaroshevskiy (approved):

PR #499, opened 2021-01-29
FIX-495: fix load/store over-aligned

DenisYaroshevskiy:
I dunno if there is something similar I should do for load/store of logical. Or does it always forward to the wide?

Also not too sure about tests.

### DIFF
--- a/test/unit/api/wide/load/arithmetic/load.hpp
+++ b/test/unit/api/wide/load/arithmetic/load.hpp
@@ -13,6 +13,9 @@
 #include <eve/memory/aligned_ptr.hpp>
 #include <eve/function/load.hpp>
 #include <eve/wide.hpp>
+
+#include <array>
+#include <numeric>
 #include <vector>
 #include <list>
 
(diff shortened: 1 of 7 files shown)


test/unit/api/wide/store/arithmetic/store.hpp
```diff
@@ -55,4 +58,34 @@ TTS_CASE_TPL("Check store behavior to aligned arithmetic pointer", EVE_TYPE)
   eve::store(value, eve::as_aligned<algt>(&target[ 2 * EVE_CARDINAL ]) );
 
   TTS_ALL_EQUAL(target, ref);
-}
\ No newline at end of file
+}
+
+TTS_CASE_TPL("load for different alignment", EVE_TYPE )
```
jfalcou, line 63: s/load/store

test/unit/api/wide/store/arithmetic/store.hpp
```diff
@@ -55,4 +58,34 @@ TTS_CASE_TPL("Check store behavior to aligned arithmetic pointer", EVE_TYPE)
...
+{
+  using e_t = eve::element_type_t<T>;
```
jfalcou, line 65: this is given vy EVE_VALUE

test/unit/api/wide/load/arithmetic/load.hpp
```diff
@@ -293,3 +296,35 @@ TTS_CASE_TPL("Check load from range for wide", EVE_TYPE )
   T from_begin_end(ref_ptr.begin(), ref_ptr.end());
   TTS_EQUAL(from_begin_end, ref);
 }
+
+TTS_CASE_TPL("load for different alignment", EVE_TYPE )
+{
+  using e_t = eve::element_type_t<T>;
```
jfalcou, line 302: this is accessible via EVE_VALUE

jfalcou (changes requested):
LGTM, small stylistic issues.

jfalcou (approved):

PR #506, opened 2021-01-30
Move bitwise infix operator tests to unit.api.*

jfalcou:


include/eve/detail/function/simd/x86/arithmetic_compounds.hpp
```diff
@@ -334,4 +334,34 @@ namespace eve::detail
...
+        self = _mm512_div_pd(self, other);
+      }
+      else if constexpr( std::same_as<T, float> )
+      {
+        self = _mm512_div_ps(self, other);
+      }
+      else
+      {
+        auto that = self;
+        self      = aggregate([](auto a, auto b) { return a /= b; }, that, other);
```
DenisYaroshevskiy, line 361: What about `_mm512_div_epi64` and friends?
Or since they are truncating, they don't do the right thing?

What's with div by 0?

test/unit/api/bitwise/bit_and/regular/bit_and.hpp
```diff
@@ -0,0 +1,94 @@
+//==================================================================================================
```
DenisYaroshevskiy, line 1: Why did you have to write these tests? Didn't we have them for not 512?

include/eve/detail/function/simd/x86/arithmetic_compounds.hpp
jfalcou: Which intrinsics are you talking about ?
https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm512_div&expand=412,451,451,5867,2473,451&techs=AVX_512

test/unit/api/bitwise/bit_and/regular/bit_and.hpp
jfalcou: The functions were tested, not the operators.

include/eve/detail/function/simd/x86/arithmetic_compounds.hpp
DenisYaroshevskiy: These ones
https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm512_div&expand=412,451,451,5867,2473,451

PR #521, opened 2021-02-02
Fix #501: Top bits everywhere: tests + emulated and arm-v7 implementations

### LINKED ISSUE #501: [FEATURE] top_bits everywhere.
Even if it might not be the most optimal thing to always do, top_bits are the best solution in many cases.
We decided to port top_bits on all platforms.

simde port for reference: 
https://github.com/simd-everywhere/simde/blob/12069d720f43830ae9791e8b0f4c4fa3c88012a0/simde/x86/sse2.h#L3607

DenisYaroshevskiy:
Implementation of top_bits for emulated/arm-v7.

On emulated the tricky part is for when I don't fit into the 64 bits of a max number.
I just work with it as emulated.

For arm64 I think this is pretty good. Though chars have a shift, ah well, we need to measure.

For 128 it's roughly the same as for 64 except for chars.

I've tried a ton of different things, but I don't see it - so I left the 64 bit integer - let the compiler deal with the emulation.

The simd everywhere fellows have a movemask implementation for chars
https://github.com/simd-everywhere/simde/blob/12069d720f43830ae9791e8b0f4c4fa3c88012a0/simde/x86/sse2.h#L3607
Maybe that's the way to go - I dunno at the moment.
I guess will see what I write for first true, I only did 64 bits.

### DIFF
--- a/include/eve/detail/function/simd/arm/neon/movemask.hpp
+++ b/include/eve/detail/function/simd/arm/neon/movemask.hpp
@@ -1,0 +1,116 @@
+//==================================================================================================
+/**
+  EVE - Expressive Vector Engine
+  Copyright 2021 Joel FALCOU
+  Copyright 2021 Jean-Thierry LAPRESTE
+  Copyright 2021 Dens YAROSHEVKIY
+
+  Licensed under the MIT License <http://opensource.org/licenses/MIT>.
+  SPDX-License-Identifier: MIT
+**/
+//==================================================================================================
+#pragma once
+
+#include <eve/arch/logical.hpp>
+#include <eve/detail/meta.hpp>
+#include <eve/function/convert.hpp>
+
+#include <bit>
+
+namespace eve::detail
+{
+  template<typename T>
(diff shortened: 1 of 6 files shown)


include/eve/detail/function/simd/arm/neon/movemask.hpp
```diff
@@ -0,0 +1,116 @@
...
+#include <eve/function/convert.hpp>
+
+#include <bit>
+
+namespace eve::detail
+{
+  template<typename T>
+  EVE_FORCEINLINE std::uint32_t every_2nd_byte_arm64(T const& v) noexcept
+  {
+    using u8_8  = wide<std::uint8_t,  eve::fixed<8>, arm_64_>;
```
jfalcou, line 25: I am rather unconfortable by instanciating wide like that. This was historically a source of longer compile time.
This can be set as `typename T::template rebase<stduint8_t,fixed<8>>;`

DenisYaroshevskiy: Done

PR #522, opened 2021-02-03
Fix #463 - Update to use new SPY detection of neon/asimd discrimination

### LINKED ISSUE #463: [FEATURE] arm-v8 api
There are only neon64_ and neon128_ apis. 
There are some very useful instructions available only in arm-v8, like: `vminvq , vmaxvq`

We need a special test target and a current_api option.

jfalcou:
This patch adapts existing infrastructure to properly discriminate neon and asimd extensions set as such.
It fixes urgent issue when trying to compile for cortex A15 but doesn't remove all the #ifdef on aarch64 yet.

### DIFF
--- a/include/eve/arch/arm/as_register.hpp
+++ b/include/eve/arch/arm/as_register.hpp
@@ -18,7 +18,6 @@ namespace eve
 {
   template<typename T>
   struct logical;
-  struct neon64_;
   struct neon128_;
 }
 
@@ -37,8 +36,11 @@ namespace eve
       }
       else if constexpr( std::is_same_v<T,double> && Size::value <= 1 )
       {
-        if constexpr(spy::supports::aarch64_) return float64x1_t{};
-        else                                  return emulated_{};
+        #if defined(SPY_SIMD_IS_ARM_ASIMD)
+        return float64x1_t{};
+        #else
+        return emulated_{};
+        #endif
       }
       else if constexpr( std::is_integral_v<T> )
       {
@@ -72,8 +74,11 @@ namespace eve
       }
       else if constexpr( std::is_same_v<T,double> )
       {
-        if constexpr(spy::supports::aarch64_) return float64x2_t{};
-        else                                  return emulated_{};
+        #if defined(SPY_SIMD_IS_ARM_ASIMD)
+        return float64x2_t{};
+        #else
+        return emulated_{};
+        #endif
       }
       else if constexpr( std::is_integral_v<T> )
       {
--- a/include/eve/detail/spy.hpp
+++ b/include/eve/detail/spy.hpp
@@ -1,12 +1,13 @@
 //==================================================================================================
 /*
   SPY - C++ Informations Broker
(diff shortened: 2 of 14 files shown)


include/eve/arch/arm/as_register.hpp
```diff
@@ -37,8 +36,11 @@ namespace eve
       }
       else if constexpr( std::is_same_v<T,double> && Size::value <= 1 )
       {
-        if constexpr(spy::supports::aarch64_) return float64x1_t{};
-        else                                  return emulated_{};
+        #if defined(SPY_SIMD_IS_ARM_ASIMD)
+        return float64x1_t{};
```
DenisYaroshevskiy, line 40: What about 64 bit integers? Are they available?

include/eve/detail/spy.hpp
```diff
@@ -74,7 +75,7 @@ namespace spy
   constexpr inline auto ppc_    = detail::arch_info<detail::archs::ppc_>{};
   constexpr inline auto arm_    = detail::arch_info<detail::archs::arm_>{};
 }
-#include <ostream>
+#include <iosfwd>
```
DenisYaroshevskiy, line 78: Why? This doesn't seem right. You don't have any cpp files, do you?
You streaming operator requires `ostream` included

test/unit/api/wide/load/arithmetic/load.hpp
```diff
@@ -131,8 +131,8 @@ TTS_CASE_TPL("Check conditional load from pointer for wide", EVE_TYPE)
     auto i1 = eve::ignore_first(EVE_CARDINAL/4);
     auto kf = eve::keep_first(EVE_CARDINAL/4);
     auto kl = eve::keep_last(EVE_CARDINAL/4);
-    auto kb = eve::keep_between ( std::min(0L,EVE_CARDINAL/3)
-                                , std::max(0L,(EVE_CARDINAL*2)/3)
+    auto kb = eve::keep_between ( std::min<std::ptrdiff_t>(0L,EVE_CARDINAL/3)
+                                , std::max<std::ptrdiff_t>(0L,(EVE_CARDINAL*2)/3)
```
DenisYaroshevskiy, line 135: Specifing template parameters for standard template parameters is no the supported usecase as far as I understand. You can cast though

include/eve/detail/spy.hpp
jfalcou: This one escaped the cleaning of SPY. Will fix.

test/unit/api/wide/load/arithmetic/load.hpp
jfalcou: OK will cast

include/eve/arch/arm/as_register.hpp
jfalcou: like uint64_t ? Yes. Only thing we don't have is SIMD register of 64 bits integer

DenisYaroshevskiy (approved):

PR #529, opened 2021-02-05
FIX-455: arm-v7, all

DenisYaroshevskiy:
A better implementation of all on arm-v7.
For 64 bit the trick is `vpmin_u32` - the smallest uint32 should be still all ones (within the number of bits).
So min + get + test

I dunno enough about arm, maybe get just goes away since I get 0's element

For arm128 for anything but chars we just convert down a size.
So convert + min + get + test.
On x86 this is mmask + test so less by 2 instructions, but I don't see how I can do anything cooler on v7.

Unfortunately, for chars I didn't come up with anything smart, so split and merge.
Which also means I have to deal with ignore at the 16 byte level.

So split + and + min + get + test

I thought about using shift with insert to combine two halves
https://developer.arm.com/documentation/dui0472/m/using-neon-support/neon-intrinsics-for-shifts-with-insert

But it seems for that I need to split a vector, in which case I might just `and` it all.

I had to add a `mask_invert` to conditionals. Should be tested OK using `all` tests.
! doesn't seem to clear out the top half of the logical, + I am very much not sure compiler will see through it.

### DIFF
--- a/include/eve/module/real/algorithm/function/regular/simd/arm/neon/all.hpp
+++ b/include/eve/module/real/algorithm/function/regular/simd/arm/neon/all.hpp
@@ -17,35 +17,59 @@
 
 namespace eve::detail
 {
-  template <typename T, typename N, typename ABI, typename C>
-  EVE_FORCEINLINE bool all_arm_impl(logical<wide<T, N, ABI>> const &v0, C const & cond)
+  template <typename T, typename N, typename C>
+  EVE_FORCEINLINE bool all_arm_impl(logical<wide<T, N, arm_64_>> v0, C const & cond)
   {
-         if constexpr ( C::is_complete && !C::is_inverted ) return true;
-    else if constexpr ( std::same_as<ABI, arm_128_> )
+    using l_t = logical<wide<T, N, arm_64_>>;
+    if constexpr ( C::is_complete && !C::is_inverted ) return true;
+    else
     {
-      if constexpr ( sizeof( T ) >= 2 )
-      {
-        using half_e_t = make_integer_t<sizeof(T) / 2, unsigned>;
-        auto halved = eve::convert(v0, eve::as_<eve::logical<half_e_t>>{});
-        return all_arm_impl(halved, cond);
-      }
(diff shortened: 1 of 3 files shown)


include/eve/module/real/algorithm/function/regular/simd/arm/neon/all.hpp
```diff
@@ -11,48 +11,77 @@
...
+  template <typename T, typename N, typename C>
+  EVE_FORCEINLINE bool all_arm_impl(logical<wide<T, N, arm_128_>> v0, C const & cond)
+  {
+    using l_t = logical<wide<T, N, arm_128_>>;
+
+         if constexpr ( C::is_complete && !C::is_inverted ) return true;
+    else if constexpr ( sizeof( T ) >= 2 )
+    {
+      using half_e_t = make_integer_t<sizeof(T) / 2, unsigned>;
+      auto halved = eve::convert(v0, eve::as_<eve::logical<half_e_t>>{});
```
jfalcou, line 65: again, is this really `convert` we want ?

test/unit/module/real/algorithm/all/regular/all.hpp
```diff
@@ -137,11 +143,18 @@ TTS_CASE_TPL("Check eve::all[ignore]", EVE_TYPE)
...
         TTS_EXPECT_NOT(eve::all[eve::ignore_extrema_(i, j)](mask));
         TTS_EXPECT(eve::all[eve::ignore_extrema_(i + 1, j)](mask));
+
+        TTS_EXPECT_NOT(eve::all[eve::keep_between(i, T::static_size - j)](mask));
+        TTS_EXPECT(eve::all[eve::keep_between(i + 1, T::static_size - j)](mask));
+
         mask.set(i, true);
 
         mask.set(T::static_size - j - 1, false);
         TTS_EXPECT_NOT(eve::all[eve::ignore_extrema_(i, j)](mask));
```
jfalcou, line 153: ignore_extrema_ is supposed to be generated by `&&`. Do you think it is helpful we have it as a 1st class citizen?

PR #530, opened 2021-02-05
Fix 273 - Implement swizzle over wide/logical

jfalcou:
Implement swizzle of wide contents using the following api:

```
wide<T> x;
auto y =x[ pattern<1,2,4,5> ];
```

this maps to the proper shuffling intrinsics and detects some special cases.
The special cases can be later extended per architecture.


include/eve/detail/function/simd/arm/neon/swizzle.hpp
```diff
@@ -0,0 +1,86 @@
...
+        if constexpr( std::same_as<that_abi_t, arm_64_> )
+        {
+          using out_t       = typename that_t::template rebind<std::uint8_t,fixed<8>>;
+          return bit_cast(out_t{vtbl2_u8(lh, vget_low_u8 (mask))},as_<that_t>());
+        }
+        else
+        {
+          bytes_t that  = vcombine_u8 ( vtbl2_u8(lh, vget_low_u8 (mask))
+                                      , vtbl2_u8(lh, vget_high_u8(mask))
+                                      );
```
DenisYaroshevskiy, line 80: How would this work for when your left and right mix?

Like revrse.

Also you do: lh twice

DenisYaroshevskiy (commented):
this is way too much at the moment, will look later

include/eve/detail/function/simd/arm/neon/swizzle.hpp
```diff
@@ -0,0 +1,86 @@
...
+    // We're swizzling so much we aggregate the output
+    if constexpr( has_aggregated_abi_v<that_t> )
+    {
+      return aggregate_swizzle(v,q);
+    }
+    // Check for patterns
+    else if constexpr( !std::same_as<void, decltype(swizzle_pattern(neon128_{},v,q))> )
+    {
+      return swizzle_pattern(neon128_{},v,q);
+    }
```
DenisYaroshevskiy, line 38: This should probably be in the common code

include/eve/detail/function/simd/arm/neon/swizzle.hpp
```diff
@@ -0,0 +1,86 @@
...
+    {
+      if constexpr( std::same_as<ABI, arm_64_> )
+      {
+        using that_abi_t  = typename that_t::abi_type;
+        using bytes_t     = typename that_t::template rebind<std::uint8_t,fixed<8>>;
+        auto b0 = bit_cast(v,as_<bytes_t>());
+
+        if constexpr( std::same_as<that_abi_t, arm_64_> )
+        {
+          bytes_t that = vtbl1_u8(b0,as_bytes<wide<T,N,ABI>>(q,as_<bytes_t>()));
```
DenisYaroshevskiy, line 49: What about < 8 bytes

include/eve/detail/function/gather_high.hpp
```diff
@@ -0,0 +1,32 @@
...
+{
+  EVE_MAKE_CALLABLE(gather_high_, gather_high);
+
+  // Acts as a pattern itself
+  template<> struct is_pattern<tag::gather_high_> : std::true_type {};
+
+  // Premade pattern generator
+  template<std::ptrdiff_t N>
+  inline constexpr auto gather_high_n = fix_pattern<N>([](auto i, auto c) { return c/2 + i%(c/2); });
+}
```
DenisYaroshevskiy, line 26: Why gather high?

Hmm.

No, I don't understand.

For 8 elements:

```
4, 5, 6, 7, 4, 5, 6, 7
```
?

As we discussed I really think this should just bulk up to half the register size.
But ok.

include/eve/detail/function/gather_low.hpp
```diff
@@ -0,0 +1,33 @@
...
+  EVE_MAKE_CALLABLE(gather_low_, gather_low);
+
+  // Acts as a pattern itself
+  template<> struct is_pattern<tag::gather_low_> : std::true_type {};
+
+  // Premade pattern generator
+  template<std::ptrdiff_t N>
+  inline constexpr
+  auto gather_low_n = fix_pattern<N>( [](auto i, auto c) { return i < c/2 ? i : i-c/2;} );
+}
```
DenisYaroshevskiy, line 27: Should then gather_high_n be:

```
i >= c/2 ? i : i + c/2
```

I like this style better tbh

include/eve/detail/function/simd/common/patterns.hpp
```diff
@@ -0,0 +1,63 @@
...
+{
+  //================================================================================================
+  // Helpers for detecting patterns
+  //================================================================================================
+
+  // Detects <N,...,N> as a broadcast
+  template<int I0, int... I> inline constexpr bool is_broadcast = ((I0 !=-1) && ... && (I0==I));
+
+  // Detects <-1,...,-1> as zeroing
+  template<int... I> inline constexpr bool is_zero = ((I==-1) && ...);
```
DenisYaroshevskiy, line 27: Why zero and not just identity?
Is -1 zero? Or "I don't care"?

This -1 should be a meaningful enum

We seem to be inconsistent again about what we support where, no? is everything supports -1s?

include/eve/detail/function/patterns.hpp
```diff
@@ -0,0 +1,35 @@
...
+{
+  //================================================================================================
+  // Look to see if a given pattern is optimizable and returns the optimized function object
+  //================================================================================================
+  template<int... I> consteval auto find_optimized_pattern()
+  {
+          if constexpr( is_zero<I...>       ) return zero;
+    else  if constexpr( is_broadcast<I...>  ) return broadcast;
+    else  if constexpr( is_identity<I...>   ) return perform_identity{};
+    else  if constexpr( is_movhl<I...>      ) return gather_high;
```
DenisYaroshevskiy, line 31: Please be consistent with naming

include/eve/detail/function/simd/arm/neon/swizzle.hpp
```diff
@@ -0,0 +1,89 @@
...
+          return bit_cast(that,as_<that_t>());
```
DenisYaroshevskiy, line 44: How does swizzle change abi? I am confused:
can you swizzle to a different size/cardinality/type or not?
If you can, broadcast pattern and what not seem broken.
I don't see at all how swizzling to a different size works.

include/eve/detail/function/simd/common/broadcast.hpp
```diff
@@ -0,0 +1,42 @@
...
+#include <eve/pattern.hpp>
+#include <eve/forward.hpp>
+
+namespace eve::detail
+{
+  template<simd_value Wide>
+  EVE_FORCEINLINE auto broadcast_(EVE_SUPPORTS(cpu_), Wide v, auto Index) noexcept
+  {
+    return v[broadcast_n<Index,cardinal_v<Wide>>];
+  }
```
DenisYaroshevskiy, line 23: Hmm?
Is this a lookup fallback? Can we please not use the same `operator[]` with very very similar parameters at least in the implementation?

Also this is a questionable CPU impl.

Let's do Wide(v.get(Index)).

Ho is this suppose to change the ABI (which can happen according to generic shuffle)

include/eve/detail/function/simd/common/broadcast.hpp
```diff
@@ -0,0 +1,42 @@
...
+
+  template<typename Wide, shuffle_pattern Pattern>
+  EVE_FORCEINLINE auto broadcast_(EVE_SUPPORTS(cpu_), logical<Wide> v, Pattern p) noexcept
+  {
+    constexpr auto sz = Pattern::size(cardinal_v<logical<Wide>>);
+    using   that_t    = as_wide_t<logical<Wide>,fixed<sz>>;
+
+    return that_t{ v.get(p(0,sz)) };
```
DenisYaroshevskiy, line 31: What's pattern of 0/size? This is not an overload that makes sense

include/eve/detail/function/simd/common/gather_high.hpp
```diff
@@ -0,0 +1,45 @@
...
+#include <eve/detail/abi.hpp>
+#include <eve/pattern.hpp>
+#include <eve/forward.hpp>
+
+namespace eve::detail
+{
+  template<simd_value Wide> EVE_FORCEINLINE auto gather_high_(EVE_SUPPORTS(cpu_), Wide v) noexcept
+  {
+    // Well, we dunno how to do this, so we just reshuffle
+    return v[gather_high_n<cardinal_v<Wide>>];
```
DenisYaroshevskiy, line 22: ? It will end up here again or not? How is this helpful?

include/eve/detail/function/simd/common/gather_high.hpp
```diff
@@ -0,0 +1,45 @@
...
+
+  template<typename Wide, shuffle_pattern Pattern>
+  EVE_FORCEINLINE auto gather_high_( EVE_SUPPORTS(cpu_)
+                                  , logical<Wide> const& v, Pattern p
+                                  ) noexcept
+  {
+    constexpr auto sz = Pattern::size(cardinal_v<logical<Wide>>);
+    using   that_t    = as_wide_t<logical<Wide>,fixed<sz>>;
+
+    return bit_cast( v.mask()[p], as<that_t>() );
```
DenisYaroshevskiy, line 33: bits?

DenisYaroshevskiy (approved):

include/eve/detail/function/simd/common/gather_high.hpp
jfalcou: No, you won't end there. The [] will catch the proper gather_low call, the one below

include/eve/detail/function/simd/common/broadcast.hpp
jfalcou: p(0,sz) is computing the value of the pattern at index 0 for a given size

include/eve/detail/function/simd/common/broadcast.hpp
jfalcou: No. lookup is never call from there

include/eve/arch/cardinals.hpp
```diff
@@ -45,5 +45,10 @@ namespace eve
 
   template<std::ptrdiff_t Cardinal>
   inline constexpr fixed<Cardinal> const lane = {};
-}
 
+  //================================================================================================
+  // Template inline object for passing index within a wide
+  //================================================================================================
+  template<std::uint64_t N>
```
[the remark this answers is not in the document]
jfalcou: Fixed

include/eve/constant/zero.hpp
```diff
@@ -11,18 +11,35 @@
...
+    EVE_FORCEINLINE auto zero_(EVE_SUPPORTS(cpu_), Wide const&) noexcept
+    {
+      return Wide{0};
+    }
+
+    template<simd_value Wide, shuffle_pattern Pattern>
+    EVE_FORCEINLINE auto zero_(EVE_SUPPORTS(cpu_), Wide const&, Pattern) noexcept
+    {
+      return as_wide_t<Wide,fixed<Pattern::size(cardinal_v<Wide>)>>{0};
+    }
```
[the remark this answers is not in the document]
jfalcou: Fixed

include/eve/detail/function/simd/common/broadcast.hpp
jfalcou: Fixed

include/eve/detail/function/simd/common/broadcast.hpp
jfalcou: Fixed

include/eve/detail/function/patterns.hpp
jfalcou: Fixed

include/eve/detail/function/simd/common/patterns.hpp
jfalcou: Fixed

include/eve/detail/function/simd/common/swizzle.hpp
```diff
@@ -0,0 +1,94 @@
...
+      return that_t{ v[pattern_view<0,sz/2,sz>(q)], v[pattern_view<sz/2,sz,sz>(q)] };
+    }
+    else
+    {
+      return [=]<std::ptrdiff_t... I>(pattern_t<I...> const&)
+      {
+        return that_t{ (I == -1 ? T{0} : v.get(I))... };
+      }(p);
+    }
+  }
```
DenisYaroshevskiy, line 93: Will fix this in post I guess.

include/eve/detail/function/simd/common/swizzle_helpers.hpp
```diff
@@ -0,0 +1,78 @@
...
+    using type = typename Pack::value_type;
+
+    if constexpr( p.has_zeros(sz) )
+    {
+      // Turn pattern into 0/~0 then mask
+      auto const impl = [=](auto... i)
+                        {
+                          using i_t = as_integer_t<type,unsigned>;
+                          constexpr i_t nz = ~i_t(0), zz = i_t(0);
+                          return Pack(bit_cast((p(i,sz)==-1 ? zz : nz),as_<type>())...);
```
DenisYaroshevskiy, line 35: I rly rly think that -1 should be a enum somewhere.

include/eve/detail/function/simd/common/swizzle_helpers.hpp
```diff
@@ -0,0 +1,78 @@
...
+                        };
+
+      return v & apply<sz>(impl);
+    }
+    else
+    {
+      return v;
+    }
+  }
```
DenisYaroshevskiy, line 44: I tend to think this should be another operation.
And then the compiler should return:

```
operations<bound<broadcast, args>, bound<replace_zeroes, >>
```

Or smth like that

include/eve/detail/function/simd/x86/swizzle.hpp
```diff
@@ -0,0 +1,258 @@
...
+  //================================================================================================
+  // Unary swizzle - logical on AVX512 ABI, call general case on others
+  //================================================================================================
+  template<typename T, typename N, x86_abi ABI, shuffle_pattern Pattern>
+  EVE_FORCEINLINE auto basic_swizzle_ ( EVE_SUPPORTS(sse2_)
+                                      , logical<wide<T,N,ABI>> const& v, Pattern p
+                                      ) noexcept
+  {
+    if constexpr( current_api >= avx512 ) return to_logical((v.mask())[p]);
+    else                                  return basic_swizzle_(EVE_RETARGET(cpu_),v,p);
```
DenisYaroshevskiy, line 38: Why retarget cpu? I feel that just shuffle mask and bitcast back

include/eve/detail/function/simd/x86/swizzle.hpp
```diff
@@ -0,0 +1,258 @@
...
+    else if constexpr( current_api >= ssse3 )
+    {
+      using st_t    = typename that_t::storage_type;
+      using bytes_t = typename that_t::template rebind<std::uint8_t,fixed<16>>;
+      using i_t     = as_integer_t<wide<T,N,ABI>>;
+
+      return that_t ( (st_t)_mm_shuffle_epi8( bit_cast(v,as_<i_t>{}).storage()
+                                            , as_bytes<that_t>(q,as_<bytes_t>())
+                                            )
+                    );
```
DenisYaroshevskiy, line 82: And zeroes?

include/eve/detail/function/simd/x86/swizzle.hpp
```diff
@@ -0,0 +1,258 @@
...
+    }
+    else
+    {
+      if constexpr(sizeof(T) == 8)
+      {
+        using f_t         = as_floating_point_t<that_t>;
+        auto const vv     = bit_cast(v,as_<f_t>{});
+        constexpr auto m  = _MM_SHUFFLE2(q(1,2)&1, q(0,2)&1);
+        return bit_cast(process_zeros(f_t{_mm_shuffle_pd(vv,vv,m)},q),as_<that_t>{});
+      }
```
DenisYaroshevskiy, line 92: I feel this version is better for not sse2 as well.
Even if the compiler will do it, we probably better off doing this explicitly

include/eve/detail/function/simd/x86/swizzle.hpp
```diff
@@ -0,0 +1,258 @@
...
+    for(int i=c/2;i<c;++i)  if(idx[i] <  c/2 && idx[i] != -1) return false;
+
+    return true;
+  }
+
+  //================================================================================================
+  // SSE2-SSSE3 variant
+  //================================================================================================
+  template<typename T, typename N, x86_abi ABI, shuffle_pattern Pattern>
+  EVE_FORCEINLINE auto basic_swizzle_( EVE_SUPPORTS(sse2_), wide<T,N,ABI> const& v, Pattern const&)
```
DenisYaroshevskiy, line 61: Shouldn't this be x86_128 abi?

include/eve/detail/function/simd/x86/swizzle.hpp
```diff
@@ -0,0 +1,258 @@
...
+          return process_zeros(that_t{_mm_bsrli_si128(_mm_shufflehi_epi16(v,m),8)},q);
+        }
+        else
+        {
+          constexpr auto lp = pattern_view<0,4,8>(q);
+          constexpr auto hp = pattern_view<4,8,8>(q);
+
+                if constexpr( lp < 4 && hp >= 4) return process_zeros(that_t{v[lp],v[hp]},q);
+          else  if constexpr( hp < 4 && lp >= 4) return process_zeros(that_t{v[lp],v[hp]},q);
+          else  return basic_swizzle_(EVE_RETARGET(cpu_),v,q);
```
DenisYaroshevskiy, line 118: There should be a shuffle, shuffle, blend version here as well, right?

include/eve/detail/function/simd/x86/swizzle.hpp
```diff
@@ -0,0 +1,258 @@
...
+        else                                  return process_zeros(that_t{_mm_shuffle_epi32(v,m)},q);
+      }
+      else if constexpr(sizeof(T) == 2)
+      {
+        if constexpr( (sz < 8) && (q < 4) )
+        {
+          constexpr auto m  = _MM_SHUFFLE(q(3,4)&3, q(2,4)&3, q(1,4)&3, q(0,4)&3);
+          return process_zeros(that_t{_mm_shufflelo_epi16(v,m)},q);
+        }
+        else if constexpr( (sz < 8) && (q >= 4) )
```
DenisYaroshevskiy, line 106: That's what I meant by this bool for every element being confusing. I am entirely not sure what's going on here.

include/eve/detail/function/simd/x86/swizzle.hpp
```diff
@@ -0,0 +1,258 @@
...
+        {
+          constexpr auto m  = _MM_SHUFFLE(q(3,4)&3, q(2,4)&3, q(1,4)&3, q(0,4)&3);
+          return process_zeros(that_t{_mm_bsrli_si128(_mm_shufflehi_epi16(v,m),8)},q);
+        }
```
DenisYaroshevskiy, line 110: You do bsrli here but not for 8 byte shuffles. Why?

include/eve/detail/function/simd/x86/swizzle.hpp
```diff
@@ -0,0 +1,258 @@
...
+
+                if constexpr( lp < 4 && hp >= 4) return process_zeros(that_t{v[lp],v[hp]},q);
+          else  if constexpr( hp < 4 && lp >= 4) return process_zeros(that_t{v[lp],v[hp]},q);
+          else  return basic_swizzle_(EVE_RETARGET(cpu_),v,q);
+        }
+      }
+      else
+      {
+        return basic_swizzle_(EVE_RETARGET(cpu_),v,q);
+      }
```
DenisYaroshevskiy, line 124: Chars should convert to shorts, swizzle and convert back I believe.

include/eve/detail/function/simd/x86/swizzle.hpp
```diff
@@ -0,0 +1,258 @@
...
+    else if constexpr( width_in == 64 )
+    {
+      /// TODO: AVX512VBMI supports 128-bits permutexvar
+      if constexpr(width_out <= 16)       return basic_swizzle_(EVE_RETARGET(cpu_),v,q);
+      else if constexpr(width_out == 32)
+      {
+              if constexpr( q <  cd/2 )  return v.slice(lower_)[ q ];
+        else  if constexpr( q >= cd/2 )  return v.slice(upper_)[ slide_pattern<cd/2,sz>(q) ];
+        /// TODO: optimize using SSSE3 + binary shuffle
+        else                             return basic_swizzle_(EVE_RETARGET(cpu_),v,q);
```
DenisYaroshevskiy, line 157: Why retarget cpu then?

include/eve/detail/function/simd/x86/swizzle.hpp
```diff
@@ -0,0 +1,258 @@
...
+      }
+    }
+    else if constexpr( width_in == 32 )
+    {
+      if constexpr(width_out <= 16)
+      {
+              if constexpr( q <  (cd/2) ) return v.slice(lower_)[ q ];
+        else  if constexpr( q >= (cd/2) ) return v.slice(upper_)[ slide_pattern<cd/2,sz>(q) ];
+        /// TODO: optimize using SSSE3 + binary shuffle
+        else                              return basic_swizzle_(EVE_RETARGET(cpu_),v,q);
```
DenisYaroshevskiy, line 181: Again retarget cpu and not sse2

include/eve/pattern.hpp
```diff
@@ -0,0 +1,138 @@
...
+#include <eve/concept/pattern.hpp>
+
+namespace eve
+{
+  //================================================================================================
+  // Index-based pattern holder
+  //================================================================================================
+  template<std::ptrdiff_t... I> struct pattern_t
+  {
+    static inline constexpr auto size(int) noexcept { return sizeof...(I); }
```
DenisYaroshevskiy, line 24: Why does this have an int parameter?

include/eve/pattern.hpp
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    template<std::ptrdiff_t... O>
+    constexpr auto is_similar(pattern_t<I..., O...>) const noexcept requires(sizeof...(I)>0)
+    {
+      return true;
+    }
+
+    template<std::ptrdiff_t... O>
+    constexpr auto is_similar(pattern_t<O...>) const  noexcept { return false; }
```
DenisYaroshevskiy, line 59: Come on - just if constexpr. Not sure what is similar is.

include/eve/pattern.hpp
```diff
@@ -0,0 +1,138 @@
...
+
+    constexpr bool operator>  (std::ptrdiff_t n) const noexcept { return ((I  > n || I == -1) && ...); }
+    constexpr bool operator<  (std::ptrdiff_t n) const noexcept { return ((I  < n || I == -1) && ...); }
+    constexpr bool operator>= (std::ptrdiff_t n) const noexcept { return ((I >= n || I == -1) && ...); }
+    constexpr bool operator<= (std::ptrdiff_t n) const noexcept { return ((I <= n || I == -1) && ...); }
+
```
DenisYaroshevskiy, line 65: Please don't make this operator overloads.

include/eve/arch/cardinals.hpp
```diff
@@ -45,5 +45,10 @@ namespace eve
 
   template<std::ptrdiff_t Cardinal>
   inline constexpr fixed<Cardinal> const lane = {};
-}
 
+  //================================================================================================
+  // Template inline object for passing index within a wide
+  //================================================================================================
+  template<std::ptrdiff_t N>
+  inline constexpr auto index = std::integral_constant<std::ptrdiff_t,N>{};
```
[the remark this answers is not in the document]
jfalcou: Sure

include/eve/detail/function/patterns.hpp
```diff
@@ -0,0 +1,42 @@
...
+{
+  //================================================================================================
+  // Wrap a callable with its additional constexpr info
+  //================================================================================================
+  template<typename Callable, typename... Args>
+  struct bound
+  {
+    constexpr bound(Callable, Args...) {}
+    template<typename W> auto operator()(W w) const noexcept { return Callable{}(w, Args{}...); }
+  };
```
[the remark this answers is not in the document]
jfalcou: They're all constexpr so their type are enough. The only runtime parameters is the wide

include/eve/concept/pattern.hpp
```diff
@@ -0,0 +1,27 @@
...
+{
+  //================================================================================================
+  // Concept for swizzling pattern
+  //================================================================================================
+  template<typename T> concept shuffle_pattern = requires(T p)
+  {
+    { T::size(0)      };
+    { T::has_zeros(0) };
+    { T::validate(0)  };
+  };
```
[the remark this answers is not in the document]
jfalcou: Yes but it may be useless now

include/eve/detail/function/simd/common/patterns.hpp
```diff
@@ -0,0 +1,73 @@
...
+  //================================================================================================
+  struct perform_identity
+  {
+    friend std::ostream& operator<<(std::ostream& os, perform_identity)
+    {
+      return os << "perform_identity";
+    }
+
+    template<typename Wide, typename Pattern>
+    EVE_FORCEINLINE auto operator()(Wide v, Pattern p) const
```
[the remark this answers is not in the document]
jfalcou: a mistake

include/eve/detail/function/simd/common/patterns.hpp
```diff
@@ -0,0 +1,73 @@
...
+
+  //================================================================================================
+  // Local silly helper that doesn't warrant an actual functor
+  //================================================================================================
+  struct perform_identity
+  {
+    friend std::ostream& operator<<(std::ostream& os, perform_identity)
+    {
+      return os << "perform_identity";
+    }
```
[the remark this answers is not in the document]
jfalcou: sure

include/eve/detail/function/simd/x86/swizzle.hpp
jfalcou: becasue right now we don't have the shuffle for optimization

include/eve/detail/function/simd/x86/swizzle.hpp
jfalcou: because those are AVX register, you can not pass them to sse2

include/eve/pattern.hpp
jfalcou: I think it is now redundant. it was used in the old version but it needs removal

include/eve/pattern.hpp
jfalcou: is_similar checks two patterns starts with similar values like <1 2 5 5 > is similar to <1 2>

include/eve/pattern.hpp
jfalcou: Sure

include/eve/detail/function/simd/common/swizzle.hpp
```diff
@@ -0,0 +1,94 @@
...
+  // Emulation
+  //================================================================================================
+  template<typename T, typename N, typename ABI, shuffle_pattern Pattern>
+  EVE_FORCEINLINE auto basic_swizzle_(EVE_SUPPORTS(cpu_), wide<T,N,ABI> const& v, Pattern const& p)
+  {
+    constexpr auto cd = N::value;
+    constexpr auto sz = Pattern::size(cd);
+    using that_t      = as_wide_t<wide<T,N,ABI>,fixed<sz>>;
+
+    constexpr Pattern q = {};
```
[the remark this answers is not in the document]
jfalcou: p is a function parameter so I can't use it in constexpr context

include/eve/detail/function/simd/common/patterns.hpp
```diff
@@ -0,0 +1,73 @@
...
+      constexpr auto sz = Pattern::size(cd);
+      using that_t      = as_wide_t<Wide,fixed<sz>>;
+
+            if constexpr(sz     >= cd ) return that_t(v.storage());
+      else  if constexpr(cd/sz  == 2  ) return that_t(v.slice(lower_).storage());
+      else                              return v.slice(lower_)[p];
+    }
+  };
+
+  struct perform_zero
```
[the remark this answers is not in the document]
jfalcou: zero_swizzle is good

include/eve/detail/function/simd/common/swizzle_helpers.hpp
jfalcou: it is and we need to use it everywhere

include/eve/detail/function/simd/x86/swizzle.hpp
jfalcou: zeroes are handled directly by shuffle_epi8

From the Intel guide
```
__m128i _mm_shuffle_epi8 (__m128i a, __m128i b)
Synopsis
__m128i _mm_shuffle_epi8 (__m128i a, __m128i b)
#include <tmmintrin.h>
Instruction: pshufb xmm, xmm
CPUID Flags: SSSE3
Description
Shuffle packed 8-bit integers in a according to shuffle control mask in the corresponding 8-bit element of b, and store the results in dst.
Operation
FOR j := 0 to 15
	i := j*8
	IF b[i+7] == 1
		dst[i+7:i] := 0
	ELSE
		index[3:0] := b[i+3:i]
		dst[i+7:i] := a[index*8+7:index*8]
	FI
ENDFOR
```

include/eve/detail/function/simd/x86/swizzle.hpp
jfalcou: Maybe yes

include/eve/detail/function/simd/x86/swizzle.hpp
jfalcou: which is what the CPU logical version does. I call it to prevent code duplication.

include/eve/detail/function/simd/common/swizzle.hpp
```diff
@@ -0,0 +1,94 @@
...
+    }
+    // We're swizzling the first half of an aggregate
+    else if constexpr( has_aggregated_abi_v<wide<T,N,ABI>> && (Pattern{} < cd/2) )
+    {
+      return v.slice(lower_)[q];
+    }
+    // We're swizzling the second half of an aggregate
+    else if constexpr( has_aggregated_abi_v<wide<T,N,ABI>> && (Pattern{} >= cd/2) )
+    {
+      return v.slice(upper_)[ slide_pattern<cd/2,sz>(q) ];
```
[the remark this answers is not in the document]
jfalcou: Done

include/eve/detail/function/simd/common/swizzle.hpp
```diff
@@ -0,0 +1,94 @@
...
+
+    constexpr Pattern q = {};
+
+    // We're swizzling so much we aggregate the output
+    if constexpr( has_aggregated_abi_v<that_t> && !has_aggregated_abi_v<wide<T,N,ABI>> )
+    {
+      return aggregate_swizzle(v,q);
+    }
+    // We're swizzling the first half of an aggregate
+    else if constexpr( has_aggregated_abi_v<wide<T,N,ABI>> && (Pattern{} < cd/2) )
```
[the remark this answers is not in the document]
jfalcou: Changed to under/strictly_under and over/stricly_over

include/eve/concept/pattern.hpp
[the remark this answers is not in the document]
jfalcou: It is and was removed

include/eve/detail/function/broadcast.hpp
```diff
@@ -0,0 +1,37 @@
...
+  //================================================================================================
+  // Premade pattern generator
+  template<std::ptrdiff_t I, std::ptrdiff_t N>
+  inline constexpr auto broadcast_n = fix_pattern<N>( [](auto, auto) { return I; } );
+
+  namespace detail
+  {
+    //==============================================================================================
+    // Detects <N,...,N> as a broadcast
+    template<int I0, int... I> inline constexpr bool is_broadcast = ((I0 !=-1) && ... && (I0==I));
```
DenisYaroshevskiy, line 29: -1 left

include/eve/detail/function/simd/common/swizzle.hpp
```diff
@@ -0,0 +1,94 @@
...
+                      && pattern_view<sz/2,sz  ,sz>(q).strictly_under(cd/2)
+                      )
+    {
+      return that_t{ v[pattern_view<0,sz/2,sz>(q)], v[pattern_view<sz/2,sz,sz>(q)] };
+    }
+    else
+    {
+      return [=]<std::ptrdiff_t... I>(pattern_t<I...> const&)
+      {
+        return that_t{ (I == -1 ? T{0} : v.get(I))... };
```
DenisYaroshevskiy, line 90: -1

include/eve/detail/function/simd/common/swizzle_helpers.hpp
DenisYaroshevskiy: Still -1

include/eve/detail/function/simd/common/swizzle_helpers.hpp
```diff
@@ -0,0 +1,78 @@
...
+
+  //----------------------------------------------------------------------------------------------
+  // Index to bytes conversion
+  template<typename Pack, typename Shuffler, std::size_t... I>
+  EVE_FORCEINLINE constexpr auto as_bytes_impl(Shuffler p, std::index_sequence<I...> const &)
+  {
+    constexpr auto sz = Shuffler::size();
+    constexpr auto b = sizeof(typename Pack::value_type);
+
+    return values< (p(I/b,sz) == -1 ? 0xFF : p(I/b,sz)*b+I%b)... >{};
```
DenisYaroshevskiy, line 69: -1

include/eve/detail/function/simd/x86/swizzle.hpp
DenisYaroshevskiy: Please fix?

DenisYaroshevskiy (commented):
Some -1s marked

include/eve/detail/function/simd/common/swizzle.hpp
jfalcou: Fixed.

include/eve/detail/function/simd/common/swizzle_helpers.hpp
jfalcou: I like the idea it encodes the fact we need to process zeros into the type but I wonder if this si not a bit overkill.

include/eve/detail/function/simd/x86/swizzle.hpp
jfalcou: Done.

include/eve/detail/function/simd/x86/swizzle.hpp
jfalcou: Done

include/eve/detail/function/simd/x86/swizzle.hpp
jfalcou: What do you mean ?

include/eve/detail/function/simd/x86/swizzle.hpp
jfalcou: it won't help the case like pattern<1,0,3,2,...> where you need to shuffle at the byte level.
We can add those special cases later IMHO. The pattern to detect is non-trivial and will probably make the PR larger than required.

include/eve/detail/function/simd/common/swizzle_helpers.hpp
jfalcou: Also, process_zeros only use is non-complete swizzle arch. SSSE3 doesn't need it, neither ARM or PPC, so embedding it in the operation is maybe pessimistic.

include/eve/detail/function/simd/x86/swizzle.hpp
DenisYaroshevskiy: We'll do it later.

include/eve/detail/function/simd/x86/swizzle.hpp
DenisYaroshevskiy: ? It will. You just convert to wide<16> shorts from wide<16> chars. We will do that later.

DenisYaroshevskiy (approved):
I still think that shuffles for x86 should be reordered a bit but it's good enough.

PR #531, opened 2021-02-05
FIX-455: arm-v7, any

DenisYaroshevskiy:
A sister pr to https://github.com/jfalcou/eve/pull/529

### DIFF
--- a/include/eve/module/real/algorithm/function/regular/simd/arm/neon/any.hpp
+++ b/include/eve/module/real/algorithm/function/regular/simd/arm/neon/any.hpp
@@ -13,31 +13,55 @@
 #include <eve/concept/value.hpp>
 #include <eve/detail/implementation.hpp>
 #include <eve/function/bit_cast.hpp>
+#include <eve/function/convert.hpp>
 
 namespace eve::detail
 {
-  template<real_scalar_value T, typename N>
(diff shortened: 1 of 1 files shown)


include/eve/module/real/algorithm/function/regular/simd/arm/neon/any.hpp
```diff
@@ -13,31 +13,55 @@
...
+  template <typename T, typename N, typename C>
+  EVE_FORCEINLINE bool any_arm_impl(logical<wide<T, N, arm_128_>> v0, C const & cond)
+  {
+    using l_t = logical<wide<T, N, arm_128_>>;
+
+         if constexpr ( C::is_complete && !C::is_inverted ) return false;
+    else if constexpr ( sizeof( T ) >= 2 )
+    {
+      using half_e_t = make_integer_t<sizeof(T) / 2, unsigned>;
+      auto halved = eve::convert(v0, eve::as_<eve::logical<half_e_t>>{});
```
jfalcou, line 45: are you sure you want convert here and not bit_cast ?
Not even sure we have non-emulated code for this conversion.

DenisYaroshevskiy: I'm sure - I want vector narrow conversions: https://developer.arm.com/documentation/dui0472/m/using-neon-support/neon-intrinsics-for-converting-vectors

But not all of them seem to be in place, hmm:
I see `vmovn_u32` but not `vmovn_u16` for example.

PR #540, opened 2021-02-07
FIX-449: first true on top bits

DenisYaroshevskiy:
Messed up and merge the previous one without review.

In this pr I touched every file from that one except for tests.
Tests are here: https://github.com/jfalcou/eve/blob/05dfce179f73290e7ea644c86e0978dce10350de/test/unit/module/real/algorithm/first_true/regular/first_true.hpp#L1

### DIFF
--- a/include/eve/module/real/algorithm/function/regular/generic/first_true.hpp
+++ b/include/eve/module/real/algorithm/function/regular/generic/first_true.hpp
@@ -14,6 +14,7 @@
 #include <eve/concept/value.hpp>
 #include <eve/logical.hpp>
 #include <eve/detail/has_abi.hpp>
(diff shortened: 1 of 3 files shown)


include/eve/module/real/algorithm/function/regular/generic/first_true.hpp
```diff
@@ -25,34 +26,25 @@ namespace eve::detail
...
         if (v.get(first)) return first;
         ++first;
       }
+
       return {};
     }
+    else
+    {
+      // No ignore, we might luck out even if some elements should not be counted.
+      if ( !eve::any(v) ) return {};
```
jfalcou, line 44: Is it efficient ?

Issue #545, opened 2021-02-09
[BUG] strlen codegen for types bigger than char on arm.


DenisYaroshevskiy:
Goes bad: https://godbolt.org/z/M3Eh9K

I suspect - it's this convert call (and similar ones in other functions): https://github.com/jfalcou/eve/blob/23bd48ac9f2dcd798016b4adc502043f78eccd3e/include/eve/module/real/algorithm/function/regular/simd/arm/neon/any.hpp#L45

There is code for wide conversions but logical didn't forward to it properly I suspect.

Yeah - if I just call `convert` it goes to `movn` instruction.

PR #546, opened 2021-02-09
Optimize logical to logical convert

jfalcou:
This is an indirect fix for #545

### DIFF
--- a/include/eve/module/real/core/function/regular/generic/convert.hpp
+++ b/include/eve/module/real/core/function/regular/generic/convert.hpp
@@ -12,8 +12,10 @@
 
 #include <eve/as.hpp>
 #include <eve/concept/value.hpp>
+#include <eve/detail/function/bit_cast.hpp>
 #include <eve/detail/implementation.hpp>
 #include <eve/detail/has_abi.hpp>
+#include <eve/function/combine.hpp>
 #include <eve/traits/as_wide.hpp>
 
 namespace eve::detail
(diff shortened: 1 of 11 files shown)


include/eve/module/real/core/function/regular/generic/convert.hpp
```diff
@@ -50,4 +50,64 @@ namespace eve::detail
...
+        else if constexpr( has_aggregated_abi_v<logical<IN>> || has_aggregated_abi_v<out_t> )
+        {
+          // Extract bits then convert to proper wide type
+          auto[l,h] = v0.slice();
+          auto lb = convert(l.bits(), as_<typename logical<OUT>::bits_type>{});
+          auto hb = convert(h.bits(), as_<typename logical<OUT>::bits_type>{});
+
+          // Combine to_logical of them to fill up the bits if true
+          return bit_cast ( to_logical(eve::combine(lb,hb)), as_<out_t>{} );
+        }
```
DenisYaroshevskiy, line 103: Why do you need this?
You already dealt with not wide logical

test/unit/module/real/core/convert/double/convert.hpp
```diff
@@ -40,3 +40,22 @@ TTS_CASE_TPL("Check eve::convert behavior", EVE_TYPE)
...
+  TTS_EQUAL(eve::convert(eve::logical<T>(true)  , eve::as<eve::logical<double>>()), target_t(true) );
+  TTS_EQUAL(eve::convert(eve::logical<T>(false) , eve::as<eve::logical<double>>()), target_t(false) );
+
+#if defined(EVE_SIMD_TESTS)
+  eve::logical<T> mixed( [](auto i, auto) { return i%2 == 0;});
+  target_t        ref( [](auto i, auto) { return i%2 == 0;});
+
+  TTS_EQUAL(eve::convert(mixed, eve::as<eve::logical<double>>()), ref );
+#endif
+}
```
DenisYaroshevskiy, line 61: Just do two tests?

jfalcou: Same. This test is run in both mode scalar and simd so I'll end up havign to protect the lambda constructor

DenisYaroshevskiy (approved):

PR #553, opened 2021-02-13
Fix #548: better arm v7 codegen for first_true/any and somewhat all

### LINKED ISSUE #548: [FEATURE] first_true[ignore_none] shouldn't do a check that mmask is not 0.
If any fails here:
https://github.com/jfalcou/eve/blob/e352e4b058fbcb337f27f5a5868eb50ac3cd7f1c/include/eve/module/real/algorithm/function/regular/generic/first_true.hpp#L46

And there is nothing to ignore

We should not check for top_bits not being zero.

Probably implementation of first_true_guaranteed is in order.

DenisYaroshevskiy:
####  first_true

Turns out the issue with `first_true` unsafe is not so simple: `std::countr_zero` already checks for zero.
Also on gcc the codegen is either way OK, it's all clang problems.
(we could go to intrinsics but there is another way)

Clang can optimise that check away based on me doing a dereference: https://godbolt.org/z/W5b1vY
So I just added that.

gcc asm btw: https://github.com/DenisYaroshevskiy/unsq_eve/blob/93f862cd3c0f2e6cd6ff904118c293551b263bdc/disassemble/disassemble_arm.s#L49

####  any/ all

The split in two registers is a lot of asm. Much better option seems to be cast to `uint32_t` and test that against zero.
A much similar option is available for all but I need to test against FFFFF.

### DIFF
--- a/include/eve/module/real/algorithm/function/regular/simd/arm/neon/all.hpp
+++ b/include/eve/module/real/algorithm/function/regular/simd/arm/neon/all.hpp
@@ -29,12 +29,12 @@ namespace eve::detail
       if constexpr( N::value == 1 ) return v0.get(0);
       else
       {
-        using u32_2    = wide<std::uint32_t, eve::fixed<2>, arm_64_>;
-        auto as_uint32 = eve::bit_cast(v0.bits(), eve::as_<u32_2> {});
+        using u32_2 = typename wide<T, N, arm_64_>::template rebind<std::uint32_t, eve::fixed<2>>;
+        auto dwords = eve::bit_cast(v0.bits(), eve::as_<u32_2> {});
 
-        if constexpr( sizeof(T) * N() > 4u ) as_uint32 = vpmin_u32(as_uint32, as_uint32);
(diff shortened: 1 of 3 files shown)


include/eve/module/real/algorithm/function/regular/simd/arm/neon/all.hpp
```diff
@@ -68,8 +68,12 @@ namespace eve::detail
...
     {
       if constexpr ( !C::is_complete ) v0 = v0 || cond.mask_inverted(eve::as_<l_t>{});
-      auto [l, h] = v0.slice();
-      return all_arm_impl(l && h, ignore_none);
+
+      using u32_4 = typename wide<T, N, arm_128_>::template rebind<std::uint32_t, eve::fixed<4>>;
+      auto dwords = eve::bit_cast(v0, eve::as<u32_4>());
+
+      // not the same logic as for uint_32 plain so duplicated.
+      return all_arm_impl(dwords == std::numeric_limits<std::uint32_t>::max(), ignore_none);
```
jfalcou, line 76: why not just 0xFFFFFFFFUL ?

jfalcou (approved):
LGTM

PR #555, opened 2021-02-13
Add simple docs on how to build things

DenisYaroshevskiy:
I mostly wanted to have my build commands written down. But should be useful.

### DIFF
--- a/docs/developer/build_and_test.html
+++ b/docs/developer/build_and_test.html
@@ -13,60 +13,61 @@
 
 (insert ../crumbs.html here)
 
-# Building EVE: how to do local development.
+# Building EVE.
 
-*NOTE: this documentation was valid at [commit](https://github.com/jfalcou/eve/commit/2fe1048aa31eb3271666f1cca04bf2f5f34eeb6e) and might get outdated.* <br/>
+*NOTE: this documentation was valid at [commit](https://github.com/jfalcou/eve/commit/f631a04f5f6afefff132dadfbda48553400eb594).* <br/>
 
-There are two typical ways to build EVE: you can use your local compilers and tools or you can use the docker we use on CI.
+## Supported compilers/architectures.
 
-## Building for your architecture with your compiler.
+We work on extending and accomodating more compilers and architectures.
 
-*NOTE: this is for linux/mac os. I have not experimented with EVE on Windows.*
+Currently supported:
+  * clang > 10, gcc > 10
+  * C++20 is required
+  * x86: 64 bit, see2 - sse4.2, avx, avx2, avx-512(*)
+  * arm: little endian v7, little endian arm-v8 (aarch64)
+  * emulation for other platforms
 
-First, let's talk about developing locally. Let's say you just want to optimize something for your compiler.
-The tools we use are pretty standard: it's CMAKE and we support ctest.
-The two tricky pieces are: you need c++20 and you need to specify the architecture (march native works for me).
+(*) For avx-512 we decided on supporting skylake-avx512: F, CD, ER, PF.
 
-This is how I do cmake setup for my machine:
+We are interested in getting more usage for the library, do not hesitate to reach out if you need some other configuration. <br/>
+Some work has been done on power architecture and MSVC compilator support - we can prioritize those if necessary. <br/>
+
+## Building/using in your project.
+
+EVE is header only, so in order to use EVE you do not need any linking. <br/>
+However, since the sole purpose of EVE is to expose extra capabilities of your processor, <br/>
+you need to tell the compiler which architecture you compile for (via apropritate march options). <br/>
+
+We recommend running EVE's tests as part of your CI for your configuration since our CI capabilities are limited <br/>
+and some issues might slip through. <br/>
+
+In order to run all tests just:
 
 <script type="preformatted">
   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ bash
   # pwd eve
   mkdir build && cd build
-  cmake -G Ninja .. '-DCMAKE_CXX_FLAGS=-march=native -g -DEVE_NO_FORCEINLINE' -DCMAKE_EXPORT_COMPILE_COMMANDS=ON
+  cmake -G Ninja .. <your options>
+  ninja -j <num jobs> unit.exe
+  ctest
   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 </script>
 
-The only non standard option here is -DEVE_NO_FORCEINLINE. EVE force inlines aggressively
-but it is a massive drag on the compilation speed. So, when working on tests we disable it.
+## EVE_NO_FORCEINLINE
 
-Now we need to actually build and run tests.
-To build and run everything you type:<br/>
-*Adjust -j accordingly to your machine.*
+For various reasons EVE heavily uses forceinle which substantially slows down compilation.
+When compiling tests we recommend passing `-DEVE_NO_FORCEINLINE`
 
-<script type="preformatted">
-  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ bash
-  # pwd eve/build
-  time ninja -j 16 unit.exe && ctest -j 16
-  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-</script>
+## Working on EVE.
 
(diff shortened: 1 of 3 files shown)


docs/developer/build_and_test.html
```diff
@@ -0,0 +1,188 @@
...
+ -->
+
+ <meta charset="utf-8" lang="en"><style class="fallback">body{visibility:hidden;}</style>
+ **Expressive Vector Engine**
+
+(insert ../crumbs.html here)
+
+# Building EVE: how to do local development.
+
+*NOTE: this documentation was valid at [commit](https://github.com/jfalcou/eve/commit/2fe1048aa31eb3271666f1cca04bf2f5f34eeb6e) and might get outdated.* <br/>
```
jfalcou, line 18: as any file, I guess this kind of disclaimer is not useful.

docs/developer/build_and_test.html
```diff
@@ -0,0 +1,188 @@
...
+  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ bash
+  # pwd eve
+  mkdir build_avx512 && cd build_avx512
+  cmake -G Ninja .. '-DCMAKE_CXX_FLAGS=-march=skylake-avx512 -DEVE_NO_FORCEINLINE'
+  ninja -j 16 unit.real.algorithm.exe
+  sde -- ./unit/unit.real.algorithm.all.regular.simd.exe
+  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+</script>
+
+ctest doesn't work with sde for me and I didn't need it yet.
```
jfalcou, line 162: This is where we should talk about the CMAKE_CROSSCOMPILING_CMD option

docs/developer/build_and_test.html
```diff
@@ -0,0 +1,188 @@
...
+  ninja -j 16 unit.meta.exe
+  ../cmake/toolchain/run_arm.sh ./unit/unit.meta.as_integer.exe
+  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+</script>
+
+## AVX-512 Docker Example
+
+Now, avx-512 is not supported on my machine directly, so I use sde to run it.
+It is supported in our build envirenment so we do not use it for CI.
+I also use the EVE_OPTIONS override to march.
```
jfalcou, line 129: s/I/We

docs/developer/build_and_test.html
```diff
@@ -0,0 +1,188 @@
...
+  cmake -G Ninja .. -DCMAKE_TOOLCHAIN_FILE=cmake/toolchain/gcc.arm.cmake
+  ninja -j 16 unit.meta.exe
+  ../cmake/toolchain/run_arm.sh ./unit/unit.meta.as_integer.exe
+  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+</script>
+
+## AVX-512 Docker Example
+
+Now, avx-512 is not supported on my machine directly, so I use sde to run it.
+It is supported in our build envirenment so we do not use it for CI.
```
jfalcou, line 128: spurrious details

docs/developer/build_and_test.html
```diff
@@ -0,0 +1,188 @@
...
+How we build for certain platforms can be found in: [config](https://github.com/jfalcou/eve/blob/2fe1048aa31eb3271666f1cca04bf2f5f34eeb6e/.circleci/config.yml#L14)<br/>
+Toolchains suppresses the default CMAKE compiler options, so we have EVE_OPTIONS macro you can define instead.<br/>
+For example for different x86 we use the same toolchain, so you need to use EVE_OPTIONS to specify the march.
+
+## Arm v7 Docker Example.
+
+<script type="preformatted">
+  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ bash
+  # pwd eve
+  # in docker to the left I see root@<commit>:
```
jfalcou, line 117: We can actually just copy/paste an actual output

docs/developer/build_and_test.html
```diff
@@ -0,0 +1,188 @@
...
+  # pwd eve
+  dockhere jfalcou/compilers:latest
+  # cd where eve is again
+  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+</script>
+
+5: Now you can use all of the EVE's prepackaged compilers.
+
+You probably want to use our toolchain files for it.
+How we build for certain platforms can be found in: [config](https://github.com/jfalcou/eve/blob/2fe1048aa31eb3271666f1cca04bf2f5f34eeb6e/.circleci/config.yml#L14)<br/>
```
jfalcou, line 108: We can make a table so for each Arch we display the toolchain + option as hints

docs/developer/build_and_test.html
```diff
@@ -0,0 +1,188 @@
...
+<script type="preformatted">
+  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ bash
+  sudo snap install docker
+  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+</script>
+
+2: Make docker work.
+
+Get to the point when "docker run hello-world" is succesfull.
+Google solved my problems reasonably fast
```
jfalcou, line 80: Hahaha 'put under the carpet' award of the year ^^
I think point 1 and 2. can be merged into a "Here is a good tutorial on settign up docker, you should follow tseps up to the poitn where .... works".

docs/developer/build_and_test.html
```diff
@@ -0,0 +1,188 @@
...
+
+<script type="preformatted">
+  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ bash
+  # pwd eve/build
+  time ninja -j 16 unit.exe && ctest -j 16
+  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+</script>
+
+This will run all of the EVE unittests. Which maybe a bit much, depending on what you do.
+
```
jfalcou, line 55: There we should explain the test naming scheme and target network. I an write that if you want

docs/developer/build_and_test.html
```diff
@@ -0,0 +1,188 @@
...
+<script type="preformatted">
+  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ bash
+  # pwd eve
+  mkdir build && cd build
+  cmake -G Ninja .. '-DCMAKE_CXX_FLAGS=-march=native -g -DEVE_NO_FORCEINLINE' -DCMAKE_EXPORT_COMPILE_COMMANDS=ON
+  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+</script>
+
+The only non standard option here is -DEVE_NO_FORCEINLINE. EVE force inlines aggressively
```
jfalcou, line 40: This is to be made a note.

docs/developer/build_and_test.html
```diff
@@ -0,0 +1,188 @@
...
+  mkdir build_avx512 && cd build_avx512
+  cmake -G Ninja .. -DCMAKE_TOOLCHAIN_FILE=../cmake/toolchain/clang.x86.cmake -DEVE_OPTIONS="-march=skylake-avx512"
+  ninja -j 16 unit.meta.exe
+  sde -- ./unit/unit.meta.as_integer.exe
+  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+</script>
+
+## Other saved examples
+
+I prefer to not work from docker when I don't have to.
```
jfalcou, line 144: s/I/We

jfalcou (changes requested):
I think the contents is OK but:
 - we need to work on the tone first/3rd person style so it feels like an actual doc page more than a blog page
 - We need a clear separation between building locally, building in the Docker. I think we should start with building locally using g++ or/and Clang, setting up options, and getting a grasp of the test target naming schemes. Then, we introduce the docker and how you can run an arbitrary compiler INSIDE the docker. If needed, I can write the part where we tell the people how to call the Docker compiler from outside the docker.
 - For the bash example, just copy paster w/e appears in your actual bash.
 - We need to explain the CMAKE_CROSSCOMPILING_CMD option at some point

docs/developer/build_and_test.html
DenisYaroshevskiy: I think this is a good practice in docs.

PR #557, opened 2021-02-13
Add ARM v7 targets to CI

jfalcou:


test/unit/meta/has_abi.cpp
```diff
@@ -15,34 +15,38 @@
...
-  TTS_EXPECT( eve::has_native_abi_v<eve::wide<double>>                );
-  TTS_EXPECT( eve::has_native_abi_v<eve::logical<eve::wide<double>>>  );
+  TTS_EXPECT( eve::has_native_abi_v<eve::wide<float>>                );
+  TTS_EXPECT( eve::has_native_abi_v<eve::logical<eve::wide<float>>>  );
 #endif
 
-  TTS_EXPECT( eve::has_native_abi_v<double>                           );
-  TTS_EXPECT( eve::has_native_abi_v<eve::logical<double>>             );
+  TTS_EXPECT( eve::has_native_abi_v<float>                           );
+  TTS_EXPECT( eve::has_native_abi_v<eve::logical<float>>             );
```
DenisYaroshevskiy, line 32: Eeh? This seems like a duplicate

test/unit/meta/has_abi.cpp
```diff
@@ -15,34 +15,38 @@
...
 
-  TTS_EXPECT_NOT( eve::has_aggregated_abi_v<eve::wide<double>>                );
-  TTS_EXPECT_NOT( eve::has_aggregated_abi_v<eve::logical<eve::wide<double>>>  );
+  TTS_EXPECT_NOT( eve::has_aggregated_abi_v<eve::wide<float>>                );
+  TTS_EXPECT_NOT( eve::has_aggregated_abi_v<eve::logical<eve::wide<float>>>  );
 
-  TTS_EXPECT_NOT( eve::has_aggregated_abi_v<double>                           );
-  TTS_EXPECT_NOT( eve::has_aggregated_abi_v<eve::logical<double>>             );
+  TTS_EXPECT_NOT( eve::has_aggregated_abi_v<float>                           );
+  TTS_EXPECT_NOT( eve::has_aggregated_abi_v<eve::logical<float>>             );
```
DenisYaroshevskiy, line 51: No idea what this is

test/unit/meta/has_abi.cpp
jfalcou: We check that scalar have indeed a native ABI.

test/unit/meta/has_abi.cpp
jfalcou: We check scalars don't get assigned the aggregated ABI
Those are old tests, but they check out ;)

DenisYaroshevskiy (approved):

Issue #538, opened 2021-02-07
[FEATURE] top bits, arm-v8 and power


DenisYaroshevskiy:

DenisYaroshevskiy:
Will do as part of: https://github.com/jfalcou/eve/issues/541

PR #571, opened 2021-02-19
FIX-#538 support unsafe(load[ignore])

DenisYaroshevskiy:
Fixing the api gap.

We can argue wether this testing is good enough but it seems to be able to catch most of the bugs I can think off.

Generally speaking this API is a bit weird and I don't know when you'd need it, but it makes some sense in the context of just thread sanitizer.

`load[ignore]` for aggregated does not work on avx-512 and avx2 as well I suspect. I put an forward to memcpy for now and reported a bug.

### DIFF
--- a/include/eve/module/real/core/function/regular/generic/load.hpp
+++ b/include/eve/module/real/core/function/regular/generic/load.hpp
@@ -11,9 +11,10 @@
 #pragma once
 
 #include <eve/concept/memory.hpp>
-#include <eve/function/unsafe.hpp>
 #include <eve/detail/implementation.hpp>
 #include <eve/detail/spy.hpp>
+#include <eve/function/unsafe.hpp>
+#include <eve/wide.hpp>
 #include <type_traits>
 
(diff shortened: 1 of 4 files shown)


include/eve/module/real/core/function/regular/generic/load.hpp
```diff
@@ -11,9 +11,10 @@
 #pragma once
 
 #include <eve/concept/memory.hpp>
-#include <eve/function/unsafe.hpp>
 #include <eve/detail/implementation.hpp>
 #include <eve/detail/spy.hpp>
+#include <eve/function/unsafe.hpp>
+#include <eve/wide.hpp>
```
jfalcou, line 17: Normally, we include `eve/forward.hpp`, is there a reason we need the full blown wide here ?

jfalcou: OK issue is concept related, will fix later

jfalcou (approved):

PR #580, opened 2021-02-21
FIX-576: all dependency bug

DenisYaroshevskiy:
This specific dependency issue occurs from assert in lookup.hpp

included from `include/eve/arch/cpu/base.hpp`

```
#include <eve/detail/function/lookup.hpp>
```

Finished simplification of all here. We will do the rest separately I think.

### DIFF
--- a/include/eve/detail/top_bits.hpp
+++ b/include/eve/detail/top_bits.hpp
@@ -76,10 +76,33 @@ struct top_bits
       }
     }
 
+    static constexpr bool is_cheap_impl()
+    {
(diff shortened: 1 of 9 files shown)


include/eve/detail/top_bits.hpp
```diff
@@ -76,11 +76,34 @@ struct top_bits
       }
     }
 
+    static constexpr bool is_cheap_impl()
```
jfalcou, line 79: if we're 500% sure this is only used in constexpr mode, use consteval

jfalcou (approved):
LGTM. Let's merge

PR #595, opened 2021-02-28
AVX512: all tests compile

DenisYaroshevskiy:
I started by just adding sde toolchain.
But then I fixed all of the compilation for unit.exe.

I still cannot do ctest for all the tests but at least everything compiles.


include/eve/detail/function/bitmask.hpp
```diff
@@ -8,6 +8,7 @@
 #pragma once
 
 #include <eve/arch.hpp>
+
```
jfalcou, line 11: why the line break ?

include/eve/function/add.hpp
```diff
@@ -14,18 +14,4 @@ namespace eve
...
 
-#include <eve/arch.hpp>
 #include <eve/module/real/core/function/regular/generic/add.hpp>
-
-#if defined(EVE_HW_X86)
-#  include <eve/module/real/core/function/regular/simd/x86/add.hpp>
-#endif
-
-#if defined(EVE_HW_POWERPC)
-#  include <eve/module/real/core/function/regular/simd/ppc/add.hpp>
```
jfalcou, line 25: why removing this ??

include/eve/function/bit_mask.hpp
```diff
@@ -12,7 +12,7 @@
 namespace eve
 {
   EVE_MAKE_CALLABLE(bit_mask_, bit_mask);
+  EVE_MAKE_CALLABLE(to_mask_, to_mask);
```
jfalcou, line 15: why are we piggybacking bit_mask here ?

PR #606, opened 2021-03-04
FIX-602: reverse

DenisYaroshevskiy:
This is a stub to do the reverse first issue

### DIFF
--- a/include/eve/detail/function/reverse.hpp
+++ b/include/eve/detail/function/reverse.hpp
@@ -1,0 +1,39 @@
+//==================================================================================================
+/**
+  EVE - Expressive Vector Engine
+  Copyright : EVE Contributors & Maintainers
+  SPDX-License-Identifier: MIT
+**/
+//==================================================================================================
+#pragma once
+
+#include <eve/detail/overload.hpp>
+#include <array>
+
+namespace eve
+{
+  EVE_MAKE_CALLABLE(reverse_, reverse);
+
+  namespace detail
+  {
+    template <int N, int ...I> inline constexpr bool is_reverse = []
+    {
+      std::array idxs {I...};
+      for (int i = 0; i != static_cast<int>(idxs.size()); ++i) {
+        if (idxs[i] != (static_cast<int>(idxs.size()) - i - 1))
+          return false;
+      }
+      return static_cast<int>(idxs.size()) == N;
+    }();
+  }
+}
+
+#include <eve/detail/function/simd/common/reverse.hpp>
+
+#if defined(EVE_HW_X86)
+#  include <eve/detail/function/simd/x86/reverse.hpp>
+#endif
+
+#if defined(EVE_HW_ARM)
+#  include <eve/detail/function/simd/arm/neon/reverse.hpp>
+#endif
--- a/include/eve/detail/function/simd/arm/neon/reverse.hpp
+++ b/include/eve/detail/function/simd/arm/neon/reverse.hpp
@@ -1,0 +1,31 @@
+//==================================================================================================
+/**
+  EVE - Expressive Vector Engine
+  Copyright : EVE Contributors & Maintainers
+  SPDX-License-Identifier: MIT
+**/
+//==================================================================================================
+#pragma once
+
+#include <eve/detail/abi.hpp>
+#include <eve/detail/function/swap_adjacent_groups.hpp>
+
+namespace eve::detail
+{
+  template<typename T, typename N, arm_abi ABI>
+  EVE_FORCEINLINE wide<T,N,ABI> reverse_( EVE_SUPPORTS(neon128_), wide<T,N,ABI> v) noexcept
+  {
+    constexpr bool one_instruction_basic_swizzle = (current_api >= asimd) || std::same_as<ABI, arm_64_>;
+
+    // two reverse instructions seem better than one table lookup.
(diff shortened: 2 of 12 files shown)


include/eve/detail/function/simd/x86/swizzle.hpp
```diff
@@ -190,7 +190,7 @@ namespace eve::detail
           if constexpr(match(c,category::int64x4 ,category::uint64x4) )
           {
             auto vc = bit_cast(v, as_<as_floating_point_t<that_t>>{});
-            return bit_cast( v[q] , as(s));
+            return bit_cast( basic_swizzle(vc, q) , as(s));
```
jfalcou, line 193: if you do this, then the swizzle won't be optimized by the find_optimized_pattern.
Is there a reason you changed that ?

include/eve/detail/function/simd/arm/neon/swap_adjacent_groups.hpp
```diff
@@ -53,7 +53,12 @@ namespace eve::detail
         else  if constexpr( c == category::float64x2  )
         {
           if constexpr(current_api > asimd) return vextq_f64(v,v,1);
-          else                              return that_t(v.get(1),v.get(0));
+          else
+          {
+            auto u64 = vreinterpretq_u64_f64(v);
```
jfalcou, line 58: does it gives better codegen than before ?

include/eve/detail/function/simd/arm/neon/reverse.hpp
```diff
@@ -0,0 +1,31 @@
...
+  template<typename T, typename N, arm_abi ABI>
+  EVE_FORCEINLINE wide<T,N,ABI> reverse_( EVE_SUPPORTS(neon128_), wide<T,N,ABI> v) noexcept
+  {
+    constexpr bool one_instruction_basic_swizzle = (current_api >= asimd) || std::same_as<ABI, arm_64_>;
+
+    // two reverse instructions seem better than one table lookup.
+    if constexpr ( N() >= 8 && one_instruction_basic_swizzle) return reverse_(EVE_RETARGET(cpu_), v);
+    else
+    {
+      if constexpr ( N() >= 16 ) v = eve::swap_adjacent_groups(v, lane<8>);
```
jfalcou, line 24: can you clarify the algorithm here ? Why is it enough ?

include/eve/detail/function/reverse.hpp
```diff
@@ -0,0 +1,39 @@
...
+//==================================================================================================
+#pragma once
+
+#include <eve/detail/overload.hpp>
+#include <array>
+
+namespace eve
+{
+  EVE_MAKE_CALLABLE(reverse_, reverse);
+
```
jfalcou, line 16: We usually provide a pattern generator too.

jfalcou (changes requested):

include/eve/detail/function/simd/x86/swizzle.hpp
DenisYaroshevskiy: that was the point. 

We don't call main swizzle from the parts. Specifically that caused me a recursion when I fall back on basic swizzle.

include/eve/detail/function/simd/arm/neon/swap_adjacent_groups.hpp
DenisYaroshevskiy: At least it should. I mean it's one instruction.

include/eve/detail/function/reverse.hpp
DenisYaroshevskiy: Oh - I saw that. What's the point?

include/eve/detail/function/simd/arm/neon/swap_adjacent_groups.hpp
DenisYaroshevskiy: Oh...
So there is smth nuts going on:
https://godbolt.org/z/vqnr93fh6

We generate some ultra hot garbage. I have no idea what's up.

And apparently these casts should only be avaliable on aarch64,  but our gcc arm builds them fine. What's up with that?
https://godbolt.org/z/vd6M86Px3

include/eve/detail/function/simd/arm/neon/reverse.hpp
DenisYaroshevskiy: Well, you can't really reverse 16 chars with one instruction.
But you can:
Swap two 64 bit groups => reverse each halve (so - swap 32 bit groups => reverse each halve etc)

jfalcou: ok

include/eve/detail/function/reverse.hpp
jfalcou: so we have a shortcut for user `v[ reverse_n<8> ]`

include/eve/detail/function/simd/x86/swizzle.hpp
jfalcou: but then if this casted to float can be optimized, it wont ??

DenisYaroshevskiy: If we are here, it's already couldn't be optimized.

include/eve/detail/function/reverse.hpp
DenisYaroshevskiy: Discussed. I think we shoul delete it everywhere.

jfalcou (approved):

PR #620, opened 2021-03-12
[WIIP] New version of tests system

jfalcou:
Use a type+lambda based system to generate tests in a way we don't depend on macros and CMake file generation.
This temporary PR is there to check the compile-time and the scalability of the system.

Will only touch api and internals for now.


test/unit/internals/top_bits.cpp
```diff
@@ -0,0 +1,389 @@
...
+
+      actual = top_bits<logical>(eve::keep_first(logical::static_size - i));
+      TTS_EQUAL(expected, eve::detail::to_logical(actual));
+    }
+  }
+};
+
+EVE_TEST_BED("Check top_bits from ignore_*"
+            , eve::test::simd::all_types
+            , eve::test::generate(eve::test::no_data)
```
DenisYaroshevskiy, line 349: What I would suggest maybe is to, when you have no generation, just use another macro?
This is a touch awkward I think.

test/unit/api/arithmetic.cpp
```diff
@@ -0,0 +1,144 @@
...
+            );
+
+//==================================================================================================
+// scalar (*) wide tests
+//==================================================================================================
+auto mixed_tests = []<typename T>( auto& runtime, bool verbose, auto const&
+                                , T const& a0, T const& a1, T const&
+                                )
+{
+  TTS_EQUAL( (a0 + 3), T([&](auto i, auto) { return a0.get(i) + 3; }) );
```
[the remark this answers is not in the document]
jfalcou: Solved

test/unit/internals/top_bits.cpp
jfalcou: Solved

DenisYaroshevskiy (approved):

PR #622, opened 2021-03-24
more new style tests

jtlap:


jfalcou (changes requested):
Actually, just need a proper rebase over develop and you need to remove the old tests files of the tests you changed.

PR #623, opened 2021-03-26
Enable Github Actions for CI

jfalcou:


jfalcou:
Question: what do we do with Circle CI now ?

PR #626, opened 2021-03-30
displacing horner and correcting 5 tests

jtlap:
displaced also legendre gegenbauer hermite laplace


jfalcou (changes requested):
Tests must be moved to unit/polynomial/

jfalcou (approved):

PR #641, opened 2021-04-17
Small fix on clang-format so that it doesn't destroy everything

jfalcou:


jfalcou:
Main issue is the destructuration of our if else constexpr nests. Bug has been filled at LLVM, waiting for a fix this week.

PR #653, opened 2021-04-23
Fix bad code generation when returning wide in non-inline context

jfalcou:
We were a bit too enthusiastic when manually writing copy and default constructor for SIMD value types. Using =default fix the codegen issue.

Code:

```
wide<float> h()
{
  return wide<float>(5.74f);
}
```

Before:
```
h():
        vmovaps xmm0, XMMWORD PTR .LC1[rip]
        mov     rax, rdi
        vmovaps XMMWORD PTR [rdi], xmm0
        ret
```

After:
```
h():
        vmovaps xmm0, XMMWORD PTR .LC0[rip]
        ret
```

### DIFF
--- a/include/eve/arch/cpu/logical_wide.hpp
+++ b/include/eve/arch/cpu/logical_wide.hpp
@@ -60,8 +60,7 @@ namespace eve
     //==============================================================================================
     // Constructors
     //==============================================================================================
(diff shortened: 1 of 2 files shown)


DenisYaroshevskiy (approved):
I like that we'd be able to easier see asm on godbolt.

PR #655, opened 2021-04-23
More conversions for aggregate types

jfalcou:
IN PROGRESS
 - conversions from aggregate to x86_128 cases
 - missing inverted cases in convert_128

### DIFF
--- a/include/eve/module/real/core/function/regular/simd/x86/convert_128.hpp
+++ b/include/eve/module/real/core/function/regular/simd/x86/convert_128.hpp
@@ -78,6 +78,105 @@ namespace eve::detail
       }
     }
     //==============================================================================================
+    // Convert to 64 bits integer
+    //==============================================================================================
+    else if constexpr( std::is_integral_v<Out> && (sizeof(Out) == 8) )
+    {
+      //============================================================================================
+      // 8 -> 64 bits
+      //============================================================================================
+      if constexpr( std::is_integral_v<In> && (sizeof(In) == 1))
+      {
+        if constexpr( current_api >= sse4_1 )
+        {
+          if constexpr( std::is_signed_v<In> && (N::value <= 2) )
+          {
+            return _mm_cvtepi8_epi64(v0);
+          }
+          else if constexpr( !std::is_signed_v<In>  && (N::value <= 2) )
+          {
+            return _mm_cvtepu8_epi64(v0);
+          }
+          else
+          {
+            return convert_(EVE_RETARGET(simd_), v0, tgt);
+          }
+        }
+        else
+        {
+          return convert_(EVE_RETARGET(simd_), v0, tgt);
+        }
+      }
+      //============================================================================================
+      // 16 -> 64 bits
+      //============================================================================================
+      if constexpr( std::is_integral_v<In> && (sizeof(In) == 2))
+      {
+        if constexpr( current_api >= sse4_1 )
+        {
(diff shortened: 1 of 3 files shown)


include/eve/module/real/core/function/regular/simd/x86/convert_128.hpp
```diff
@@ -78,6 +78,105 @@ namespace eve::detail
...
+        {
+          if constexpr( std::is_signed_v<In> && (N::value <= 2) )
+          {
+            return _mm_cvtepi16_epi64(v0);
+          }
+          else if constexpr( !std::is_signed_v<In>  && (N::value <= 2) )
+          {
+            return _mm_cvtepu16_epi64(v0);
+          }
+          else
```
DenisYaroshevskiy, line 125: Is there a specific reason not to use short ifs?

include/eve/module/real/core/function/regular/simd/x86/convert_128.hpp
```diff
@@ -78,6 +78,105 @@ namespace eve::detail
...
+        {
+          return convert_(EVE_RETARGET(simd_), v0, tgt);
+        }
+      }
+      //============================================================================================
+      // 32 -> 64 bits
+      //============================================================================================
+      if constexpr( std::is_integral_v<In> && (sizeof(In) == 4))
+      {
+        if constexpr( current_api >= sse4_1 )
```
DenisYaroshevskiy, line 140: We have this code already, no? 
https://github.com/jfalcou/eve/blob/0e5d657d7ea13344e1ec3e31287efc3179e11b84/include/eve/module/real/core/function/regular/simd/x86/convert_128.hpp#L98

Let's reuse that

include/eve/module/real/core/function/regular/simd/x86/convert_128.hpp
```diff
@@ -173,97 +272,42 @@ namespace eve::detail
...
-          }
-          else if constexpr( !std::is_signed_v<In>  && (N::value <= 2) )
-          {
-            return _mm_cvtepu16_epi64(v0);
-          }
-          else
-          {
-            return convert_(EVE_RETARGET(simd_), v0, tgt);
-          }
+          v0 &= static_cast<In>(0x00FFU);
```
DenisYaroshevskiy, line 289: Why the mask? Did you test and you need it?
I mean we have pack signed and pack unsigned.

include/eve/module/real/core/function/regular/simd/x86/convert_aggregated.hpp
```diff
@@ -0,0 +1,82 @@
...
+          h &= static_cast<In>(0x0000FFFFU);
+
+          return _mm_packus_epi32(l,h);
+        }
+        else if constexpr( current_api >= ssse3 )
+        {
+          auto l16 = bit_cast(l, as_<wide<Out>>());
+          auto h16 = bit_cast(h, as_<wide<Out>>());
+
+          return l16[pattern<0,2,4,6,-1,-1,-1,-1>] | h16[pattern<-1,-1,-1,-1,0,2,4,6>];
```
DenisYaroshevskiy, line 52: Why not just call the propr intrinsic?
I don't think we should use engage all of the insane shuffle machinery internally unless we have to (like in sort or smth).

DenisYaroshevskiy (commented):
There is a lot of things happening. What's going on with the basic shuffle/

PR #658, opened 2021-04-24
Remove useless swizzle generator

jfalcou:


include/eve/detail/function/swap_adjacent_groups.hpp
```diff
@@ -18,32 +18,33 @@ namespace eve
...
+  namespace detail
   {
-    // List all possible swags for a current size
-    constexpr auto sz = sizeof...(I);
-    constexpr auto x = []<std::size_t... N>( std::index_sequence<N...> )
+    //==============================================================================================
+    // Classify a pattern as a swag
+    template<std::ptrdiff_t G, std::ptrdiff_t N>
+    inline constexpr
+    auto swap_adjacent_groups_n = fix_pattern<N> ( [](auto i, auto)
```
DenisYaroshevskiy, line 27: Can we call these `*_pattern`? `_n` doesn't tell me anything.

jfalcou: sure

PR #661, opened 2021-04-26
Implementation for slide_left swizzle

jfalcou:


include/eve/detail/function/simd/x86/slide_left.hpp
```diff
@@ -0,0 +1,147 @@
...
+    {
+      if constexpr( std::same_as<ABI,x86_128_>)
+      {
+        /// Mask noises from smaller sized registers
+        if constexpr(N::value < expected_cardinal_v<T,x86_128_>)
+        {
+          wide<T,expected_cardinal_t<T,x86_128_>> w = v.storage();
+          w &= keep_first(N::value).mask(as(w)).bits();
+          v = w.storage();
+        }
```
DenisYaroshevskiy, line 33: I don't think I understand this. This is FFFF ...00000? For 8 chars it will be top 8 elements, that are zero. And you clear all of the rest. No?

include/eve/detail/function/simd/x86/slide_left.hpp
```diff
@@ -0,0 +1,147 @@
...
+
+          // Slide lower parts using _mm_alignr_epi8
+          constexpr auto sz = sizeof(T)* Shift;
+          using byte_t = typename wide<T,N,ABI>::template rebind<std::uint8_t,fixed<16>>;
+
+          byte_t bytes = _mm_alignr_epi8(bit_cast(h,as_<byte_t>{}), bit_cast(l,as_<byte_t>{}), sz);
+
+          // Shift everything in place
+          return wide<T,N,ABI>{bit_cast(bytes, as(h0)),h0};
+        }
```
DenisYaroshevskiy, line 120: For avx2 consider: 

```
template <std::size_t byte_shift>
__m256i alignr256_emulated(__m256i a, __m256i b) {
  __m256i b1_a0 = _mm256_permute2x128_si256(a, b, 0x03);
  if constexpr (byte_shift == 16)
    return b1_a0;
  else if constexpr (byte_shift < 16)
    return _mm256_alignr_epi8(b1_a0, b, byte_shift);
  else
    return _mm256_alignr_epi8(a, b1_a0, byte_shift - 16);
}

wide<T,N,ABI> zeroes(0);
alignr256_emulated<-Shift * sizeof(T)>(v, zeroes);
```

So setzero + shuffle + _mm256_alignr_epi8 - 3 instructions (that if your algorithm doesn't have a zero from somewhere already)

include/eve/detail/function/simd/x86/slide_left.hpp
```diff
@@ -0,0 +1,147 @@
...
+
+#include <eve/detail/implementation.hpp>
+#include <eve/conditional.hpp>
+
+namespace eve::detail
+{
+  template<real_scalar_value T, typename N, x86_abi ABI, std::ptrdiff_t Shift>
+  EVE_FORCEINLINE wide<T,N,ABI> slide_left_ ( EVE_SUPPORTS(sse2_)
+                                            , wide<T,N,ABI> v, index_t<Shift>
+                                            ) noexcept
```
DenisYaroshevskiy, line 18: I think here we might be missing aggregated case for 2 registers.
Those can be done fairly well with alignr, similar to what you do for avx.

On the other hand, requirement on zero might be tricky.

DenisYaroshevskiy (commented):
Very couple of things, I need to be more alive to do it.

include/eve/detail/function/simd/x86/slide_left.hpp
jfalcou: This can be done mater as a shuffle. Also requirement on zero can be done like with if_else by detecting we pas eve::zero.

include/eve/detail/function/simd/x86/slide_left.hpp
jfalcou: Will test this

include/eve/detail/function/simd/x86/slide_left.hpp
jfalcou: I keep the first N element with respect to the endianess. This is masking the data in the _m128 but outside the logical size of the wide so when shifting occurs, no garbage data are pulled in.

FYI, this has been tested to be required.

jfalcou: Now maybe I can express it in a better way I concede that to you.

include/eve/detail/function/simd/arm/neon/slide_left.hpp
```diff
@@ -0,0 +1,67 @@
...
+      constexpr auto c = categorize<that_t>();
+      that_t z{0};
+
+      // Mask noises from smaller sized registers
+      if constexpr(N::value < expected_cardinal_v<T,arm_64_>)
+      {
+        wide<T,expected_cardinal_t<T,arm_64_>> w = v.storage();
+        w &= keep_first(N::value).mask(as(w)).bits();
+        v = w.storage();
+      }
```
DenisYaroshevskiy, line 43: I think that you should mask after the shift.
I believe that as is it leaves the shifted values instead of 0s.

DenisYaroshevskiy (commented):
Still didn't look through instructions sorry.

include/eve/detail/function/simd/arm/neon/slide_left.hpp
jfalcou: I mask over the expected size, this means I mask values physically in the storage but not in the logical view.
Maybe it'll be simpler to do it the other way. Will try.

jfalcou: Shift have been simplified

include/eve/detail/function/simd/x86/slide_left.hpp
jfalcou: Better codegen, adopted for AVX2

include/eve/detail/function/simd/x86/slide_left.hpp
```diff
@@ -0,0 +1,156 @@
...
+          else if constexpr(Shift == 3)
+          {
+            auto const s0 = _mm256_blend_ps(w, _mm256_permute2f128_ps(w,w,0x81), 0x77);
+            return bit_cast(f_t(_mm256_permute_ps( s0, 0x93)), as(v));
+          }
+          else if constexpr(Shift == 4)
+          {
+            return bit_cast(f_t( _mm256_permute2f128_ps(w,w,0x81)),as(v));
+          }
+          else if constexpr(Shift == 5)
```
DenisYaroshevskiy, line 97: For 5 and more I think it should be enough to:

_mm256_permute2f128_ps and _mm256_shuffle_ps

include/eve/detail/function/simd/x86/slide_left.hpp
```diff
@@ -0,0 +1,156 @@
...
+            return bit_cast(f_t(_mm256_shuffle_pd(s0,f_t(0),1)), as(v));
+          }
+        }
+        else if constexpr(sizeof(T) == 4)
+        {
+          // Seven optimal shuffles for sliding left by 1->7 doctors want you to not know about
+          using f_t = as_floating_point_t<wide<T,N,ABI>>;
+          auto const w = bit_cast(v, as_<f_t>{});
+
+          if constexpr(Shift == 1)
```
DenisYaroshevskiy, line 78: `_mm256_shuffle_ps`([a b c d e f g h], [0 0 0 0 a b c d]) for the < 4
`_mm256_shuffle_ps`([a b c d e f g h], [e f g k 0 0 0 0] ) for > 4

include/eve/detail/function/simd/x86/slide_left.hpp
```diff
@@ -0,0 +1,133 @@
...
+          else                              return _mm256_bsrli_epi128(bv, offset - 16);
+        }
+        else
+        {
+          constexpr auto shifted_bytes = sizeof(T)* Shift;
+          using f_t = typename wide<T,N,ABI>::template rebind<float>;
+          auto const w  = bit_cast(v, as_<f_t>{});
+          auto const s0 = _mm256_permute2f128_ps(w,w,0x81 );
+
+          if constexpr(shifted_bytes == 28)
```
DenisYaroshevskiy, line 59: Hmm.

Can we please do this as `7 * 4` ? I think it'd be easier to read.

include/eve/detail/function/simd/x86/slide_left.hpp
```diff
@@ -0,0 +1,133 @@
...
+          }
+          else  if constexpr(shifted_bytes == 8 )
+          {
+            return bit_cast(f_t(_mm256_shuffle_ps(w, s0, 0x4e)), as(v));
+          }
+          else  if constexpr(shifted_bytes == 4 )
+          {
+            return bit_cast(f_t(_mm256_permute_ps(_mm256_blend_ps(w, s0, 0x11), 0x39)), as(v));
+          }
+          else
```
DenisYaroshevskiy, line 87: I'm pretty sure that you can do 2 instructions and not 3 for anything divisible by 4 as I mentioned previously.

include/eve/detail/function/simd/x86/slide_left.hpp
```diff
@@ -0,0 +1,133 @@
...
+          {
```
DenisYaroshevskiy, line 88: By using zero extend for shuffles > 128 you can reduce the work here I believe.

include/eve/detail/function/simd/x86/slide_left.hpp
jfalcou: Done

DenisYaroshevskiy (approved):

PR #662, opened 2021-04-26
Implementation for slide_right swizzle

jfalcou:


include/eve/detail/function/simd/x86/slide_right.hpp
```diff
@@ -0,0 +1,160 @@
...
+            return bit_cast(bvi,as(v));
+          }
+          else  if constexpr(offset <  16)
+          {
+            return bit_cast(i_t(_mm256_alignr_epi8 (vi, bvi, 16 - offset)) , as(v));
+          }
+          else
+          {
+            return bit_cast(i_t(_mm256_bslli_epi128(bvi, offset - 16)), as(v));
+          }
```
DenisYaroshevskiy, line 53: Maybe format these in one line each like for shift_left?

include/eve/detail/function/simd/x86/slide_right.hpp
```diff
@@ -0,0 +1,160 @@
...
+                                                      );
+
+              // permutes so we have [ <0 .. 0> 0 .. v0 ]
+              auto g = bit_cast(h, as_<f_t>{});
+              return bit_cast(_mm256_permute2f128_ps(g,g,0x01), as(v));
+            }
+            else
+            {
+              // Which side are we ?
+              constexpr bool phase = Shift > (N::value/2);
```
DenisYaroshevskiy, line 107: Can this ever be true?
We are in `avx` so total size is 32 bytes.
Therefore - `shifted_bytes > 16` covers `Shift > (N::value / 2)`

include/eve/detail/function/simd/x86/slide_right.hpp
jfalcou: Sure

DenisYaroshevskiy (approved):
LGTM if passes. `Shift>(N::value/2)) ? N::value/2` is left is `slide_left`, consider removing that maybe.

Issue #562, opened 2021-02-14
[FEATURE] minimal power on CI


DenisYaroshevskiy:
At the moment there is a confusion as to which power we support, and at least power 7 is broken.

Proposal: make the power we support (which is probably little endian 64 bit v8) run on ci and fix what doesn't work.
At some point maybe port to different powers (though probably this is less priority than let's say AVX-512).
Don't touch until needed.

PR #663, opened 2021-04-26
Supports test for PPC

jfalcou:
As per #562 , here is the state of PPC support:

 - minimal version is POWERPC8 with VSX enabled
 - all targets compile and pass


DenisYaroshevskiy:
Ok - so ppc8 little endian, ppc9 we will add later?

PR #685, opened 2021-05-08
FIX-681: ptr iterator

DenisYaroshevskiy:
Pointer based iterators, first implementation.

Added some tests and a basic README - doing proper documentation is too expensive while the algorithms are under active development.


test/unit/algo/ptr_iterator.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+#include <eve/algo/ptr_iterator.hpp>
+
+#include "iterator_conept_test.hpp"
+
+#include <array>
+#include <numeric>
+
+EVE_TEST_TYPES("Check unaligned_ptr_iterator", algo_test::selected_types)
+<typename T>(eve::as_<T>)
```
DenisYaroshevskiy, line 19: Messed up the test name.
Will fix here if you have comments.
Otherwise - will do in the next patch.

include/eve/algo/concepts.hpp
```diff
@@ -0,0 +1,19 @@
...
+  SPDX-License-Identifier: MIT
+**/
+//==================================================================================================
+#pragma once
+
+#include <concepts>
+
+namespace eve::algo
+{
+  template <typename T>
```
jfalcou, line 14: I like when we have both the `_t` and the normal traits. Unless this one is for implementation only, maybe we can add the missing one? If it's internal, then this is fine.

include/eve/algo/ptr_iterator.hpp
```diff
@@ -0,0 +1,107 @@
...
+  struct aligned_ptr_iterator : operations_with_distance, forward_to_unaligned
+  {
+    using cardinal = Cardinal;
+    using aligned_ptr_type = eve::aligned_ptr<T, Cardinal{}() * sizeof(T)>;
+
+    aligned_ptr_iterator();
+    explicit aligned_ptr_iterator(aligned_ptr_type ptr) : ptr{ptr} {}
+
+    auto unaligned() const { return unaligned_ptr_iterator<T, Cardinal>{ptr.get()}; }
+    aligned_ptr_iterator& operator+=(std::ptrdiff_t n) { ptr += n; return *this; }
```
jfalcou, line 46: should we `EVE_ASSERT` that n is preserving alignment?

test/unit/algo/algo_test.hpp
```diff
@@ -0,0 +1,26 @@
+//==================================================================================================
```
jfalcou, line 1: test.hpp is maybe enough for the name, we know we're in algo/

test/unit/algo/iterator_conept_test.hpp
```diff
@@ -0,0 +1,130 @@
+//==================================================================================================
```
jfalcou, line 1: iterator_conCept_test.hpp maybe ?

test/unit/algo/algo_test.hpp
DenisYaroshevskiy: Cannot do that.

Just "test.hpp" will collide with our normal "test.hpp"

"algo/test.hpp" does not work either.

Feel free to try later.

include/eve/algo/concepts.hpp
DenisYaroshevskiy: Had to do like this:

```
  template <typename T>
  using unaligned_t = decltype(std::declval<T>().unaligned());

  template <typename T>
  struct unaligned
  {
    using type = unaligned_t<T>;
  };
```
Otherwise the compilation breaks for some bizarre reason.

jfalcou: This is fine

test/unit/algo/algo_test.hpp
jfalcou: ok

jfalcou (approved):

Issue #690, opened 2021-05-08
[FEATURE] Get all of the components for `any` in place.


DenisYaroshevskiy:
The goal is: the minimal `any_of` is implemented.

No alignment tricks, unrolling or knowing about precise end.
No concepts.
Should work for 2 `std::vector::iterator`, raw pointers and eve iterators passed in.
Aligned_ptr to pointer is not part of this.

The following components should be implemented:

* `canonicalise_range(f, l)` - returns `unaligned_ptr_iterator` for pointers and `contiguous_iterator`. At this task - does not have to be extensible.
* `for_each_iteration(traits, f, l, delegate)`  - the basic iteration pattern. No unrolling support.
* `bool any_of(f, l, predicate)` - the first algorithm interface.

traits at this point just an empty struct.

Well maybe unrolling support.
No - that requires a sketch for traits and I don't know how that should look really.

DenisYaroshevskiy:
Fixed in #694

PR #694, opened 2021-05-08
FIX-#690: any_of basics

DenisYaroshevskiy:
This is a sort of sketchy patch, none of the components are complete.
But I think we are better off getting the ball rolling by having smth that works end to end that we can enhance later.

### DIFF
--- a/include/eve/algo/any_of.hpp
+++ b/include/eve/algo/any_of.hpp
@@ -1,0 +1,58 @@
+//==================================================================================================
+/**
+  EVE - Expressive Vector Engine
+  Copyright : EVE Contributors & Maintainers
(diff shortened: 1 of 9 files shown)


include/eve/algo/any_of.hpp
```diff
@@ -0,0 +1,58 @@
...
+  struct any_of_
+  {
+    struct default_traits {};
+
+    template <typename P>
+    struct delegate
+    {
+      explicit delegate(P p) : p(p) {}
+
+      bool small_step(auto it, eve::relative_conditional_expr auto ignore)
```
jfalcou, line 28: why not just step ?

jfalcou (commented):
LGTM

PR #700, opened 2021-05-11
Mark some header sources as SYSTEM

jfalcou:


DenisYaroshevskiy:
Fixes the warning for sure, checking if the ticket can be closed completly

Issue #691, opened 2021-05-08
[FEATURE] Get mechanism for `traits` in `algo`.


DenisYaroshevskiy:
We need some sort of general purpose mechanism for passing a bag of traits in `algo` to control the algorithm execution.

Things like:
* unrolling
* assume that the start is aligned
* assume that the distance between start and end is divisible by cardinality

We need to support overrides and defaults. We need to support getting traits without pain.

@jfalcou suggested using named parameters which sounds like a good idea.

I'd prefer to have a function:

`eve::algo::traits` to create and pass around traits.

jfalcou:
the function can just be an aggregator of named parameters. Raberu need to gatehr all keyword into a settings object, this is what this raits function can do:

```c++
template<rbr::keyword_parameters... Options>
auto traits(Options const&... opts)
{
  return rbr::settings(opts...);
}
```

example call:

```c++
eve::algo::soem_algo( ptr1, ptr2, alg::traits(  unroll = 4, aligned=true ) );
```

DenisYaroshevskiy:
This looks awesome.
Can we do it as a first parameter same as std policies?
The last is usually the lambda

```
eve::algo::any_of( algo::traits( unroll = 1), f, l, [] (auto x){
   return 3 < x && x < 5;
});

PR #711, opened 2021-05-16
FIX-#691: traits and basic unrolling.

DenisYaroshevskiy:
This is an implementation of a very basic unrolling.

Since we do not yet allow to override the traits with default ones, there is no logic to added a default unroll if the user didn't provide one.

The testing is so-so, we have a separate story to design an better testing (faster with more coverage and weird types).


include/eve/algo/unroll_op.hpp
```diff
@@ -0,0 +1,24 @@
...
+#include <eve/eve.hpp>
+
+namespace eve::algo
+{
+  template<std::ptrdiff_t how_much, typename Op>
+  EVE_FORCEINLINE auto unroll_op(Op op)
+  {
+    return [&]<std::ptrdiff_t... idxs>(std::integer_sequence<std::ptrdiff_t, idxs...>)
+    {
+      return (op(eve::index<idxs>) || ...);
```
jfalcou, line 19: why || ? Unrollign statement can be done with operator, in fold expression.

include/eve/algo/any_of.hpp
```diff
@@ -7,37 +7,57 @@
...
+        eve::logical test = p(eve::load[ignore](it));
         res = eve::any[ignore](test);
         return res;
       }
 
+      template <typename I, std::size_t size>
+      EVE_FORCEINLINE bool unrolled_step(std::array<I, size> arr)
+      {
+        // TODO: Issue #710 to do the proper reduciton;
+        auto f = arr.begin(), l = arr.end();
```
jfalcou, line 41: eve/detail/meta.hpp had a for_ that helps unrolling, maybe use it here ?

DenisYaroshevskiy: Didn't touch this.
This one is a very temp one, we'll fix it in a little bit.

PR #712, opened 2021-05-16
FIX-710: array_map/array_reduce

DenisYaroshevskiy:
Common utilities to deal with arrays of unrolled stuff.

If you think this is correct, I can probably manually implement for all of the individual arrays up to 4 to reduce template instantiation.

### DIFF
--- a/include/eve/algo/array_utils.hpp
+++ b/include/eve/algo/array_utils.hpp
@@ -1,0 +1,47 @@
+//==================================================================================================
+/**
+  EVE - Expressive Vector Engine
+  Copyright : EVE Contributors & Maintainers
+  SPDX-License-Identifier: MIT
+**/
+//==================================================================================================
+#pragma once
+
+#include <eve/eve.hpp>
+
+#include <array>
+#include <utility>
+
+namespace eve::algo
+{
+  namespace detail
+  {
+    template <std::size_t f, std::size_t l, typename T, std::size_t N, typename Op>
+    EVE_FORCEINLINE constexpr T array_reduce_impl(std::array<T, N> const& x, Op op)
+    {
+      if constexpr ((l - f) == 1) return x[f];
+      else
+      {
+        constexpr std::size_t n = l - f;
+        T left_sum  = array_reduce_impl<f, f + n / 2>(x, op);
+        T right_sum = array_reduce_impl<f + n / 2, l>(x, op);
+        return op(left_sum, right_sum);
+      }
+    }
+  }
+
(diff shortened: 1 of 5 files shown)


include/eve/algo/array_utils.hpp
```diff
@@ -0,0 +1,47 @@
...
+        T right_sum = array_reduce_impl<f + n / 2, l>(x, op);
+        return op(left_sum, right_sum);
+      }
+    }
+  }
+
+  template <typename T, std::size_t N, typename F>
+  EVE_FORCEINLINE constexpr auto array_map(std::array<T, N> const& x, F f)
+  {
+    return [&]<std::size_t ...i>(std::index_sequence<i...>){
```
jfalcou, line 36: Maybe reuse std::apply here ?
`return std::apply( [](auto... xs) { return std::array{xs...};}, x);`

test/unit/algo/array_utils.cpp
```diff
@@ -0,0 +1,42 @@
...
+  constexpr std::array<int, 4> expected = { 2, 4, 6, 8 };
+  constexpr auto actual = eve::algo::array_map(input, [](char x) -> int { return x * 2; });
+  TTS_CONSTEXPR_EXPECT(expected == actual);
+}
+
+TTS_CASE("eve.algo.array_utils array_map")
+{
+  {
+    constexpr std::array in = {1, 2, 3, 4, 5};
+    constexpr int actual = eve::algo::array_reduce(in, std::plus<>{});
```
jfalcou, line 28: you can use eve::add directly, so we don't need to use `<functional>`

jfalcou (commented):
LGTM apart from the small comments

test/unit/algo/array_utils.cpp
DenisYaroshevskiy: What's wrong with using functional?

jfalcou: well, compile time and all that ;)
It's not a big deal but I think it could be good to check it also works with our own callables.

DenisYaroshevskiy: Can't do because then the `eve::add` and `eve::min` are not constexpr. And these functions seems like they should be constexpr friendly.

I'm also not sure about compile time at all.

include/eve/algo/array_utils.hpp
DenisYaroshevskiy: I can do that but it requires including `<tuple>` since this is where `apply` is defined.
Do we want that?

PR #720, opened 2021-05-20
Implements bundle<simd_value...>

DenisYaroshevskiy:
a bundle is a tuple-like aggregation of wid eof various types
usable in variosu scenario (zip_iterator, complex, pixel).

It relies on kumi::tuple and a bunch of adjustment on the existing
tuple handling.


include/eve/bundle.hpp
```diff
@@ -0,0 +1,77 @@
...
+
+  template<simd_value Member, simd_value... Members>
+  requires( (Member::size() == Members::size()) && ... && true )
+  struct bundle<Member, Members...>
+  {
+    using is_product_type = void;
+
+    bundle() = default;
+
+    bundle(Member const& m,Members const&... ms)    : contents{m,ms...} {}
```
DenisYaroshevskiy, line 27: explicit and probably by value?

include/eve/bundle.hpp
```diff
@@ -0,0 +1,77 @@
...
+    bundle(kumi::tuple<Member,Members...> const& t) : contents{t} {}
```
DenisYaroshevskiy, line 28: Probably explicit

include/eve/bundle.hpp
```diff
@@ -0,0 +1,77 @@
...
+#include <eve/traits/cardinal.hpp>
+#include <eve/traits/element_type.hpp>
+#include <eve/concept/vectorized.hpp>
+
+namespace eve
+{
+  template<simd_value... Members> struct bundle;
+
+  template<simd_value Member, simd_value... Members>
+  requires( (Member::size() == Members::size()) && ... && true )
```
DenisYaroshevskiy, line 20: Wait.

What `size` this is. Is it `wide::size()` or `bundle::size()` because `bundle::size()` is `tuple_size`.

Probably `cardinal`?

include/eve/bundle.hpp
```diff
@@ -0,0 +1,77 @@
...
+  };
+
+  template<simd_value... Members> bundle(kumi::tuple<Members...> const&)  -> bundle<Members...>;
+  template<simd_value... Members> bundle(Members const&...)               -> bundle<Members...>;
+
+  template<typename Function, kumi::product_type... Bundles>
+  decltype(auto) map(Function&& f, Bundles const&... bs)
+  {
+    return bundle{kumi::map(f, bs...)};
+  }
```
DenisYaroshevskiy, line 53: Let's make these friends?

Can we still do `kumi` operations directly to get a `kumi::tuple`? We'll need it.

include/eve/bundle.hpp
```diff
@@ -0,0 +1,77 @@
...
+
+  template<typename... Members>
+  decltype(auto) flatten(bundle<Members...> const& b)
+  {
+    return bundle{kumi::flatten(b)};
+  }
```
DenisYaroshevskiy, line 59: Does this work with `bundle<bundle<wide<int>, wide<float>>>`? 
It should right?
You'll get `bundle<wide<int>, wide<float>>`

include/eve/bundle.hpp
jfalcou: you can yes. 
friends is obviosu yes

include/eve/bundle.hpp
jfalcou: I didn't propagate flatten_all yet

include/eve/bundle.hpp
jfalcou: Done

include/eve/bundle.hpp
jfalcou: cardinal_v is the correct thing

include/eve/arch/cpu/wide.hpp
```diff
@@ -79,7 +79,24 @@ inline namespace EVE_ABI_NAMESPACE
...
+      {
+        // use the largest type as a reference for the alignment computation
+        constexpr auto sizes  = kumi::fold_left ( [](auto acc, auto m)
+                                                  {
+                                                    return acc < sizeof(m) ? sizeof(m) : acc;
+                                                  }
+                                                , Type{}, 0ULL
+                                                );
+        return sizes*Cardinal{};
+      }
```
DenisYaroshevskiy, line 94: This does not make sense much to me.
Why are we requiring the maximum alignment for all types? And for what?

include/eve/concept/vectorizable.hpp
```diff
@@ -7,22 +7,49 @@
...
+  template<typename T>
+  static constexpr bool check_tuple()
+  {
+    return []<std::size_t... I>( std::index_sequence<I...> )
+    {
+      return  ( is_scalar_value<std::remove_cvref_t<decltype( get<I>(std::declval<T>()) )>>::value
+                && ...
+                && true
+              );
+    }(std::make_index_sequence<kumi::size<T>::value>{});
```
DenisYaroshevskiy, line 42: Wait... Does this work for `tuple<tuple>` ?
I mean I see specialisation below, but does it kick in here?

include/eve/detail/function/simd/arm/neon/lookup.hpp
```diff
@@ -7,68 +7,88 @@
...
+          return bit_cast(bytes_t{lookup(b, i1)},as(a));
+        }
+        else
+        {
+          auto bb = bit_cast(idx, as_<typename wide<I,N>::template rebind<std::uint8_t,fixed<16>>>());
+          constexpr auto fix = fix_pattern<N::value>(as_pattern { [](auto i, auto)
+                                                                  { return sizeof(I) * i; }
+                                                                }
+                                                    );
+          return lookup(a, bb[fix]);
```
DenisYaroshevskiy, line 53: Wait... I thought we decided not to use shuffle internally?
What is this even doing?

include/eve/detail/function/simd/common/broadcast.hpp
```diff
@@ -14,13 +14,31 @@ namespace eve::detail
   template<simd_value Wide>
   EVE_FORCEINLINE auto broadcast_(EVE_SUPPORTS(cpu_), Wide v, auto Index) noexcept
   {
-    return Wide{ v.get(Index) };
+    if constexpr( is_bundle_v<typename Wide::abi_type> )
+    {
+      return Wide(kumi::map( [&]<typename T>(T m) { return broadcast(m,Index); }, v.storage()));
```
DenisYaroshevskiy, line 19: Hmm, can we like make a standard helper to do this?

include/eve/detail/function/simd/common/combine.hpp
```diff
@@ -34,6 +34,13 @@ namespace eve::detail
 
       return that;
     }
+    else if constexpr( is_bundle_v<abi_t<T, N>> )
+    {
+      return that_t ( kumi::map ( [](auto ml, auto mh) { return eve::combine(ml,mh); }
```
DenisYaroshevskiy, line 39: I think you can just pass `eve::combine` here.

include/eve/detail/function/simd/common/slice.hpp
```diff
@@ -49,31 +49,16 @@ namespace eve::detail
...
+    if constexpr( is_aggregated_v<abi_t> )  return a.storage().segments;
+    else                                    return std::array{a.slice(lower_), a.slice(upper_)};
   }
 
   //================================================================================================
   // Partial slice
   //================================================================================================
   template<typename P, typename Slice>
-  EVE_FORCEINLINE auto slice_impl(P const &a, Slice const &) noexcept
+  EVE_FORCEINLINE auto slice_impl(P const &a, [[maybe_unused]] Slice const & s) noexcept
```
DenisYaroshevskiy, line 61: Is this maybe unused necessary? It seems weird, we have this used under an if constexpr

include/eve/detail/function/simd/lookup_helpers.hpp
```diff
@@ -67,5 +68,14 @@ namespace eve::detail
...
   }();
+
+/*
+  template<typename I, bool isLittleEndian = true>
+  inline constexpr auto offset2 = [](auto i, auto)
+  {
+    if constexpr( isLittleEndian )  return sizeof(I) - (i%sizeof(I));
+    else                            return i % sizeof(I);
+  };
+*/
```
DenisYaroshevskiy, line 79: Remove dead code?

include/eve/detail/is_wide.hpp
```diff
@@ -1,23 +1,24 @@
...
+//   {};
 
-  template<typename Type, typename N>
-  struct is_wide<eve::wide<Type,N>> : std::true_type
-  {};
-}
+//   template<typename Type, typename N>
+//   struct is_wide<eve::wide<Type,N>> : std::true_type
+//   {};
+// }
```
DenisYaroshevskiy, line 23: Delete?

include/eve/module/real/core/function/regular/generic/load.hpp
```diff
@@ -43,7 +43,67 @@ namespace eve::detail
...
+  }
+
+/*  template<typename... Pointers,typename Cardinal>
+  EVE_FORCEINLINE auto load_( EVE_SUPPORTS(cpu_), kumi::tuple<Pointers...> ptr
+                            , Cardinal
+                            ) noexcept
+  {
+    return bundle { kumi::map ( [](auto p) { return eve::load(p, Cardinal{}); }, ptr) };
+  }
+*/
```
DenisYaroshevskiy, line 63: Delete?

test/unit/memory/store/tuple/aligned.cpp
```diff
@@ -0,0 +1,66 @@
...
+  {
+    ref0[i] = static_cast<std::int8_t>('a'+i);
+    ref1[i] = static_cast<T>(i + 1);
+    ref2[i] = 1.5*(1+i);
+  }
+
+  constexpr auto algt = eve::alignment_v<w_t>;
+  alignas(algt) std::array<std::int8_t, w_t::size()> target0;
+  alignas(algt) std::array<T          , w_t::size()> target1;
+  alignas(algt) std::array<double     , w_t::size()> target2;
```
DenisYaroshevskiy, line 52: Yeah, I have a problem with 1 alignment value in bytes

DenisYaroshevskiy (commented):
Most of it looks OK. I need to dig dip through lookup on arm, I don't like the usage of patter there.

Other than that - we should talk alignment.

Can we maybe have something simple and geenric that just "for this operation do a kumi::map"?

include/eve/detail/function/simd/common/combine.hpp
jfalcou: Right, probably wrote that before adapting combine

include/eve/detail/function/simd/common/slice.hpp
jfalcou: I think that when we don't trigger the if constexpr it is unused. I'll retest

include/eve/detail/function/simd/lookup_helpers.hpp
jfalcou: Will do at the end yes

include/eve/detail/is_wide.hpp
jfalcou: Oh yeah forgot tot do it !

include/eve/module/real/core/function/regular/generic/load.hpp
jfalcou: it's a placeholder for me, code will appear here soon

test/unit/memory/store/tuple/aligned.cpp
jfalcou: Will as fix as per discussion

include/eve/arch/cpu/wide.hpp
jfalcou: We decided that alignment should die. Reverting this

include/eve/detail/function/simd/common/broadcast.hpp
jfalcou: Will propagate required kumi functions as friend of wide

include/eve/concept/vectorizable.hpp
jfalcou: do you think we need to call flatten_all before ?

test/unit/api/tuple/swizzle/broadcast.cpp
```diff
@@ -0,0 +1,44 @@
...
+#include <cstdint>
+
+template<int I, int N>
+inline constexpr auto broadcast = eve::fix_pattern<N>( [](int, int){ return I; } );
+
+//==================================================================================================
+// Identity test
+//==================================================================================================
+EVE_TEST_TYPES( "Check behavior of broadcast swizzle", eve::test::scalar::all_types)
+<typename T>(eve::as_<T>)
```
DenisYaroshevskiy, line 20: Why not just run these for all `simd` types and take the cardinal from there.
That way you don't need to do wto level of constexpr for_

include/eve/detail/function/simd/common/slice.hpp
```diff
@@ -49,31 +49,16 @@ namespace eve::detail
...
-
-    if constexpr( is_emulated_v<abi_t> )
-    {
-      auto eval = [&](auto... I) {
-        using that_t = std::array<sub_t, 2>;
-        return that_t {sub_t {a.get(I)...}, sub_t {a.get(I + card_t::value / 2)...}};
-      };
 
-      return apply<card_t::value / 2>(eval);
-    }
```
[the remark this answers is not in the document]
jfalcou: Calling both slice gave the same codegen than the ùanual complicated code so I decided to merge them

Issue #716, opened 2021-05-18
[FEATURE] aligning in the algorithms


DenisYaroshevskiy:
The way I see it we have the following possible `alignment` for iterator.

`unaligned_iterator` - no alignment what so ever. Can represent any position in the range. Example `unaligned_ptr_iterator`.
`aligned_iterator` - we cannot add an arbitrary number but loading/storing is more efficient. + if we ever get to do unguarded algorithms, like strlen, this type will become much more important. Example `aligned_ptr_iterator`
`partially_aligned` - we cannot add an arbitrary number and we also cannot do unguarded algorithms. Presumably the load and store are more efficient than making the same iterator unaligned. `zip<aligned, unaligned>` for example.
`always_aligned` - this means we can add arbitrary number, we still efficient and we can do unguarded algorithms. Example `iota_iterator`.

Any iterator should be able to be converted to `unaligned`, otherwise we can't do find. We already have that.

In this task I want to also implement `previous_partially_aligned` method. I think it can be required for all iterator types, at least at this point.

I also want for us to recognise `partially_aligned_iterator` concept: type returned by `previous_partially_aligned` matches the type of the iterator.

`prepare_range` needs to recognise `partially_aligned` and not and depending on that populate a trait `partially_align`. And `for_each_iteration`, if this trait is present, should use that to do aligned iteration instead of unaligned.

PR #723, opened 2021-05-21
FIX-#716: aligning in algorithms MVP.

DenisYaroshevskiy:
* Iterators now have `previous_partially_aligned` method. I am not sure if every iterator should have it (efficiency of it is a different question).

* Introduced some of the basic concepts. Unfortunately a bit struggle with `writeable_iterator` so `iterator_basics` and `readable_iterator` for now. Also they are not fully implemented.

* Introduced `no_aligning` the tell us not to use any of the `prervious_partially_aligned` business.

* Hacked together defaulting logic for traits since I need it.

* `preprocess_range` now can deduce some of the traits: like that `no_aligning` and `divisible_by_cardinal`. I'm not actually using `divisible_by_cardinal` yet.

* `preprocess_range` now can accept `aligned_ptr` as a first parameter.

* `for_each_iteration` now checks that the range it calls the last iteration with is not empty. The reason why it's important is that we require to `load` from an actually valid address. But `l` can be the end of the page in which case calling with `load(l)` is not cool.

* `for_each_iteration` now checks traits to figure wether it should try to align or not.

### DIFF
--- a/include/eve/algo/concepts.hpp
+++ b/include/eve/algo/concepts.hpp
@@ -16,13 +16,13 @@ namespace eve::algo
   namespace detail
   {
     template <typename>
-    struct is_eve_fixed  : std::false_type {};
+    struct is_fixed  : std::false_type {};
 
     template <std::ptrdiff_t N>
-    struct is_eve_fixed<eve::fixed<N>> : std::true_type {};
+    struct is_fixed<eve::fixed<N>> : std::true_type {};
 
     template <typename T>
-    concept is_eve_fixed_v = is_eve_fixed<T>::value;
+    concept is_fixed_v = is_fixed<T>::value;
   }
 
   template <typename T>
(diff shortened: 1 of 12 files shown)


include/eve/algo/traits.hpp
```diff
@@ -30,11 +30,54 @@ namespace eve::algo
...
+
+    auto no_aligning_ = [&] (auto ...other){
+      if constexpr (o_t::contains(no_aligning) ||
+                    d_t::contains(no_aligning))
+      {
+        return divisible_(no_aligning, other...);
+      }
+      else
+      {
+        return unroll_(other...);
```
DenisYaroshevskiy, line 77: Oopss

include/eve/algo/for_each_iteration.hpp
```diff
@@ -19,19 +20,48 @@ namespace eve::algo
...
+      if (precise_l == l) return;
+
       eve::keep_first ignore{l - precise_l};
       delegate.step(f, ignore);
     }
 
+    template <typename Traits, unaligned_iterator I, sentinel_for<I> S, typename Delegate>
+    EVE_FORCEINLINE void operator()(Traits traits, I f, S l, Delegate& delegate) const
+      requires (!Traits::contains(no_aligning)())
+    {
```
DenisYaroshevskiy, line 42: Both should assert a non-empty range otherwise load might have issues

include/eve/algo/concepts.hpp
```diff
@@ -8,9 +8,23 @@
...
 #include <concepts>
+#include <type_traits>
+#include <eve/function/load.hpp>
 
 namespace eve::algo
 {
+  namespace detail
+  {
+    template <typename>
+    struct is_eve_fixed  : std::false_type {};
```
jfalcou, line 19: Not a fan of using eve_ in name of traits.

include/eve/algo/concepts.hpp
```diff
@@ -21,6 +35,59 @@ namespace eve::algo
...
+  using partially_aligned_t = decltype(std::declval<T>().previous_partially_aligned());
+
+  template <typename T>
+  struct partially_aligned
+  {
+    using type = partially_aligned_t<T>;
+  };
+
+  template <typename I>
+  concept unaligned_check = std::same_as<unaligned_t<I>, I>;
```
jfalcou, line 48: not fan of _check either. If needed, eve::concept namespace could be a thing

include/eve/algo/concepts.hpp
```diff
@@ -21,6 +35,59 @@ namespace eve::algo
...
+
+  template <typename I>
+  concept partially_aligned_check = std::same_as<partially_aligned_t<I>, I>;
+
+  template <typename I>
+  concept iterator_basics =
```
jfalcou, line 54: why not just iterator ?

include/eve/algo/concepts.hpp
DenisYaroshevskiy: Did `is_fixed`. It's in detail`

PR #731, opened 2021-05-23
Move to g++ and 11 clang++ 12 in Github Actions

jfalcou:


DenisYaroshevskiy:
If you change docker, can you please update how to build doc

PR #732, opened 2021-05-23
FIX-670-A: automated run

DenisYaroshevskiy:
Seeing what I managed to generate with clang tooling (I only ran on a small part, some things do break).

I dunno if should be merged - I still didn't do all platforms.


DenisYaroshevskiy:
Split into bits

PR #734, opened 2021-05-25
Improve == and != on integral AVX/AVX2

jfalcou:


include/eve/detail/function/simd/x86/friends.hpp
```diff
@@ -163,7 +173,20 @@ namespace eve::detail
...
+      constexpr auto c = categorize<wide<T, N, ABI>>();
+
+      if constexpr( ( c == category::int64x4 || c == category::uint64x4 ||
+                      c == category::int32x8 || c == category::uint32x8 )
+                  && current_api == avx
+                  )
+      {
+        // AVX has a bad == for those integers so we use XOR
+        return bit_cast(~v.bits() ^ w.bits(), as(v));
+      }
```
DenisYaroshevskiy, line 185: I suspect that `~v.bits()` is `andnot_zero`.
I suspect you can do:

`_mm256_cmp_ps(v, w, _CMP_UNORD_Q)` (well pd for 64 bit)

Since true is Nan. Judging by the description of `_mm_cmpunord_pd` this is exactly what we need.

include/eve/detail/function/simd/x86/friends.hpp
```diff
@@ -149,6 +150,15 @@ namespace eve::detail
...
         else  if constexpr( c == category::uint16x8 )             return _mm_cmpeq_epi16(v,w);
         else  if constexpr( c == category::uint8x16 )             return _mm_cmpeq_epi8 (v,w);
+        else  if constexpr( c == category::int64x4 || c == category::uint64x4 ||
+                            c == category::int32x8 || c == category::uint32x8
+                          )
+        {
+          using f_t = as_floating_point_t<wide<T, N, ABI>>;
+          auto fv   = bit_cast( v , as_<f_t>{});
+               fv  ^= bit_cast( w, as_<f_t>{});
+          return bit_cast ( !to_logical( fv ), as_<logical<wide<T, N, ABI>>>{} );
```
DenisYaroshevskiy, line 160: This is 4 instructions, right?

xor + not + zero + andnot ?

(well zero maybe hoisted out).

DenisYaroshevskiy (approved):

PR #744, opened 2021-05-29
FIX-670: manual fixes for specifically mentioned ABI

DenisYaroshevskiy:
These fixes are not as dumb, but I refactor a bit and also add support for avx512 where it's easy.

in max/ min for avx went to if_else for 4 byte integers since there is a blend.

### DIFF
--- a/include/eve/detail/function/simd/x86/combine.hpp
+++ b/include/eve/detail/function/simd/x86/combine.hpp
@@ -92,7 +92,7 @@ namespace eve::detail
   combine ( avx512_ const & , logical<wide<T, N>> const &l
                             , logical<wide<T, N>> const &h
           ) noexcept
-      requires x86_abi<abi_t<T, N>>
+    requires x86_abi<abi_t<T, N>>
   {
     using s_t = typename logical<wide<T, typename N::combined_type>>::storage_type;
     using i_t  = typename s_t::type;
--- a/include/eve/detail/function/simd/x86/combine.hpp
+++ b/include/eve/detail/function/simd/x86/combine.hpp
@@ -92,7 +92,7 @@ namespace eve::detail
   combine ( avx512_ const & , logical<wide<T, N>> const &l
                             , logical<wide<T, N>> const &h
           ) noexcept
-    requires x86_abi<abi_t<T, N>>
+      requires x86_abi<abi_t<T, typename N::combined_type>>
   {
     using s_t = typename logical<wide<T, typename N::combined_type>>::storage_type;
     using i_t  = typename s_t::type;
--- a/include/eve/detail/function/simd/x86/combine.hpp
+++ b/include/eve/detail/function/simd/x86/combine.hpp
@@ -7,8 +7,10 @@
 //==================================================================================================
 #pragma once
 
+#include <eve/concept/abi.hpp>
 #include <eve/detail/function/simd/x86/make.hpp>
 #include <eve/detail/implementation.hpp>
+
 
 namespace eve::detail
 {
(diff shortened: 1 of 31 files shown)


include/eve/function/add.hpp
```diff
@@ -91,3 +91,7 @@ namespace eve
 }
 
 #include <eve/module/real/core/function/regular/generic/add.hpp>
+
+#if defined(EVE_INCLUDE_X86_HEADER)
+#  include <eve/module/real/core/function/regular/simd/x86/floor.hpp>
```
jfalcou, line 96: why adding floor to add ?

include/eve/module/real/core/function/regular/simd/x86/average.hpp
```diff
@@ -16,32 +16,24 @@
...
-  EVE_FORCEINLINE wide<T, N, x86_256_> average_(EVE_SUPPORTS(avx_)
-                                           , wide<T, N, x86_256_> const &a
-                                           , wide<T, N, x86_256_> const &b) noexcept
-  {
-    if constexpr(current_api >= avx2)
+    if constexpr ( sizeof(T) == 2 )
     {
-             if constexpr(sizeof(T) == 1) return _mm256_avg_epu8(a, b);
-        else if constexpr(sizeof(T) == 2) return _mm256_avg_epu16(a, b);
-        else                              return (a & b) + ((a ^ b) >> 1);
```
jfalcou, line 42: where does this one went ?

include/eve/module/real/core/function/regular/simd/x86/ceil.hpp
```diff
@@ -20,21 +20,20 @@ namespace eve::detail
   //-----------------------------------------------------------------------------------------------
   // 128 bits implementation
   template<floating_real_scalar_value T, typename N, typename ABI>
-  EVE_FORCEINLINE wide<T, N, x86_128_> ceil_(EVE_SUPPORTS(sse4_1_)
-                                        , wide<T, N, x86_128_> const &a0) noexcept
```
jfalcou, line 24: will we change those to if constexpr later ?

include/eve/module/real/core/function/saturated/simd/x86/add.hpp
```diff
@@ -14,103 +14,37 @@
...
+    else if constexpr ( std::same_as<abi_t<T, N>, x86_256_> && current_api == avx ) return add_(EVE_RETARGET(cpu_), st, v0, v1);
+    else if constexpr ( std::signed_integral<T> && sizeof(T) == 1 )
     {
-      return add(v0, v1);
+           if constexpr ( std::same_as<abi_t<T, N>, x86_512_> ) return _mm512_adds_epi8(v0, v1);
+      else if constexpr ( std::same_as<abi_t<T, N>, x86_256_> ) return _mm256_adds_epi8(v0, v1);
+      else if constexpr ( std::same_as<abi_t<T, N>, x86_128_> ) return _mm_adds_epi8   (v0, v1);
     }
-    else
+    else if constexpr ( std::signed_integral<T> && sizeof(T) == 2 )
```
jfalcou, line 31: not sure we'll have float of size 2 on x86 so th eis_integral is maybe useless

jfalcou (changes requested):

include/eve/detail/function/simd/x86/combine.hpp
```diff
@@ -7,155 +7,92 @@
...
-      else  if constexpr( std::integral<T> )
-      {
-        return _mm_castps_si128(
-            _mm_shuffle_ps(_mm_castsi128_ps(l.storage()), _mm_castsi128_ps(h.storage()), 0x44));
-      }
+      else                                          return _mm_castps_si128(_mm_shuffle_ps(_mm_castsi128_ps(l),
+                                                                            _mm_castsi128_ps(h),
+                                                                            0x44));
     }
-    //==============================================================================================
```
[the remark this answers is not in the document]
DenisYaroshevskiy: I don't see why in the world they are useful

include/eve/module/real/core/function/regular/simd/x86/ceil.hpp
DenisYaroshevskiy: I just didn't find the funcitons. Will fix.

include/eve/module/real/core/function/saturated/simd/x86/add.hpp
DenisYaroshevskiy: This is about `signed` and `unsigned` not about floats

include/eve/module/real/core/function/regular/simd/x86/ceil.hpp
DenisYaroshevskiy: Done

include/eve/module/real/core/function/regular/simd/x86/frac.hpp
```diff
@@ -35,12 +35,12 @@ namespace eve::detail
   // -----------------------------------------------------------------------------------------------
   // 256 bits implementation for xop
   template<floating_real_scalar_value T, typename N>
-  EVE_FORCEINLINE auto frac_(EVE_SUPPORTS(avx_),
-                             wide<T, N, x86_256_> const &a0) noexcept
+  EVE_FORCEINLINE wide<T, N> frac_(EVE_SUPPORTS(avx_), wide<T, N> const &a0) noexcept
```
jfalcou, line 38: why the change from `auto`?

DenisYaroshevskiy: Because `_mm256_frcz_pd` does not return `wide`

include/eve/module/real/core/function/regular/simd/x86/average.hpp
DenisYaroshevskiy: CPU should  do this https://github.com/jfalcou/eve/blob/1e865bfb459ea3dfb4a6c9b10944c55ae761a841/include/eve/module/real/core/function/regular/generic/average.hpp#L42

jfalcou (approved):

Issue #725, opened 2021-05-21
[FEATURE] for each iteration direct testing


DenisYaroshevskiy:
I figured how to write really nice for each iteration tests. This'd allow to reduce the load on the algorithm testing.

DenisYaroshevskiy:
https://github.com/jfalcou/eve/pull/759

PR #759, opened 2021-06-05
FIX-#725: for each improvements

DenisYaroshevskiy:
Changes:
* Split `for_each` into the `selection` part and the `execution` part. This is needed because many algorithms actually need to know the `base` of iteration before the actual algorithm starts (for example in-place transform).
* Added support for `divisible_by_cardinal`
* Added proper testing to `for_each`. This should help to reduce the need for testing all cases in all algorithms, though I'm not entirely sure how that'd look.

### DIFF
--- a/include/eve/conditional.hpp
+++ b/include/eve/conditional.hpp
@@ -18,6 +18,7 @@
 #include <eve/traits/as_arithmetic.hpp>
 #include <eve/traits/cardinal.hpp>
 #include <bitset>
+#include <compare>
 #include <ostream>
 
 namespace eve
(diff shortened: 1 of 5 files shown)


include/eve/conditional.hpp
```diff
@@ -99,10 +100,13 @@ namespace eve
       return 0ULL;
     }
 
+    constexpr auto friend operator<=>(ignore_all_ const&, ignore_all_ const&) = default;
```
jfalcou, line 103: doesnt this give conditional < and > ? I fear they are nto what we think they should be ?

Issue #726, opened 2021-05-21
[FEATURE] one range algorithm common interface


DenisYaroshevskiy:
We should expect that all of the one range algorithms will have similar interfaces.
We should have just one component to do it.

DenisYaroshevskiy:
Went a different way

PR #761, opened 2021-06-05
FIX-#726: one range algorithm common interface + find_if

DenisYaroshevskiy:
Introducing a common interface adapter for all our algorithms over one range.
The goal is to get to a canonicalised `traits, iterator, sentinel` interface from all of the possible variations on the subject.

Added `find_if` since it utilises more features of the API.
So far - not sure how to go about `first_true(array<wide>)` - created a story to think about it: https://github.com/jfalcou/eve/issues/764
We'll see what makes sense.

Had to add a few apis to aligned_ptr.

UPD: ranges support is busted in clang, even libstdcxx is not working.
So for now - no built in arrays.

### DIFF
--- a/include/eve/algo/array_utils.hpp
+++ b/include/eve/algo/array_utils.hpp
@@ -46,10 +46,14 @@ namespace eve::algo
   }
 
   template <typename T, std::size_t N, typename Op>
-  EVE_FORCEINLINE constexpr void array_reverse_it(std::array<T, N> x, Op op)
-  {
-    [&]<std::size_t ...i>(std::index_sequence<i...>){
-      (op(x[N - i - 1]), ...);
+  EVE_FORCEINLINE constexpr std::size_t find_branchless(std::array<T, N> x, Op op) {
+    return [&]<std::size_t ...i> ( std::index_sequence<i...> ) {
+      std::size_t res = N;
+      auto update_res_if = [&](std::size_t j) mutable {
+        if (op(x[j])) res = j;  // should generate cmov
+      };
+      (update_res_if(N - i - 1), ...);
+      return res;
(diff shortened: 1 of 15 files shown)


include/eve/algo/find.hpp
```diff
@@ -0,0 +1,94 @@
...
+        // TODO: this is not the best solution, see: #764
+        auto overall = array_reduce(tests, eve::logical_or);
+        if (!eve::any(overall)) return false;
+
+        std::size_t i = size - 1;
+        std::optional<std::ptrdiff_t> match;
+        std::size_t pos = i;
+
+        array_reverse_it(tests, [&](auto test) mutable {
+          auto local_match = eve::first_true(test);
```
DenisYaroshevskiy, line 59: I should use `any` here. After what - check the asm.

include/eve/algo/array_utils.hpp
```diff
@@ -44,4 +44,13 @@ namespace eve::algo
   {
     return detail::array_reduce_impl<0, N>(x, op);
   }
+
+  template <typename T, std::size_t N, typename Op>
+  EVE_FORCEINLINE constexpr void array_reverse_it(std::array<T, N> x, Op op)
```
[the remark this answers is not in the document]
DenisYaroshevskiy: I called it `find_branchless`.

I'd need to iterate on it on godbolt a bit. I'd prefer to do it when I have all the bits and pieces.

include/eve/algo/find.hpp
DenisYaroshevskiy: Not sure if it's a good idea. Will have a look on godbolt once everything is merged.

jfalcou (approved):

PR #769, opened 2021-06-10
FIX-768: unsafe(compress_store) basic tests

DenisYaroshevskiy:
Started sketching compress store.
I feel like we can argue the specifics of the interface a bit later, like how are the modifiers suppose to be called and so on.

So far - very very simple stuff.
Unfortunately, the proper implementations even for all arches I can run natively, has to wait a bit until I'm able to compile all of my benchmarks for any platform (that's not so long: I need like `inclusive_scan` and `compress_store`).

However, I can implement at least one version that I wanted to try out.


jfalcou (approved):
Seems fine. I have one side question we can resolve later :
 - compressed_store return the pointer where it stored last. It is mandatory for usign c_store properly
 - store returns void
 
 Shouldn't store also return the address of where it last stored even if it's trivial ?

Issue #768, opened 2021-06-10
[FEATURE] compress_store, basic x86


DenisYaroshevskiy:
This is a tough one obv.

I'm going to start with the one proposed here:
https://chat.stackoverflow.com/rooms/212510/discussion-between-denis-yaroshevskiy-and-peter-cordes
by @aqrit

Because I know that the BMI one that I have is extremely bad on amd.

DenisYaroshevskiy:
https://github.com/jfalcou/eve/pull/842

PR #771, opened 2021-06-11
FIX-#768: compress store, T > 4 bytes, ssse3.

DenisYaroshevskiy:
* as_int() for non-aggregated top_bits. We just need it
* couple of uniform ptr operations in detail to make our life easier.
* removed slide left masking. It's not doing anything.
* utilised some of the main tricks from SIMD.JSON compress store: https://gist.github.com/aqrit/6e73ca6ff52f72a2b121d584745f89f3#file-despace-cpp-L141

Tricks used:
* Last element can be always written
* We can `shift_left` by 1 and blend to get rid of 1 element before that (only did it for 2 elements) in order to not do masks.
* For 4 elements we only need to store 9 masks. (we can get away with 2 by using previous trick, but I figured 9 masks is not too bad).
* We can shuffle with extra bits on top 
* I store the `popcount` in the mask.

Storing popcount is not a given that a win, but before sse4.2 the popcount is a mutli instruction disaster: https://godbolt.org/z/1vqjd5qx9

### DIFF
--- a/include/eve/memory/aligned_ptr.hpp
+++ b/include/eve/memory/aligned_ptr.hpp
@@ -520,29 +520,4 @@ namespace eve
   struct  pointer_alignment<aligned_ptr<Type, Alignment>>
         : std::integral_constant<std::size_t,Alignment>
   {};
-
-namespace detail
-{
-  template <typename T>
-  EVE_FORCEINLINE auto as_raw_pointer(T p)
-  {
-    if constexpr ( !std::is_pointer_v<T> ) return p.get();
-    else                                   return p;
-  }
-
-  template <typename U, typename T, std::size_t size>
-  EVE_FORCEINLINE auto ptr_cast(eve::aligned_ptr<T, size> p)
-    requires (sizeof(U) == sizeof(T))
-  {
-    return aligned_ptr<U, size>{(U*)(p.get())};
-  }
-
-  template <typename U, typename T>
-  EVE_FORCEINLINE auto ptr_cast(T* p)
-    requires (sizeof(U) == sizeof(T))
-  {
-    return (U*)p;
-  }
-}
-
 }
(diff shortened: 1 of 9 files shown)


include/eve/memory/aligned_ptr.hpp
```diff
@@ -520,4 +520,29 @@ namespace eve
   struct  pointer_alignment<aligned_ptr<Type, Alignment>>
         : std::integral_constant<std::size_t,Alignment>
   {};
+
+namespace detail
```
jfalcou, line 524: coudl be good if it was in a memory/pointer.hpp file

jfalcou (approved):

aqrit:
The only thing burrowed from simdjson was the fast whitespace detection.

simdjson implementation burrows a bit from me... originally they were using a 1 MiB table.
Now they are done to 2 KiB table with another 272 bytes for "combine" table.
https://github.com/simdjson/simdjson/blob/95b4870e20be5f97d9dcf63b23b1c6f520c366c1/include/simdjson/westmere/simd.h#L119

The combine table is a bit of a question mark... there are pshufb tricks to avoid it. 


InstLatx64 supposedly developed the fastest version, but I don't know if they ever made it available.
https://github.com/InstLatx64/AVX512_VPCOMPRESSB_Emu

DenisYaroshevskiy:
@aqrit 

We do not support the avx-512 bmi, so that's not for us.

Do you maybe want to implement compress_store for eve? Or review it?
I'm happy to help you set up.

I just used bmi version before but it was horrible on amd.

DenisYaroshevskiy:
This `thin_table` does not seem terribly `thin` to me
https://github.com/simdjson/simdjson/blob/95b4870e20be5f97d9dcf63b23b1c6f520c366c1/src/internal/simdprune_tables.cpp#L39

If we were only focused on really big chunks, maybe, but we are not.

PR #774, opened 2021-06-13
Newtests/combinatorial

jtlap:


include/eve/module/real/combinatorial/function/regular/generic/gcd.hpp
```diff
@@ -158,7 +158,7 @@ namespace eve::detail
           a   = b;
           b   = r;
         }
-        return a;
+        return abs(a);
```
jfalcou, line 161: I thought we should call `eve::abs`

PR #777, opened 2021-06-13
FIX-#768: compress store 4 elements, x86.

DenisYaroshevskiy:
Tagging @aqrit as well.
FYI: we store less than register size in lower bits of _m128i.

Compress store for 4 elements, x86.

For avx2 I decided to stick with calling `popcount` over getting it from shuffle pattern.
I'd have to measure this later.

Also fixed `convert` for logicals just in case.

### DIFF
--- a/include/eve/memory/pointer.hpp
+++ b/include/eve/memory/pointer.hpp
@@ -22,14 +22,12 @@ namespace eve::detail
 
   template <typename U, typename T, std::size_t size>
   EVE_FORCEINLINE auto ptr_cast(eve::aligned_ptr<T, size> p)
-    requires (sizeof(U) == sizeof(T))
   {
     return aligned_ptr<U, size>{(U*)(p.get())};
   }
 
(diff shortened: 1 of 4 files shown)


include/eve/memory/pointer.hpp
```diff
@@ -22,14 +22,12 @@ namespace eve::detail
 
   template <typename U, typename T, std::size_t size>
   EVE_FORCEINLINE auto ptr_cast(eve::aligned_ptr<T, size> p)
-    requires (sizeof(U) == sizeof(T))
```
jfalcou, line 25: shouldn't it just be std::convertibel_to<U*,T*> ?

DenisYaroshevskiy: Nah. It's a C style cast for aligned ptr

jfalcou: OK

PR #778, opened 2021-06-14
eve::memory cleanup

jfalcou:
- Remove old code that get superceded by C++20 `<bit>`
- Make all aligne_dptr related tools to use lanes instead of raw integrals


DenisYaroshevskiy (commented):
Looks good so far

include/eve/module/real/core/function/regular/simd/x86/compress_store.hpp
```diff
@@ -137,7 +137,9 @@ namespace eve::detail
       }();
 
       top_bits mmask{mask};
-      aligned_ptr<std::uint64_t, 32> pattern_ptr{patterns[mmask.as_int() & 7].data()};
+      aligned_ptr < std::uint64_t
+                  , expected_cardinal_t<std::uint64_t>
+                  > pattern_ptr{patterns[mmask.as_int() & 7].data()};
```
DenisYaroshevskiy, line 142: No, not expected cardinal.

THis should be `eve::fixed<4>` even on avx512.

We load specifically 4 integers

include/eve/memory/pointer.hpp
```diff
@@ -20,10 +20,10 @@ namespace eve::detail
     else                                   return p;
   }
 
-  template <typename U, typename T, std::size_t size>
-  EVE_FORCEINLINE auto ptr_cast(eve::aligned_ptr<T, size> p)
+  template <typename U, typename T, typename Lanes>
+  EVE_FORCEINLINE auto ptr_cast(eve::aligned_ptr<T, Lanes> p)
   {
-    return aligned_ptr<U, size>{(U*)(p.get())};
+    return aligned_ptr<U, Lanes>{(U*)(p.get())};
```
DenisYaroshevskiy, line 26: Not `Lanes`. Alignment in bytes should stay the same.

So `eve::fixed<Lanes() * sizeof(T) / sizeof(U)>`

DenisYaroshevskiy (commented):
Looks good so far.

DenisYaroshevskiy (approved):

PR #780, opened 2021-06-14
FIX-#768: compress store 8 integers

DenisYaroshevskiy:
Smaller things
* moved all 4 element ones into one function
* extracted index constants.

Worked again from @aqrit's solution.
https://gist.github.com/aqrit/6e73ca6ff52f72a2b121d584745f89f3#file-despace-cpp-L158

There are a few things that are different:
* I need to convert from `epi32` to `epi8`. I used bits movemask + shifts to do that.
* I get an inverted `idx` and `popcount`. I have no idea how you can get a non-inverted one using `sad` like that.
`idx` is not a problem, but `popcount` means one extra scalar operation. Open to suggestions.

Obv nothing is measured, I will first finish writing some baseline and then measure things.

I tried to comment everything as much as I can, so @jfalcou - please review the comments as well.
The bar for reading this code should be somewhere reachable.

### DIFF
--- a/include/eve/module/real/core/function/regular/simd/x86/compress_store.hpp
+++ b/include/eve/module/real/core/function/regular/simd/x86/compress_store.hpp
@@ -194,21 +194,13 @@ namespace eve::detail
     However, it sums absolute differences for bytes.
     We could use a zero mask.
     But instead we can use the same mask we got to mask the 1, 3, 9.
-    This get's us a somewhat inverted result:
-    0x0101 - 0x0101 becomes a 0 instead of 2.
-    So we can just take that into account when ordering masks.
+    In order to get the result in the correct order, we need to use `andnot` instead of `and`.
+    For 8 integers we can use a popcount + some math.
 
     Example:
     (true, false, true, true, true, false, false, false)
      1             3    3      9
      reduced is 16
-     Total sum is 3^4 -1 = 26.
-     26 - 16 = 10
-     Resulting index is 10
-
-    Since we do reduction, we can also try to squese the popcount in it.
-    But, the popcount will also come inverted, so we need to subtract it from 8.
-    If native is suppored, might be not worth it again.
 
     NOTE: @aqrit does not have the problem with an inverted mask nor subtracting from 8.
           I don't understand how though so far.
@@ -220,21 +212,17 @@ namespace eve::detail
   base_3_mask(eve::logical<eve::wide<T, N>> mask) {
     if constexpr ( sizeof(T) == 4 && N() == 8 && current_api == avx2)
     {
-      // Converting to byte per element
-      // This also messes with order a bit
-      // Alternative is to convert to epi16 using packus but this seems better.
-      std::int32_t to_4_bits_per_element = _mm256_movemask_epi8(mask);
-      std::uint32_t top_half             =  to_4_bits_per_element >> 4;
-      __m128i byte_mask = _mm_set_epi32(top_half, to_4_bits_per_element,
-                                        top_half, to_4_bits_per_element);
-
-      __m128i mask = _mm_set_epi32(
-        0x01010101, 0x01010101,  // popcount bytes
-        0x00090301, 0x00090301   // idx mask
+      std::uint32_t mmask = _mm256_movemask_epi8(
+         eve::bit_cast(mask, eve::as_<logical<eve::wide<std::uint32_t, N>>>{})
       );
 
-      __m128i reduced = _mm_sad_epu8(_mm_and_si128(byte_mask, mask), mask);
-      return {_mm_cvtsi128_si32(reduced), 8 - _mm_extract_epi16(reduced, 4)};
+      // sum from popcount:   [ 1, 1, 3, 3, 4, 1]
+      // sum from extra mask: [ 0, 0, 0, 0, 5, 8]
+      std::uint32_t idx_base_3 = std::popcount(mmask & 0x001f7711);
+      std::uint32_t extra_mask = (mmask >> 17) & 0b1'101;
+
+      idx_base_3 += extra_mask;
+      return {idx_base_3, std::popcount(mmask) / 4};
     }
   }
 
@@ -248,9 +236,8 @@ namespace eve::detail
 
     for (unsigned i = 0; i != 27; ++i)
     {
-      unsigned base_3_value = 26 - i;
-
       unsigned number_of_9s = 0, number_of_3s = 0, number_of_1s = 0;
+      unsigned base_3_value = i;
 
       if (base_3_value >= 9) ++number_of_9s, base_3_value -= 9;
       if (base_3_value >= 9) ++number_of_9s, base_3_value -= 9;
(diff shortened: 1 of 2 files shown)


include/eve/module/real/core/function/regular/simd/x86/compress_store.hpp
```diff
@@ -105,47 +178,132 @@ namespace eve::detail
...
+                                        top_half, to_4_bits_per_element);
+
+      __m128i mask = _mm_set_epi32(
+        0x01010101, 0x01010101,  // popcount bytes
+        0x00090301, 0x00090301   // idx mask
+      );
+
+      __m128i reduced = _mm_sad_epu8(_mm_and_si128(byte_mask, mask), mask);
+      return {_mm_cvtsi128_si32(reduced), 8 - _mm_extract_epi16(reduced, 4)};
+    }
```
DenisYaroshevskiy, line 239: Hmm

`_mm_packus_epi32 ` is actually 1.
But extract is expensive.

`_mm256_packus_epi32` seems also cheap.

And then I can try to use: `_mm256_sad_epu8`

Dunno, maybe it will be better, everything needs a measuring.
Also depends on how well I can do the 8 shorts.

include/eve/module/real/core/function/regular/simd/x86/compress_store.hpp
```diff
@@ -105,47 +178,131 @@ namespace eve::detail
...
+      std::uint32_t top_half             =  to_4_bits_per_element >> 4;
+      __m128i byte_mask = _mm_set_epi32(top_half, to_4_bits_per_element,
+                                        top_half, to_4_bits_per_element);
+
+      __m128i mask = _mm_set_epi32(
+        0x01010101, 0x01010101,  // popcount bytes
+        0x00090301, 0x00090301   // idx mask
+      );
+
+      __m128i reduced = _mm_sad_epu8(_mm_and_si128(byte_mask, mask), mask);
```
DenisYaroshevskiy, line 236: Hmmm. Could I use popcount somehow to add up all the bits?

If I were to convert to bytes? I think it might be an option!

aqrit:
Sorry, I don't grok modern C++.

popcount & idx can be inverted by changing this line: 
`__m128i reduced = _mm_sad_epu8(_mm_and_si128(byte_mask, mask), mask);`
to 
`__m128i reduced = _mm_sad_epu8(_mm_andnot_si128(byte_mask, mask), mask);`
 remember to fixup the table if you do that.

To try the byte ssse3 strategy on dword avx2... One could try to calc the index in a general purpose register.

```
uint64_t x = _mm256_movemask_epi8(mask);
x = (x | (x << 28)) & 0x0101010101010101; // spread out the bits (they don't need to be in the correct order)
count = popcount64(x);
index = (x * magic) >> 59;
```

DenisYaroshevskiy:
> Sorry, I don't grok modern C++.
> 
> popcount & idx can be inverted by changing this line:
> `__m128i reduced = _mm_sad_epu8(_mm_and_si128(byte_mask, mask), mask);`
> to
> `__m128i reduced = _mm_sad_epu8(_mm_andnot_si128(byte_mask, mask), mask);`
> remember to fixup the table if you do that.
> 

andnot! Thanks!

> To try the byte ssse3 strategy on dword avx2... One could try to calc the index in a general purpose register.
> 
> ```
> uint64_t x = _mm256_movemask_epi8(mask);
> x = (x | (x << 28)) & 0x0101010101010101; // spread out the bits (they don't need to be in the correct order)
> count = popcount64(x);
> index = (x * magic) >> 59;
> ```

Yep. I will try this, seems better than going to `sad`

DenisYaroshevskiy:
@aqrit 

I ended up with:

```
      std::uint32_t mmask = _mm256_movemask_epi8(...);
      std::uint32_t idx_base_3 = std::popcount(mmask & 0x001f7711);
      std::uint32_t extra_mask = (mmask >> 17) & 0b1'101;
      idx_base_3 += extra_mask;
```

Do you think it can be done better?

aqrit:
Quite clever!

include/eve/module/real/core/function/regular/simd/x86/compress_store.hpp
```diff
@@ -59,9 +118,27 @@ namespace eve::detail
...
+    requires x86_abi<abi_t<T, N>> && ( N() == 4 )
   {
-    if constexpr ( std::floating_point<T> )
+    if constexpr ( N() == 4 && sizeof(T) == 8 && current_api == avx  )
+    {
+      return compress_store_aggregated_unsafe(v, mask, ptr);
+    }
+    else if constexpr ( N() == 4 && sizeof(T) == 8 )
+    {
+      alignas(32) auto patterns = pattern_4_elements(idx_dwords<std::uint64_t>);
```
jfalcou, line 129: why is htis 32 hardcoded ?

DenisYaroshevskiy: `if constexpr ( N() == 4 && sizeof(T) == 8 )`

DenisYaroshevskiy: Will do in the next patch.

jfalcou (approved):

aqrit:
> Do you think it can be done better?

This should save 4 instructions:

```
v = (uint32_t) _mm256_movemask_epi8(...);
v &= 0x88248421;
count = popcount(v);
index = (v * 0x18C21240) >> 27;

/*
 note: the "trit" "weights" changed from `0 9 3 1` to `0 9 1 3`...
 so the table needs to be updated.
*/
```

I'm wondering if it is possible to have a small (256 byte) permutation table without using a zero-extend (`_mm256_cvtepu8_epi32`).

DenisYaroshevskiy:
@aqrit 

What do you mean by "without"?
I will try to do smaller tables when I have the benchmarks set up.

aqrit:
sorry, just thinking out-loud.

The byte shuffle table could be re-used. Just zero-extend the bytes to dwords on-the-fly.
```
offset = index * 8
shuf = _mm256_cvtepu8_epi32(_mm_loadl_epi64(table + offset));
```

or the shuffle entries could be 32 bytes in length, but are overlapped:
```
/* for example:
 0x000000DD, 0x000000CC, 0x000000BB, 0x000000AA, 0x000000EE ... 
 0x00000044, 0x00000033, 0x00000022, 0x00000011
 0x000000??, 0x000000??, 0x000000??, 0x000000??
 0x000000ZZ, 0x000000YY, 0x000000XX, 0x000000WW

would "stack":
  000000DD000000CC000000BB000000AA000000EE..
  00000000440000003300000022000000110000..
  0000000000??000000??000000??000000??00..
  000000000000ZZ000000YY000000XX000000WW..

and become:
  000000DD44??ZZCC33??YYBB22??XXAA11??WWEE..
*/              
offset = ((index * 8) | index) & 0xE3;
shuf = _mm256_loadu_si256(table + offset);
```

DenisYaroshevskiy:
@aqrit 
Yeah, something like that.
I will try it once I have the benchmarks

Issue #803, opened 2021-07-05
[FEATURE] reduce algo


DenisYaroshevskiy:
Implement the reduce algorithm

DenisYaroshevskiy:
https://github.com/jfalcou/eve/pull/804

PR #804, opened 2021-07-05
FIX-#803: reduce

DenisYaroshevskiy:
This pr is a bit of a mess.
I don't mind splitting it into separate prs if you want me to, but still we can look at it now to understand the overall design.

*NOTE on reduction to a smaller type*
There is one thing that's not terribly correct in this pr - reduction to a smaller type
`std::reduce` always returns the init type.
However, clang vectorises it differently from me: https://godbolt.org/z/vezn1E76P
Clang actually uses operations on chars.
We use operations on ints at the moment.
Clang also will load a full 16 chars on sse2 from a few bigger registers.
This is a touch weird to do in the current framework, for example it's not clear what to do with aligned ptrs.

*Conclusion*: I don't think it's super important at the moment tbh. We can do smth if we want to later.

I need to do more careful thinking on reduce testing, it's lacking but we also need to be careful with our CPU time.

There are 2 main problems with reduce:
* it requires conversions to the sum type (like adding chars to int)
* the `operation` requires a zero.

Therefore I have made the following changes.

* converting_iterator/eve::algo::convert

We need to somehow cast one iterator to produce another iterator type.
After thinking about it I decided against changing the cardinal type here, did not seem very natural, especially because changing of cardinal is not OK up: we cannot just change the cardinal from 8 to 16, this messes with `alignment`.

So `converting_iterator` just calls `eve::convert`.

`convert` produces an iterator with a required type and should be an extension point:
for `ptr` iterators we should change `as_` an utilise platform specific logic, if it's available. We don't have the converting load yet.
for generating iterators, like `iota` it should change the parameters.

Generally speaking the current iterator model is not ideal, clearly, but I do not yet see how it should actually look like.

* Changing algorithm interface to pass traits as an argument/always require a range.

We wanted to do the reduce interface with a custom operation as `reduce(v, {op, zero}, init)`. Unfortunately I can't make that work, so for now it's still `reduce(v, std::pair{op, zero}, init)`. However, I still think the changes I made are good.

There are 2 changes to the algorithm interface:
* now traits are passed in `[]`. So not `eve::algo::any_of(traits, v, pred)` but `eve::algo::any_of[traits](v, pred)`.
Could be cool if we could just pass keywords in brackets `eve::algo::any_of[no_aligning, unroll<2>](v, pred)` but we can't due to `operator[]` not working with multiple parameters.
* To pass a pair of iterators, you need to wrap them in a range `eve::algo::as_range`. `std` opted out to duplicate the interface but it seems like a lot of work.
* `one_range_algorithm_adapter` is no longer needed with these.

New traits:
Since we now change the cardinal in the algorithms, we need a way to force it, so there is a new trait:
`force_cardinal<8>`
We also have a new trait to convert to the common type `common_types_with<std::int64_t, int>`.


include/eve/algo/any_of.hpp
```diff
@@ -9,22 +9,30 @@
...
 #include <eve/function/any.hpp>
 #include <eve/function/logical_or.hpp>
 
 #include <array>
+#include <utility>
 
 namespace eve::algo
 {
-  struct any_of_ : one_range_algorithm_adapter<any_of_>
+  template <instance_of<algo::traits> Traits>
```
jfalcou, line 23: where is instance_of defined ? Is it an old concept we have already ?

include/eve/algo/converting_iterator.hpp
```diff
@@ -0,0 +1,112 @@
...
+
+    T operator*() const { return static_cast<T>(*base); }
+
+    template <sentinel_for<I> I2>
+    friend bool operator==(converting_iterator const& x, converting_iterator<I2, T> const& y)
+    {
+      return x.base == y.base;
+    }
+
+    template <sentinel_for<I> I2>
```
jfalcou, line 55: why both == and <=> ?

include/eve/algo/ptr_iterator.hpp
```diff
@@ -89,6 +92,14 @@ namespace eve::algo
     auto unaligned() const { return unaligned_ptr_iterator<T, Cardinal>{ptr.get()}; }
     auto previous_partially_aligned() const { return *this; }
 
+    template <typename _Cardinal>
+    auto cardinal_cast(_Cardinal) const
```
jfalcou, line 96: is it something we don't want as a free function for any iterator type ?

include/eve/conditional.hpp
```diff
@@ -38,6 +38,13 @@ namespace eve
       return or_<C,T>{static_cast<C const&>(*this), v};
     }
 
+    auto map_alternative(auto op) const
```
jfalcou, line 41: why is this necessary ?

include/eve/algo/any_of.hpp
DenisYaroshevskiy: We do. It's in https://github.com/jfalcou/eve/blob/73c6ea8612bf0e975446b351ac94d8631cd3228c/include/eve/algo/concepts.hpp#L110

Really need it a lot.

include/eve/algo/converting_iterator.hpp
DenisYaroshevskiy: I think either compiler or I don't understand these. Maybe the one from the base class get's picked up.

Basically, doesn't work otherwise.

include/eve/algo/ptr_iterator.hpp
DenisYaroshevskiy: It's a method that we need for every iterator implemented separately.

include/eve/conditional.hpp
DenisYaroshevskiy: Because when we do:

```
load[ignore.else(smth)](converting_iterator) 
```

Or 

```
store[ignore.else(smth)](converting_iterator) 
```

The else part is provided in the type of value we convert to.
And I need to map it to the underlying type.

You will also need this for `load(as<>)` when we have it.

include/eve/algo/any_of.hpp
jfalcou: Great

include/eve/conditional.hpp
jfalcou: OK

include/eve/algo/ptr_iterator.hpp
jfalcou: OK but why as a member function ?

include/eve/algo/converting_iterator.hpp
jfalcou: OK maybe we need to have a deeper look but this is fine for now

include/eve/algo/ptr_iterator.hpp
DenisYaroshevskiy: As oppose to? You mean as a friend?

it's the same as `unaligned` or `previously_partially_aligned` or what have you.

jfalcou: ok

jfalcou (approved):

PR #818, opened 2021-07-21
Fix/avx512

jtlap:


include/eve/module/real/core/function/regular/simd/x86/add.hpp
```diff
@@ -0,0 +1,63 @@
...
+namespace eve::detail
+{
+  // -----------------------------------------------------------------------------------------------
+  // 128 bits implementation
+  template<real_scalar_value T, typename N>
+  EVE_FORCEINLINE wide<T, N> add_(EVE_SUPPORTS(sse2_), wide<T, N> v0, wide<T, N> v1) noexcept
+    requires x86_abi<abi_t<T, N>>
+  {
+    return v0 += v1;
+  }
```
DenisYaroshevskiy, line 25: Why is this necessary? I mean the base case should still work, right?

include/eve/module/real/core/function/regular/simd/x86/add.hpp
```diff
@@ -0,0 +1,63 @@
...
+
+  // -----------------------------------------------------------------------------------------------
+  // Masked case
+  template<conditional_expr C, real_scalar_value T, typename N>
+  EVE_FORCEINLINE
+  wide<T, N> add_(EVE_SUPPORTS(sse2_), C const &cx, wide<T, N> const &v, wide<T, N> const &w) noexcept
```
DenisYaroshevskiy, line 31: Supports avx512?
Or `requires !abi_t<T, N>::is_wide_logical`?

include/eve/module/real/core/function/regular/simd/x86/div.hpp
```diff
@@ -18,4 +18,31 @@ namespace eve::detail
   {
     return v0 /= v1;
   }
```
DenisYaroshevskiy, line 20: Why is this needed for x86 special?

include/eve/module/real/core/function/regular/simd/x86/popcount.hpp
```diff
@@ -127,4 +128,90 @@ namespace eve::detail
...
   }
+
+  /////////////////////////////////////////////////////////////////////////////
+  // avx512
+  template<integral_scalar_value T,  typename N>
+  EVE_FORCEINLINE auto popcount_(EVE_SUPPORTS(avx512_), wide<T, N>  v) noexcept
+      requires x86_abi<abi_t<T, N>>
+  {
+    auto constexpr bitalg = spy::supports::avx512::bitalg_;
+    auto constexpr ppctdq = spy::supports::avx512::popcntdq_;
```
DenisYaroshevskiy, line 139: but we don't have this on CI.
If we don't have something on CI, adding it is a problem.

include/eve/module/real/core/function/saturated/simd/x86/add.hpp
```diff
@@ -47,4 +47,40 @@ namespace eve::detail
...
+      else if constexpr ( current_api >= avx2 && c == category::int16x16  ) return _mm256_mask_adds_epi16(src,m,v,w);
+      else if constexpr ( current_api >= avx2 && c == category::uint16x16 ) return _mm256_mask_adds_epu16(src,m,v,w);
+      else if constexpr ( current_api >= avx2 && c == category::int8x32   ) return _mm256_mask_adds_epi8 (src,m,v,w);
+      else if constexpr ( current_api >= avx2 && c == category::uint8x32  ) return _mm256_mask_adds_epu8 (src,m,v,w);
+      else if constexpr ( c == category::int16x8                          ) return _mm_mask_adds_epi16   (src,m,v,w);
+      else if constexpr ( c == category::uint16x8                         ) return _mm_mask_adds_epu16   (src,m,v,w);
+      else if constexpr ( c == category::int8x16                          ) return _mm_mask_adds_epi8    (src,m,v,w);
+      else if constexpr ( c == category::uint8x16                         ) return _mm_mask_adds_epu8    (src,m,v,w);
+      else                                                                  return if_else(cx,eve::add(st, v, w),src);
+    }
```
DenisYaroshevskiy, line 83: Why not fwd cpu?

DenisYaroshevskiy (commented):
Is everything here tested already?

include/eve/module/real/core/function/regular/simd/x86/add.hpp
jtlap: perhaps useless I did it as it was in sub

include/eve/module/real/core/function/regular/simd/x86/div.hpp
jtlap: don't know same answer, it was there

include/eve/module/real/core/function/regular/simd/x86/popcount.hpp
jtlap: ok I just wanted to see if it worked on CI, the PR was the to see if there were problems.
On my computer I have bitalg and popcntdq and I wanted to see if it was easy to add these 
At home it works enabling the options on the compiler and also just putting -march=skylake...

include/eve/module/real/core/function/saturated/simd/x86/add.hpp
jtlap: ok can do

jtlap:
In fact this is not finished yet I sent the PR just to see it worked outside of avx512 and there are still some mistakes
I am on it
But all features added are tested in the cpp files

include/eve/module/real/core/function/regular/simd/x86/popcount.hpp
DenisYaroshevskiy: What I mean is.

By avx512 we mean:
`AVX512 with skylake like minimum flavour (f,vl,dq,bw)`

https://jfalcou.github.io/eve/index.html

I'd prefer if we either stick by it or compile versions where it's disabled on CI as well.

Otherwise we have a code that's never tested, which means that the users might hit it and get runtime bugs.

include/eve/module/real/core/function/regular/simd/x86/add.hpp
DenisYaroshevskiy: Maybe remove it everywhere if you don't mind?
Just: less code - easier to maintain?

include/eve/module/real/core/function/regular/simd/x86/popcount.hpp
jtlap: ok I finish testing my stuffs and will remove the bitalg etc part,
but it is not never tested as it was tested on my computer who have it

include/eve/module/real/core/function/saturated/simd/x86/add.hpp
jtlap: done

include/eve/module/real/core/function/regular/simd/x86/popcount.hpp
DenisYaroshevskiy: I meant on ci. The code needs to be contiguous tested so that the new changes don't break old behaviour.

include/eve/module/real/core/function/regular/simd/x86/mul.hpp
```diff
@@ -13,10 +13,33 @@
...
+      return mul_(EVE_RETARGET(cpu_),cx,v,w);
+    }
+    else
+    {
+      auto src  = alternative(cx,v,as<wide<T, N>>{});
+      auto m    = expand_mask(cx,as<wide<T, N>>{}).storage().value;
+
+            if constexpr(c == category::float32x16) return _mm512_mask_mul_ps   (src,m,v,w);
+      else  if constexpr(c == category::float64x8 ) return _mm512_mask_mul_pd   (src,m,v,w);
+#ifndef SPY_COMPILER_IS_GCC
```
jfalcou, line 36: you can use `spy::compiler == spy::gcc` in a `if constexpr`

include/eve/module/real/core/function/regular/simd/x86/rec.hpp
```diff
@@ -13,53 +13,102 @@
...
+    else  if constexpr(c == category::float32x8  )
+    {
+      if constexpr(current_api >= avx2 )            return _mm256_rcp14_epi32(v);
+      else                                          return _mm256_rcp_ps(v);
+    }
+    else  if constexpr(c == category::int64x4 )
     {
-      // The error for this approximation is no more than 1.5.e-12
-      return _mm_cvtps_pd(_mm_rcp_ps(_mm_cvtpd_ps(a0)));
+      if constexpr(current_api >= avx2 )            return _mm256_rcp14_epi64(v);
```
jfalcou, line 31: This function doesn't exist.
The whole rcp is only available on real so why bother with integer ?

include/eve/module/real/core/function/regular/simd/x86/rec.hpp
```diff
@@ -13,53 +13,102 @@
...
+      else                                          return _mm256_cvtps_pd(_mm_rcp_ps(_mm256_cvtpd_ps(v)));
+    }
+    else  if constexpr(c == category::float32x4  )
+    {
+      if constexpr(current_api >= avx2 )            return _mm_rcp14_epi32(v);
```
jfalcou, line 36: Same for rcp14_epI32 and rcp14_epi64

include/eve/module/real/core/function/regular/generic/fma.hpp
```diff
@@ -37,4 +37,16 @@ namespace eve::detail
...
+
+  //================================================================================================
+  // Masked case
+  //================================================================================================
+  template<conditional_expr C, real_value T, real_value U, real_value V>
+  EVE_FORCEINLINE auto fma_(EVE_SUPPORTS(cpu_), C const &cond, T const &a, U const &b, V const &c) noexcept
+  requires properly_convertible<U, V, T>
+  {
+    using r_t =  common_compatible_t<T, U, V>;
+    return mask_op(  cond, eve::fma, r_t(a), r_t(b), r_t(c));
```
jfalcou, line 49: Why do you do a conversion here and not in the other fma like mask functions ?

jfalcou (changes requested):
I think we need to make all the if_else in mask just retarget to CPU just in case one day we change this strategy.

include/eve/module/real/core/function/regular/simd/x86/rec.hpp
jtlap: mistakes corrected

jfalcou (approved):

PR #826, opened 2021-07-25
Minmax based family

jtlap:
_mmxxx_range_xx are a family of avx512 intrinsics allowing max min extented ops as maxmag with same latency as min or max
this commit emulate these for other extensons and use range for avx512


include/eve/module/real/core/function/diff/detail/minmax_kernel.hpp
```diff
@@ -37,18 +38,92 @@ namespace eve::detail
...
+      else
+      {
+        r_t that(0);
+        getNth(std::make_index_sequence<sizeof...(args)>{},that,args...);
+        return if_else(m == eve::abs(that), aux(that), zero);
+      }
+    }
+  }
+
+//   template<int N, typename F, typename T0, typename T1, typename... Ts>
```
jfalcou, line 93: Remove comments

jfalcou (changes requested):

include/eve/module/real/core/function/diff/detail/minmax_kernel.hpp
jtlap: done

jfalcou (approved):

Issue #829, opened 2021-07-26
[BUG] building with new libc++


DenisYaroshevskiy:
Removing hacks and trying to build with the latest libc++

DenisYaroshevskiy:
https://github.com/jfalcou/eve/pull/830

PR #830, opened 2021-07-26
FIX-#829: compile with libcxx master

DenisYaroshevskiy:
Since we already don't build with libc++ from clang-12 - just upgraded to successfully build with clang master.


jfalcou:
shouldn't the whole details/concepts.hpp file just disappear ?

include/eve/detail/kumi.hpp
```diff
@@ -6,6 +6,7 @@
 **/
 //==================================================================================================
 #pragma once
+#include <concepts>
```
jfalcou, line 9: we do need it here ?

DenisYaroshevskiy: We do yes. `std::convertible_to` https://en.cppreference.com/w/cpp/concepts/convertible_to

PR #834, opened 2021-07-28
FIX-715: zip iterator

DenisYaroshevskiy:
A step to equality.

zip iterator is again our type of iterator, where the interface is designed for convenience of the algorithm writer.
There will be `zip` - a fairly stupid helper and the preprocess range will learn to generate from `zip` `zip_iterator`s

### DIFF
--- a/include/eve/algo/zip_iterator.hpp
+++ b/include/eve/algo/zip_iterator.hpp
@@ -47,17 +47,17 @@ namespace eve::algo
       : storage{i, is...}
     {}
 
+    operator zip_iterator<unaligned_t<I>, unaligned_t<Is> ...>() const
+      requires (!unaligned_iterator<I>) || ((!unaligned_iterator<Is>) || ...)
+    {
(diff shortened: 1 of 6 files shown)


include/eve/algo/zip_iterator.hpp
```diff
@@ -83,6 +83,47 @@ namespace eve::algo
       }
     }
 
+    template <typename _Cardinal>
+    auto cardinal_cast(_Cardinal N) const
+    {
+      return zip_iterator<decltype(I{}.cardinal_cast(N)), decltype(Is{}.cardinal_cast(N))...> {
```
jfalcou, line 89: can't we use some automatic template type deduction here?

DenisYaroshevskiy: I don't think so.

`zip_iterator` will refer to this specific instantiation.

jfalcou (approved):

PR #838, opened 2021-08-02
FIX-836: tmped better traits support

DenisYaroshevskiy:
* Put all the traits support in one place.
* added ability to `[]` one trait.
* updated some docs a tiny bit.


jfalcou (approved):
LGTM 
Two questions :
  - do we still need instance_of concept ?
  - do we need a traits concept ?

PR #839, opened 2021-08-03
Avx512cmp

jtlap:
masked compares

### DIFF
--- a/include/eve/detail/function/conditional.hpp
+++ b/include/eve/detail/function/conditional.hpp
@@ -66,7 +66,7 @@ namespace eve::detail
   // What's the logical alternative for this case ?
   //================================================================================================
   template<conditional_expr C, typename Target>
-  EVE_FORCEINLINE auto lalternative(C const& c, as<Target> const&)
+  EVE_FORCEINLINE auto logical_alternative(C const& c, as<Target> const&)
   {
     if constexpr( C::has_alternative )  return Target{c.alternative};
     else                                return false_(as<Target>());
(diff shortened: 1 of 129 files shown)


DenisYaroshevskiy (commented):
I'm not against this but I have to ask why.
What would be the use-case.

jtlap:
frankly the use case is not clear but avx512 has the intrinsics that do the masking with the same lattency as the mere predicate,
so I braugth the interface to eve as we have this for many other function objects.

jtlap:
I have forgotten to add the doc, I'm on it.

include/eve/function/is_less.hpp
```diff
@@ -9,6 +9,8 @@
 
 #include <eve/detail/implementation.hpp>
 #include <eve/traits/as_logical.hpp>
+#include <eve/constant/false.hpp>
```
jfalcou, line 12: why those includes ?

jfalcou (changes requested):

include/eve/function/is_less.hpp
jtlap: suppressed

include/eve/detail/function/conditional.hpp
```diff
@@ -60,4 +61,36 @@ namespace eve::detail
                                                                           , alternative(c,a0,eve::as<r_t>{})
                                                                           );
   }
+
+  //================================================================================================
+  // What's the logical alternative for this case ?
+  //================================================================================================
+  template<conditional_expr C, typename Target>
+  EVE_FORCEINLINE auto lalternative(C const& c, as<Target> const&)
```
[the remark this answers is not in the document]
jtlap: done

jfalcou (approved):

PR #850, opened 2021-08-15
Use Javascript to dynamically generate Compiler Explorer links

jfalcou:


docs/v2/html/godbolt.js
```diff
@@ -0,0 +1,59 @@
...
+                                            "libs": [ {"id": "eve", "version": "trunk"}],
+                                            "options": "-O3 -std=c++20 -msse4.2"
+                                        }
+                                      }
+                                      ,{
+                                        "compiler":
+                                        {
+                                            "id": "clang1201",
+                                            "libs": [ {"id": "eve", "version": "trunk"}],
+                                            "options": "-O3 -std=c++20 -msse4.2"
```
DenisYaroshevskiy, line 39: `mavx2`?

jfalcou: why not ^^

Issue #719, opened 2021-05-19
[FEATURE] `zip`/  very minimal `equal`


DenisYaroshevskiy:
Requires: https://github.com/jfalcou/eve/issues/718

We want the basic implementation of `eve::algo::equal` (equivalent to `std::equal`).
The implementation at this point should be something like:

```
bool operator()(f, l, f1, pred) const {
   zip r {f, l, f1};
   return eve::algo::any(r.begin(), r.end(), [&](auto xy) { auto[x, y] = xy; return eve::not_(pred(x, y)); });
}

bool operator()(I1 f, S1 l, I2 f1) const {
   /*convert to common iterator type I1, I2*/
   return operator()(f, l, f1, eve::equal);
}
```

`zip` under the hood should just probably be two tuples, or tuple + size, not `zip_iterator` because it can't be.

P.S. equal of course should call `all` and not `any` but let's make the whole thing work first.

DenisYaroshevskiy:
Now done

PR #855, opened 2021-08-17
FIX-#719:  zip_iterator for relaxed iterators. Up to missmatch prototype

DenisYaroshevskiy:
Got `zip_iterator` to the point where mismatch is visible and actually wrote a prototype for it.
Everything that has to do with `preprocess_range` for `zip_iterator` is so far a complete nonsense, I just wanted it to work.

List of changes:
* Removed `forward_to_unaligned`. Seems to break.
* `unalign` free function. Just a tiny helper.
* `++/--` in `operations_with_distance`. zip_iterator tests cover these.
* `zip_iterator` will now work with practically anything. The enablement of `eve::algo::iterator` functions will only happen if the components actually support it.
* wrote a prototyped `preprocess_range` for `zip_iterator` only for case when the sentinel is for the first iterator in a group.
* tested a prototype missmatch.

The next step will be to do `zip` with proper `preprocess_range` that supports all the necessary things.


include/eve/algo/ptr_iterator.hpp
```diff
@@ -38,7 +38,7 @@ namespace eve::algo
     template <typename _Cardinal>
     auto cardinal_cast(_Cardinal) const { return unaligned_ptr_iterator<T, _Cardinal>{ptr}; }
 
-    value_type& operator*() const { return *ptr; }
+    auto& operator*() const { return *ptr; }
```
jfalcou, line 41: why this change ?

jfalcou (approved):

include/eve/algo/ptr_iterator.hpp
DenisYaroshevskiy: Because `value_type` for `const int*` is `int` but the reference type is `const int&`

PR #861, opened 2021-08-20
Ordering operators opt-in support for product type UDT

jfalcou:
we provide a cacth-all trait that enables SIMD ordering on a given product_type-like user-defined type. 
This makes error handling simpler and the amount of code to write in the common case lower.

### DIFF
--- a/include/eve/arch/cpu/wide.hpp
+++ b/include/eve/arch/cpu/wide.hpp
@@ -891,7 +891,7 @@ namespace eve
     // Logical operations
     //==============================================================================================
     //! @brief Element-wise equality comparison of two eve::wide
-    friend  EVE_FORCEINLINE auto operator==(wide v, wide w) noexcept
+    friend EVE_FORCEINLINE auto operator==(wide v, wide w) noexcept
     {
       return detail::self_eq(v,w);
     }
@@ -932,6 +932,7 @@ namespace eve
 
     //! @brief Element-wise less-than comparison between eve::wide
     friend EVE_FORCEINLINE auto operator<(wide v, wide w) noexcept
+    requires( !kumi::product_type<Type> )
     {
       return detail::self_less(v,w);
     }
@@ -939,6 +940,7 @@ namespace eve
     //! @brief Element-wise less-than comparison between a eve::wide and a scalar
     template<scalar_value S>
     friend EVE_FORCEINLINE auto operator<(wide v, S w) noexcept
+    requires( !kumi::product_type<Type> )
     {
       return v < wide{w};
     }
@@ -946,12 +948,14 @@ namespace eve
     //! @brief Element-wise less-than comparison between a scalar and a eve::wide
     template<scalar_value S>
     friend EVE_FORCEINLINE auto operator<(S v, wide w) noexcept
+    requires( !kumi::product_type<Type> )
     {
       return wide{v} < w;
     }
 
     //! @brief Element-wise greater-than comparison between eve::wide
     friend EVE_FORCEINLINE auto operator>(wide v, wide w) noexcept
+    requires( !kumi::product_type<Type> )
     {
       return detail::self_greater(v,w);
     }
@@ -959,6 +963,7 @@ namespace eve
     //! @brief Element-wise greater-than comparison between a eve::wide and a scalar
     template<scalar_value S>
     friend EVE_FORCEINLINE auto operator>(wide v, S w) noexcept
+    requires( !kumi::product_type<Type> )
     {
       return v > wide{w};
     }
@@ -966,12 +971,14 @@ namespace eve
     //! @brief Element-wise greater-than comparison between a scalar and a eve::wide
     template<scalar_value S>
     friend EVE_FORCEINLINE auto operator>(S v, wide w) noexcept
+    requires( !kumi::product_type<Type> )
     {
       return wide{v} > w;
     }
 
     //! @brief Element-wise greater-or-equal comparison between eve::wide
     friend EVE_FORCEINLINE auto operator>=(wide v, wide w) noexcept
+    requires( !kumi::product_type<Type> )
     {
       return detail::self_geq(v,w);
     }
@@ -979,6 +986,7 @@ namespace eve
     //! @brief Element-wise greater-or-equal comparison between a eve::wide and a scalar
     template<scalar_value S>
     friend EVE_FORCEINLINE auto operator>=(wide v, S w) noexcept
+    requires( !kumi::product_type<Type> )
     {
       return v >= wide{w};
     }
@@ -986,12 +994,14 @@ namespace eve
     //! @brief Element-wise greater-or-equal comparison between a scalar and a eve::wide
     template<scalar_value S>
     friend EVE_FORCEINLINE auto operator>=(S v, wide w) noexcept
+    requires( !kumi::product_type<Type> )
     {
       return wide{v} >= w;
     }
 
     //! @brief Element-wise less-or-equal comparison between eve::wide
     friend EVE_FORCEINLINE auto operator<=(wide v, wide w) noexcept
+    requires( !kumi::product_type<Type> )
     {
       return detail::self_leq(v,w);
     }
@@ -999,6 +1009,7 @@ namespace eve
     //! @brief Element-wise less-or-equal comparison between a eve::wide and a scalar
     template<scalar_value S>
     friend EVE_FORCEINLINE auto operator<=(wide v, S w) noexcept
+    requires( !kumi::product_type<Type> )
     {
       return v <= wide{w};
     }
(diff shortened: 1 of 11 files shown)


include/eve/detail/function/simd/common/friends.hpp
```diff
@@ -152,10 +153,7 @@ namespace eve::detail
     }
     else
     {
-      auto zipped = kumi::zip(v.storage(),w.storage());
-      return kumi::fold_right ( [](auto acc , auto e) { return acc && ( get<0>(e) == get<1>(e) ); }
-                              , zipped, true
-                              );
+      return v.storage() == w.storage();
```
DenisYaroshevskiy, line 156: What does this actually do?
I mean, what would `kumi::tuple::operator==` return here?

include/eve/detail/kumi.hpp
```diff
@@ -485,6 +486,47 @@ namespace kumi
       (std::make_index_sequence<sizeof...(Ts)>());
     }
 
+    template<product_type Other>
+    friend constexpr auto operator<(tuple const &lhs, Other const &rhs) noexcept
```
DenisYaroshevskiy, line 490: I thought you were dead set against this?
If you are on board with it here, I'd suggest to just generate less.
I don't think it's a bad idea to have a default less.

I know you feel different and we can do that too, but maybe be consistent?

include/eve/detail/function/simd/common/friends.hpp
DenisYaroshevskiy: What type will it be, for example, for `wide<pair<int, double>`

include/eve/detail/kumi.hpp
```diff
@@ -485,6 +486,47 @@ namespace kumi
...
+      }
+      (std::make_index_sequence<sizeof...(Ts)-1>());
+
+      return res;
+    }
+
+    template<product_type Other>
+    friend constexpr auto operator<=(tuple const &lhs, Other const &rhs) noexcept
+    {
+      return (lhs < rhs) || (lhs == rhs);
```
DenisYaroshevskiy, line 515: `!(rhs < lhs)`

include/eve/detail/kumi.hpp
```diff
@@ -485,6 +486,47 @@ namespace kumi
...
+    }
+
+    template<product_type Other>
+    friend constexpr auto operator>(tuple const &lhs, Other const &rhs) noexcept
+    {
+      return !(lhs <= rhs);
```
DenisYaroshevskiy, line 521: `rhs < lhs`

include/eve/detail/kumi.hpp
DenisYaroshevskiy: Or is it some Nan thing? I don't think that's a valid concern to sabotage perf for everyone.

test/unit/api/udt/udt.hpp
```diff
@@ -43,34 +35,17 @@ namespace udt
...
 // Opt-in for eve::product_type + adaptation
-template<>              struct eve::is_product_type<udt::grid2d>  : std::true_type {};
-template<>              struct std::tuple_size<udt::grid2d>       : std::integral_constant<std::size_t, 2> {};
-template<std::size_t I> struct std::tuple_element<I,udt::grid2d> { using type = int; };
+template<>              struct eve::is_product_type<udt::grid2d>    : std::true_type {};
+template<>              struct std::tuple_size<udt::grid2d>         : std::integral_constant<std::size_t, 2> {};
+template<std::size_t I> struct std::tuple_element<I,udt::grid2d>    { using type = int; };
+
+// Opt-in for SIMD < > <= >= operators
+template<>  struct eve::supports_ordering<udt::grid2d>  : std::true_type {};
```
DenisYaroshevskiy, line 51: Can't we just like inherit smth or have an internal typedef for simplicity?

test/unit/api/udt/udt.hpp
```diff
@@ -43,34 +35,17 @@ namespace udt
...
-  }
 }
 
 // Opt-in for eve::product_type + adaptation
-template<>              struct eve::is_product_type<udt::grid2d>  : std::true_type {};
-template<>              struct std::tuple_size<udt::grid2d>       : std::integral_constant<std::size_t, 2> {};
-template<std::size_t I> struct std::tuple_element<I,udt::grid2d> { using type = int; };
+template<>              struct eve::is_product_type<udt::grid2d>    : std::true_type {};
+template<>              struct std::tuple_size<udt::grid2d>         : std::integral_constant<std::size_t, 2> {};
+template<std::size_t I> struct std::tuple_element<I,udt::grid2d>    { using type = int; };
```
DenisYaroshevskiy, line 48: I really thing types like should just iherit tuple.

include/eve/detail/kumi.hpp
DenisYaroshevskiy: Nah, you do 'not' in other places, so it's not a Nan thing.
You can probably get an efficient Nan correctness if you just do a fold in `eve`

include/eve/detail/function/simd/common/friends.hpp
jfalcou: The == in kumi just do what this complicated zip does in a simpler way cause it has access to the element via ...
The type return is what `as_logical_t<wide<pair<int,double>>` does, a `logical<wide<int>>`

include/eve/detail/kumi.hpp
jfalcou: Beware, this is inside KUMI, not inside EVE.
KUMI actually has an issue about this (https://github.com/jfalcou/ofw/issues/20).

What I am is being deadset on having `operator<` on user-defined type defined by default because the user knows if their own type is OK to be ordered.

On tuple with no semantic, it is ok to have it.

include/eve/detail/kumi.hpp
jfalcou: no this is not the same

test/unit/api/udt/udt.hpp
jfalcou: If we force people to inherit from something, this means we can't adapt types we don't write. For ex., we could not adapt say QPoint or some OpenGL types.

Now, we ma have both, but it is important we have a non-intrusive option.

test/unit/api/udt/udt.hpp
jfalcou: You are allowed to do it this way, again, it is made so we can adapt types we can't change.

include/eve/detail/kumi.hpp
jfalcou: `<=` is not equivalent to `! <`

The idea here was that, as KUMI needs the ordering, we can just piggy-back it.
I admit I have been slightly lazy here and should have written properly optimized >

DenisYaroshevskiy: This is really not an OK implementation for eve. Should be done properly.

a <= b is 100% equivalent !(b < a). Unless you are talking Nan.

include/eve/detail/kumi.hpp
DenisYaroshevskiy: It is

test/unit/api/udt/udt.hpp
DenisYaroshevskiy: Sure, but let's have an easy option too.

test/unit/api/udt/udt.hpp
DenisYaroshevskiy: Inheritance usecase is an important one and is the default one. Let's test that.

include/eve/detail/kumi.hpp
jfalcou: misread duh

include/eve/detail/kumi.hpp
jfalcou: ok I'm stupid, it's a lexical sort. Gonna fix it

include/eve/detail/kumi.hpp
jfalcou: Fixed

include/eve/detail/kumi.hpp
jfalcou: Fixed

test/unit/api/udt/udt.hpp
jfalcou: The easy option is now that we have ordering by default. Other operators will have a dual traits/typedef system

test/unit/api/udt/udt.hpp
jfalcou: Added test for inheritance

include/eve/arch/cpu/wide.hpp
```diff
@@ -931,81 +931,93 @@ namespace eve
     }
 
     //! @brief Element-wise less-than comparison between eve::wide
-    friend EVE_FORCEINLINE logical<wide> operator<(wide v, wide w) noexcept
+    friend EVE_FORCEINLINE auto operator<(wide v, wide w) noexcept
+    requires( !kumi::product_type<Type> )
```
DenisYaroshevskiy, line 935: I thought we decided that you need to opt out?

DenisYaroshevskiy (commented):
One comment

include/eve/arch/cpu/wide.hpp
jfalcou: Simplified later on

DenisYaroshevskiy (approved):

PR #862, opened 2021-08-20
Don't detect AVX512 if we don't have proper setup

jfalcou:
Manually unselect AVX512 related settings if we don't have at least Skylake like architecture

### DIFF
--- a/include/eve/arch/x86/predef.hpp
+++ b/include/eve/arch/x86/predef.hpp
@@ -45,4 +45,18 @@
 #  define EVE_SUPPORTS_NATIVE_SIMD
 #  define EVE_HW_X86
 #  define EVE_INCLUDE_X86_HEADER
+
+// Don't trigger AVX512 if we don't have at least Skylake support
+#  if defined(SPY_SIMD_IS_X86_AVX512)
+#   if !(   defined(__AVX512BW__) && defined(__AVX512CD__)  \
+        &&  defined(__AVX512DQ__) && defined(__AVX512VL__)  \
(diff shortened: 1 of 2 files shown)


include/eve/arch/x86/predef.hpp
```diff
@@ -45,4 +45,17 @@
 #  define EVE_SUPPORTS_NATIVE_SIMD
 #  define EVE_HW_X86
 #  define EVE_INCLUDE_X86_HEADER
+
+// Don't trigger AVX512 if we don't have at least Skylake support
+#  if defined(SPY_SIMD_IS_X86_AVX512)
+#   if !defined(__AVX512BW__) && !defined(__AVX512CD__) \
+    && !defined(__AVX512DQ__) && !defined(__AVX512VL__)
```
DenisYaroshevskiy, line 52: Docks say:
AVX512 with skylake like minimum flavour (f,vl,dq,bw)

this does not match.

SKYLAKE on AVX512 Wikipedia page says:

```
F, CD, ER, PF
VL, DQ, BW
```

Should this be the test? https://en.wikipedia.org/wiki/AVX-512

Update the doc maybe too?

jfalcou: F is implied by IS_AVX86_AVX512. The ER and PF are indeed missing

jfalcou: In fact, it is not. ER and PF are old Xeon Phi optiosn and are discontinued.
The if logic was wrong though and has been fixed

DenisYaroshevskiy (approved):

PR #866, opened 2021-08-22
Docv2/i objects

jtlap:


include/eve/function/is_not_nan.hpp
```diff
@@ -49,7 +49,7 @@ namespace eve
   //!~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ c++
   //!logical<T> r;
   //!if   constexpr(floating_real_value<T>) r = is_equal(x, x);
-  //!else constexpr(integral_real_value<T>) r = True<T>());
+  //!else constexpr(integral_real_value<T>) r = true_(as<T>());
```
jfalcou, line 52: could be better if all those `as<T>()` would be replaced by `as(x)`

jfalcou: here and in all other files

jtlap: done

jfalcou (approved):

PR #871, opened 2021-08-23
p's +quadrant

jtlap:


include/eve/function/quadrant.hpp
```diff
@@ -33,14 +33,14 @@ namespace eve
...
-  //!`x`:   [value](@ref eve::value).
+  //!`x`:   [real_value](@ref eve::real_value).
   //!
   //! **Return value**
   //!
   //!return a value of the same type as `x` containing the quadrant number where `x` (supposed an integer) lies.
   //!
   //!@warning
-  //!   the input `x` must be an integral or floating integral value.
+  //!   the input `x` must be an integral or a flint value.
```
jfalcou, line 43: Can we add flint to the glossary? Not sure everybody knows what it means.

PR #876, opened 2021-08-24
FIX-#719: zip, end-to-end

DenisYaroshevskiy:
Got to a point where `missmatch` looks sort of right:

```
  auto found = eve::algo::find_if(eve::algo::zip(v1, v2), [](auto x1_x2) {
    auto [x1, x2] = x1_x2;
    return x1 != x2;
  });
```

Specific fixes are:
* `zip_iterator` can now form a range with both match for the first iterator and a complete zip_iterator. Not sure if it's the correct behaviour, but I thought it was at some point.
* cleaned up some things in `zip_iterator`.
* `zip` is now a thing: if it takes all iterators - returns an iterator. If it takes at least one range, will return a range.

Next step is to do all of the proper deductions, like cardinal and such traits.

**NOTE:**
It makes sense in the context of what I'm doing that `zip` is non-owning.
But is it confusing in the C++ sense? Should it be called smth else?

### DIFF
--- a/include/eve/algo/zip.hpp
+++ b/include/eve/algo/zip.hpp
@@ -1,0 +1,126 @@
+//==================================================================================================
+/*
+  EVE - Expressive Vector Engine
+  Copyright : EVE Contributors & Maintainers
+  SPDX-License-Identifier: MIT
+*/
+//==================================================================================================
+#pragma once
+
+#include <eve/algo/as_range.hpp>
+#include <eve/algo/preprocess_range.hpp>
+#include <eve/algo/traits.hpp>
+#include <eve/algo/zip_iterator.hpp>
+
+#include <eve/algo/detail/preprocess_zip_range.hpp>
+
+#include <eve/detail/kumi.hpp>
+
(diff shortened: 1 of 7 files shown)


include/eve/algo/zip.hpp
```diff
@@ -0,0 +1,126 @@
...
+      return preprocess_zip_range(traits, self.ranges);
+    }
+  };
+
+  template <typename ...Rngs>
+  zip_range(kumi::tuple<Rngs...>) -> zip_range<Rngs...>;
+
+  namespace detail
+  {
+    // FIX-874: we should do this proper
```
jfalcou, line 48: DId you check what is in `eve/concept/range.hpp` ?

DenisYaroshevskiy: That's not quite what I want.

This should be `eve::algo::light_range`.

For example `eve::aligned_ptr` is not an iterator but it should be ok for `light_range`

jfalcou (approved):

PR #877, opened 2021-08-24
t's and zeta

jtlap:


include/eve/constant/twotonmb.hpp
```diff
@@ -27,7 +27,7 @@ namespace eve
   //!
   //! | Member       | Effect                                                     |
   //! |:-------------|:-----------------------------------------------------------|
-  //! | `operator()` | Computes the twotonmb constant                               |
+  //! | `operator()` | Computes the twotonmb constant                             |
```
jfalcou, line 30: Can't we just spell it out `the two to the nbmantissabits power`? It looks hyper self-referential in this state.

jfalcou (changes requested):

include/eve/constant/twotonmb.hpp
jtlap: done

jfalcou (approved):

PR #893, opened 2021-08-29
FIX-884: mismatch 2 ranges test

DenisYaroshevskiy:
Wrote a two range iterating test.
Testing this types of algorithms is always tricky, we really should try to get approved for open fuzz and set up fuzzing.

### DIFF
--- a/test/unit/algo/mismatch.cpp
+++ b/test/unit/algo/mismatch.cpp
@@ -10,6 +10,10 @@
 
 #include <eve/algo/mismatch.hpp>
 
+#include "mismatch_generic_test.hpp"
+
+#include <set>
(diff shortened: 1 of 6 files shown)


test/unit/algo/mismatch.cpp
```diff
@@ -10,6 +10,10 @@
 
 #include <eve/algo/mismatch.hpp>
 
+#include "mismatch_generic_test.hpp"
+
+#include <set>
+#include <random>
```
jfalcou, line 16: why this include ?

DenisYaroshevskiy: Will remove in the next patch

PR #908, opened 2021-09-02
Fix 896/product type support 1

DenisYaroshevskiy:
DON'T LOOK AT THIS


DenisYaroshevskiy:
Everything is cleaned up in different prs. Nothing is left.

PR #909, opened 2021-09-03
FIX-896: patch 1.

DenisYaroshevskiy:
First batch of changes to do iteration over objects.

* value_type - a type function that deduces a value type for a range or iterator
* removed <=> from a baseclass helper. Yeah we can't default, but defaulting didn't really work. And it always was selected by accident.
* minimum common type support. At least `tuple<int, int, int, int>` will convert to `line`.
* `preprocess_range`: extracted a proper predeclaration, otherwise the dependencies stumble each other.

### DIFF
--- a/include/eve/algo/concepts/value_type.hpp
+++ b/include/eve/algo/concepts/value_type.hpp
@@ -1,0 +1,37 @@
+//==================================================================================================
+/*
+  EVE - Expressive Vector Engine
+  Copyright : EVE Contributors & Maintainers
+  SPDX-License-Identifier: MIT
+*/
+//==================================================================================================
+#pragma once
+
+#include <eve/algo/concepts/detail.hpp>
+
+#include <eve/traits.hpp>
+
+#include <type_traits>
+
+namespace eve::algo
+{
+  namespace detail
+  {
+    template <typename T>
+    constexpr auto value_type_impl()
+    {
+           if constexpr ( has_begin_end<T>            ) return value_type_impl<decltype(std::declval<T>().begin())>();
+      else if constexpr ( std::contiguous_iterator<T> ) return std::type_identity<typename std::iterator_traits<T>::value_type>{};
+      else                                              return std::type_identity<typename eve::pointer_traits<T>::value_type>{};
+    }
+  }
+
+  template <typename T>
+  struct value_type
+  {
+    using type = typename decltype(detail::value_type_impl<T>())::type;
+  };
+
+  template <typename T>
+  using value_type_t = typename value_type<T>::type;
+}
--- a/include/eve/algo/preprocess_range.hpp
+++ b/include/eve/algo/preprocess_range.hpp
@@ -7,10 +7,110 @@
 //==================================================================================================
 #pragma once
 
+#include <eve/algo/converting_iterator.hpp>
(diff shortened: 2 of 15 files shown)


test/unit/algo/iterator_concept_test.hpp
```diff
@@ -88,9 +88,13 @@ namespace algo_test
     TTS_TYPE_IS(typename decltype(res)::cardinal, eve::fixed<1>);
   }
 
+  // void is_relaxed_test(eve::algo::relaxed_iterator auto, eve::algo::relaxed_iterator auto) {}
+
   template <eve::algo::readable_iterator I, eve::algo::sentinel_for<I> S, typename T, typename ReplaceIgnored>
   void iterator_sentinel_test(I f, S l, T v, ReplaceIgnored replace)
   {
+    eve::algo::preprocess_range(eve::algo::traits{}, f, f);
+    // is_relaxed_test(f, l);
```
DenisYaroshevskiy, line 97: Auch. Wil fix in the next patch if no other comments.

include/eve/traits/common_type.hpp
```diff
@@ -63,6 +62,15 @@ namespace eve::detail
...
   };
 
+  template <typename T>
+  struct is_kumi_tuple : std::false_type {};
+
+  template <typename ...Ts>
+  struct is_kumi_tuple<kumi::tuple<Ts...>> : std::true_type {};
+
+  template <typename T>
+  concept is_kumi_tuple_v = is_kumi_tuple<T>::value;
```
jfalcou, line 72: why not using kumi::product_type ?

jfalcou (changes requested):

include/eve/algo/concepts/value_type.hpp
```diff
@@ -0,0 +1,37 @@
...
+    constexpr auto value_type_impl()
+    {
+           if constexpr ( has_begin_end<T>            ) return value_type_impl<decltype(std::declval<T>().begin())>();
+      else if constexpr ( std::contiguous_iterator<T> ) return std::type_identity<typename std::iterator_traits<T>::value_type>{};
+      else                                              return std::type_identity<typename eve::pointer_traits<T>::value_type>{};
+    }
+  }
+
+  template <typename T>
+  struct value_type
```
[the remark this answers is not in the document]
DenisYaroshevskiy: Will check it out

include/eve/algo/preprocess_range.hpp
```diff
@@ -7,10 +7,110 @@
 //==================================================================================================
 #pragma once
 
+#include <eve/algo/converting_iterator.hpp>
+#include <eve/algo/detail/preprocess_range.hpp>
```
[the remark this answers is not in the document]
DenisYaroshevskiy: Because I need convert in preprocess_rangr and preprocess_range roughly everywhere.
Like a `relaxed_range` is something that has preprocess range defined for it, that kind of stuff.

include/eve/traits/common_type.hpp
DenisYaroshevskiy: It's not the same check.  I need to prefer a product type with semantics to kumi tuple

DenisYaroshevskiy:
Agreed that I'll do `value_type` in the next patch.

Issue #906, opened 2021-09-02
[BUG] convert and traits support.


DenisYaroshevskiy:
This is a tricky tricky problem.

What should `tagged_dispatch(preprocess_range_, auto traits, convert_range<R, T>)` do?
Probably if the traits have a specific type indication - just propagate.
Otherwise, ask for a specific type for a subrange and then drop it.

Issue #880, opened 2021-08-25
[BUG] Using `zip` and `common_with_types`


DenisYaroshevskiy:
Example:

```
  soa_vector<low_resolution_point> rough_idea;
  soa_vector<high_resolution_point> prev_result;
  
  eve::algo::transform[eve::algo::common_with_types<high_resolution_point>](
      eve::algo::zip(rough_idea, prev_result), 
      [](auto rough, auto prev) {
        ///
      }); 
```

This won't work at the moment, because processing `zip_range` will think that `common_with_types` is for it and will try to apply it for every element, which will fail because there is no conversion from a point field to `high_resolution_point`.

We need to be able to remove a field from traits in order to handle it.

Needs this: https://github.com/jfalcou/ofw/issues/51

The consequence of this is that `mismatch` doesn't really support traits at the moment.

PR #914, opened 2021-09-04
FIX-896: product type support, convert range

DenisYaroshevskiy:
Still no product type support so far.
I have implemented `convert` in a different way though which is a step.

* introduced `range_ref` - a function that  takes a reference to a range. By default just stores a pointer to the original one, if `using non_owning = void` is specified will copy by value. Generally speaking, it's enough to just make `as_range` non owning but it removes a bunch of extra type wrappers that don't do anything.
* fixed the `<=>` bug. THere is no `operator<=>` for vector iterators, had to improvise.
* Now support `convert` for any `relaxed_iterator` and `relaxed_range`, not just `eve::algo::iterator`
* `zip_range` no longer has traits attached! Instead `zip` immediately converts all of the members.

This one will be fixed: https://github.com/jfalcou/eve/issues/906
This one isn't be visible in any shape so far (after this patch): https://github.com/jfalcou/eve/issues/880
There might be a case at some point, but it will need different traits for that. Let's wait for a usecase.


test/unit/algo/converting_eve_iterator.cpp
```diff
@@ -1,14 +1,6 @@
-//==================================================================================================
```
jfalcou, line 1: why the name change for the files ?

DenisYaroshevskiy: This one: because I don't test the non-eve iterator in this file. 
There is also now a convert.cpp to test the rest of the logic.

jfalcou (approved):

PR #919, opened 2021-09-05
Optimize/clarify relation between conditional and top_bits

jfalcou:
+ Make to_logical generate less branches
+ Make all conditional `.mask()` use top_bits to factorize code

### DIFF
--- a/include/eve/conditional.hpp
+++ b/include/eve/conditional.hpp
@@ -175,8 +175,29 @@ namespace eve
 
     template<typename T> EVE_FORCEINLINE as_logical_t<T> mask(eve::as<T> const&) const
     {
+      using abi_t = typename T::abi_type;
       using type  = as_logical_t<T>;
-      return to_logical(detail::top_bits<type>(*this));
+
+      if constexpr( !abi_t::is_wide_logical )
+      {
+        return detail::top_bits<type>(*this).storage;
+      }
+      else
+      {
+        using i_t   = as_integer_t<typename type::mask_type>;
+        if constexpr(eve::use_complete_storage<type>)
+        {
+          return bit_cast(detail::linear_ramp(eve::as<i_t>()) < i_t(count_), as<type>());
+        }
+        else
+        {
+          // Use the most full type to be sure to fill outside values of small wide with false
+          using e_t   = eve::element_type_t<i_t>;
+          using abi_t = typename i_t::abi_type;
+          using w_t   = eve::wide<e_t, eve::expected_cardinal_t<e_t, abi_t> >;
+          return bit_cast(detail::linear_ramp(eve::as<w_t>()) < w_t(count_), as<type>());
+        }
+      }
     }
 
     constexpr bool friend operator==(keep_first const&, keep_first const&) = default;
@@ -217,10 +238,10 @@ namespace eve
 
     template<typename V> EVE_FORCEINLINE auto else_(V const& v) const  {  return or_(*this,v);  }
 
-    template<typename T> EVE_FORCEINLINE as_logical_t<T> mask(eve::as<T> const&) const
+    template<typename T> EVE_FORCEINLINE as_logical_t<T> mask(eve::as<T> const& tgt) const
     {
-      using type  = as_logical_t<T>;
-      return to_logical(detail::top_bits<type>(*this));
+      constexpr auto card = cardinal_v<T>;
+      return keep_first{card-count_}.mask(tgt);
     }
 
     constexpr bool friend operator==(ignore_last const&, ignore_last const&) = default;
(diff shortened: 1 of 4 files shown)


include/eve/detail/function/simd/x86/friends.hpp
```diff
@@ -146,11 +146,23 @@ namespace eve::detail
...
         else  if constexpr(use_avx2 && c == category::int8x32  )  return _mm256_cmpeq_epi8 (v,w);
         else  if constexpr(use_avx2 && c == category::uint8x32 )  return _mm256_cmpeq_epi8 (v,w);
-        else  if constexpr( c == category::int64x2  )             return map(eq,v,w);
+        else  if constexpr( c == category::int64x2  )
+        {
+          if constexpr(N::value == 2)
+            return as_logical_t<wide<T, N>>{ v.get(0) == w.get(0), v.get(1) == w.get(1) };
+          else
+            return as_logical_t<wide<T, N>>{ v.get(0) == w.get(0) };
+        }
```
DenisYaroshevskiy, line 155: Why is it better than just calling map?

include/eve/detail/function/simd/x86/friends.hpp
```diff
@@ -146,11 +146,23 @@ namespace eve::detail
...
         else  if constexpr( c == category::int16x8  )             return _mm_cmpeq_epi16(v,w);
         else  if constexpr( c == category::int8x16  )             return _mm_cmpeq_epi8 (v,w);
-        else  if constexpr( c == category::uint64x2 )             return map(eq,v,w);
+        else  if constexpr( c == category::uint64x2 )
+        {
+          if constexpr(N::value == 2)
+            return as_logical_t<wide<T, N>>{ v.get(0) == w.get(0), v.get(1) == w.get(1) };
+          else
+            return as_logical_t<wide<T, N>>{ v.get(0) == w.get(0) };
+        }
```
DenisYaroshevskiy, line 165: Same here.

include/eve/detail/function/simd/x86/friends.hpp
jfalcou: g++ has issue with this one map calls, this version produce the correct code

include/eve/detail/function/simd/x86/friends.hpp
```diff
@@ -136,7 +136,7 @@ namespace eve::detail
       else
       {
         constexpr auto use_avx2 = current_api >= avx2;
-        constexpr auto eq = []<typename E>(E e, E f) { return as_logical_t<E>(e == f); };
+        constexpr auto eq = []<typename E>(E const& e, E const& f) { return as_logical_t<E>(e == f); };
```
DenisYaroshevskiy, line 139: I an worrying about lambdas.
Is this for aggregate case only now?

It's not a new problem we can fix it when we get there.

include/eve/detail/top_bits.hpp
```diff
@@ -370,15 +385,34 @@ EVE_FORCEINLINE Logical to_logical(top_bits<Logical> mmask)
...
+    using abi_t     = typename bits_wide::abi_type;
+    using fit_wide  = logical<wide<bits_et, expected_cardinal_t<bits_et, abi_t>>>;
+
+    if constexpr(bits_wide::size() < fit_wide::size())
+    {
+      fit_wide card_mmask([&](int i, int) { return i < bits_wide::size(); });
+      fit_wide r = res.storage();
+      r = r && card_mmask;
+      res = r.storage();
+    }
```
DenisYaroshevskiy, line 400: I'm willing to accept this for to_logical, since it does not matter, but I don't think it's the right solution.

The smaller wides should fill in zeroes when initialising with `[](){}`.
This does not cost anything in the initialisation.

While this is an actual code.

include/eve/conditional.hpp
```diff
@@ -170,35 +175,8 @@ namespace eve
...
-        {
-          // Use the most full type to be sure to fill outside values of small wide with false
-          using e_t   = eve::element_type_t<i_t>;
-          using abi_t = typename i_t::abi_type;
-          using w_t   = eve::wide<e_t, eve::expected_cardinal_t<e_t, abi_t> >;
-
-          return bit_cast(detail::linear_ramp(eve::as<w_t>()) < w_t(count_), as<type>());
-        }
-      }
+      return to_logical(detail::top_bits<type>(*this));
```
DenisYaroshevskiy: Wait, this is inefficient!

include/eve/detail/function/simd/x86/friends.hpp
jfalcou: I'll probably do a pass on all those comparisons, make it manual for 64 bits integral and remove the lambda.
Next PR ?

DenisYaroshevskiy (approved):

PR #920, opened 2021-09-05
FIX- 896/product type support 5

DenisYaroshevskiy:
* convert bug
* force inline in convert
* zip traits changes
* read/write

### DIFF
--- a/include/eve/algo/concepts/eve_iterator.hpp
+++ b/include/eve/algo/concepts/eve_iterator.hpp
@@ -57,7 +57,7 @@ namespace eve::algo
     std::totally_ordered_with<I, unaligned_t<I>> &&
     std::totally_ordered_with<I, partially_aligned_t<I>> &&
     requires(I i, std::ptrdiff_t n) {
-       { std::declval<typename I::value_type>() };
+       typename I::value_type;
        { i += n } -> std::same_as<I&>;
        { i - i }  -> std::same_as<std::ptrdiff_t>;
        { i.unaligned() } -> detail::unaligned_check;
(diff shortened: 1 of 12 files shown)


include/eve/algo/concepts/eve_iterator.hpp
```diff
@@ -55,6 +57,7 @@ namespace eve::algo
     std::totally_ordered_with<I, unaligned_t<I>> &&
     std::totally_ordered_with<I, partially_aligned_t<I>> &&
     requires(I i, std::ptrdiff_t n) {
+       { std::declval<typename I::value_type>() };
```
jfalcou, line 60: you don't need the declval.
`typename I::value_type;` is enough.
See Type Requirements at https://en.cppreference.com/w/cpp/language/constraints

include/eve/algo/traits.hpp
```diff
@@ -106,12 +106,13 @@ template <typename Traits, typename T>
     rbr::get_type_t<Traits, force_cardinal_key,
     eve::expected_cardinal_t<iteration_type_t<Traits, T>>>;
 
-  template <typename User, typename Default>
-  constexpr auto default_to(traits<User> const& user, traits<Default> const& defaults)
+
+  inline constexpr auto default_to =
```
jfalcou, line 110: why a lambda instead of a function here ?

DenisYaroshevskiy: To be able to pass into kumi::fold_right.

I was thinking to do a struct but seemed like lambda is enough.

I want to clean up all functions at some point

jfalcou: OK

jfalcou (approved):

Issue #921, opened 2021-09-05
[FEATURE] rest of `std::find` like algorithms that are easy to do


DenisYaroshevskiy:
I mean:

```
all_of
none_of
find
equals
```

This will make it more work for us to change things but it's good for the conference to get a more complete set.

PR #922, opened 2021-09-06
FIX-#921: find like algorithms

DenisYaroshevskiy:
Since we have done the basic cases, making these ones is very easy.

As any copy-paste, this increases our maintenance: any change would have to be applied in more places.

However, given that the conference is coming, might be worth it.

### IMPORTANT DESIGN DECISION

So, unlike `std::equal` and `std::mismatch` we allow you to pass 2 ranges.
However, we assert (inside `zip`) that they are the same size.

In isolation, it's surprising.

But the thinking is:

`eve::algo::equal(a, b)` and `eve::algo::equal(zip(a, b))` should behave the same.

Are we OK with this?


include/eve/algo/find.hpp
```diff
@@ -94,4 +94,16 @@ namespace eve::algo
...
 
   inline constexpr auto find_if = function_with_traits<find_if_>[default_simple_algo_traits];
+
+  template <typename TraitsSupport>
+  struct find_ : TraitsSupport
+  {
+    template <relaxed_range Rng, typename T>
+    EVE_FORCEINLINE auto operator()(Rng&& rng, T v) const
+    {
+      return find_if[TraitsSupport::get_traits()](std::forward<Rng>(rng), [v](auto x) { return x == v; });
```
jfalcou, line 104: my instinct say we shoudl pass eve::is_equal here so we can benefir ftom EVE_FORCEINLINE

DenisYaroshevskiy:
Agreed that I'll do predicate fix in a follw up

PR #926, opened 2021-09-08
FIX-894: slide right for 2 wides.

DenisYaroshevskiy:


jfalcou (approved):

include/eve/detail/function/simd/arm/neon/slide_right.hpp
```diff
@@ -12,52 +12,60 @@
...
-    }
-    else if constexpr(Shift == 0)
-    {
-      return v;
-    }
+    // This works for smaller than 8 byte wides on a slide right with 0.
+    // I don't understand why but it does, I tested.
+    // The previous comment, that I didn't understand was:
+    //   The actual shift uses the expected cardinal so it has enough
+    //   to push through the unused space.
```
DenisYaroshevskiy, line 25: Can you explain to me why in the world does the extract works for 2 smaller wides on arm?

PR #927, opened 2021-09-09
Improve some bit operations

jfalcou:


include/eve/module/real/core/function/regular/generic/bit_ceil.hpp
```diff
@@ -50,15 +50,8 @@ namespace eve::detail
...
-        else
-        {
-
-          auto res = if_else(vle1, one(eve::as(v)) //TODO Optimize with one_ ?
-                        , if_else(tmpltv, tmp*2,  tmp));
-          return res;
-        }
-
+        if constexpr(scalar_value<T>) return T(tmpltv ? tmp*2:  tmp);
+        else                          return if_else(vle1, one, if_else(tmpltv, tmp*2,  tmp));
```
DenisYaroshevskiy, line 54: tmp + tmp is better than tmp * 2 in case the compiler fails to optimize

DenisYaroshevskiy (commented):
Drive by review :)

Issue #929, opened 2021-09-09
[FEATURE] Missing proper x86 scans


DenisYaroshevskiy:
As part of conference prep I'm not going to do an optimal version of all scans for all avx's difficult shuffles.
This is the story to fill in the blanks.

PR #930, opened 2021-09-09
FIX-894: scan(wide, plus, zero)

DenisYaroshevskiy:
* changed `eve::zero` to map for product types. This sort of makes sense and is needed.
* saw missing force inline lambdas in a couple of places.
* wrote the basic scan implementation
* wrote an avx2 specific scan implementation. I actually don't know if it's better, but that's what I have in my code and I don't want to deal with regression later.
* The avx2 version has its own shuffle, we can extract it later. Probably as part of this: https://github.com/jfalcou/eve/issues/929
* We didn't agree on how we want to test the algorithms for tuples, I did smth.

### DIFF
--- a/include/eve/constant/zero.hpp
+++ b/include/eve/constant/zero.hpp
@@ -54,12 +54,13 @@ namespace eve
     template<typename T>
     EVE_FORCEINLINE T zero_(EVE_SUPPORTS(cpu_), eve::as<T> const &) noexcept
     {
-      if constexpr ( product_type<T> )
+      if constexpr ( kumi::product_type<T> )
       {
         // Can't just T{kumi::map} because that does not work for scalar product types
         T res;
         // This better inline.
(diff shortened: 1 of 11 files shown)


include/eve/constant/zero.hpp
```diff
@@ -52,9 +52,18 @@ namespace eve
   namespace detail
   {
     template<typename T>
-    EVE_FORCEINLINE auto zero_(EVE_SUPPORTS(cpu_), eve::as<T> const &) noexcept
+    EVE_FORCEINLINE T zero_(EVE_SUPPORTS(cpu_), eve::as<T> const &) noexcept
     {
-      return T(0);
+      if constexpr ( kumi::product_type<T> )
```
jfalcou, line 57: is it something we should generalize later ?

jfalcou (approved):

PR #936, opened 2021-09-10
FIX-934: inclusive scan algo

DenisYaroshevskiy:
* `inclusive_scan_inplace`/`inclusive_scan_to`
* `as_value` to convert a constant or a value to a value.


test/unit/constant/as_value.cpp
```diff
@@ -0,0 +1,60 @@
...
+    actual = eve::as_value(true, eve::as<U>{});
+    TTS_EQUAL(expected, actual);
+    actual = eve::as_value(expected, eve::as<U>{});
+    TTS_EQUAL(expected, actual);
+  }
+
+  {
+    using U = eve::element_type_t<T>;
+    U expected{0};
+    U actual = eve::as_value(eve::zero, eve::as<U>{});
```
jfalcou, line 42: I know it works, but maybe it would be good to test something else than zero?

PR #940, opened 2021-09-13
Implements soa_vector

jfalcou:
soa_vector provides storage for product types compatible with EVE and EVE algorithms.


include/eve/arch/cardinals.hpp
```diff
@@ -58,4 +58,12 @@ namespace eve
   // Template inline object for passing index within a wide
   //================================================================================================
   template<std::ptrdiff_t N> inline constexpr auto index = index_t<N>{};
+
+  //================================================================================================
+  // Hardware-agnostic cardinal
+  //================================================================================================
+  namespace detail
+  {
+    template<typename T> using common_cardinal = fixed<64 / sizeof(T)>;
```
DenisYaroshevskiy, line 67: common cardinal makes no sense as a name.
`cache_line_cardinal`?

include/eve/memory/soa_vector.hpp
```diff
@@ -0,0 +1,268 @@
...
+    soa_vector(std::size_t n, value_type value) { resize(n,value); }
+
+    //! Constructs the container with the contents of the initializer list `l`.
+    soa_vector( std::initializer_list<Type> l )
+    {
+      resize( l.size() );
+      auto ptr = l.begin();
+      for(std::size_t i = 0;i < size(); ++i)
+      {
+        kumi::for_each( [&](auto m, auto& v) { v[i] = m; }, *ptr++, storage );
```
DenisYaroshevskiy, line 70: Wait, what `operator*`? We removed all `operator*` and left only `read/write`

include/eve/memory/soa_vector.hpp
```diff
@@ -0,0 +1,268 @@
+//==================================================================================================
```
DenisYaroshevskiy, line 1: Memory is a bad place for this.
`algo -> memory`

This depends on algo.

include/eve/memory/soa_vector.hpp
```diff
@@ -0,0 +1,268 @@
...
+    public:
+    //==============================================================================================
+    //! @name Element access
+    //! @{
+    //==============================================================================================
+    //! Returns an aligned pointer to the beginning
+    EVE_FORCEINLINE auto aligned_data()  { return aligned_begin_impl(*this); }
+
+    //! Returns an aligned pointer to the beginning
+    EVE_FORCEINLINE auto aligned_data()  const { return aligned_begin_impl(*this); }
```
DenisYaroshevskiy, line 189: Stepanov advocated that the `what` should come first.
Like `stable_sort` is a bad name and it should be `sort_stable` because of an alphabetical listing.

Do you want to change to `data_aligned`? It's a suggestion, up to you

include/eve/memory/soa_vector.hpp
```diff
@@ -0,0 +1,268 @@
...
+
+    //==============================================================================================
+    //! @name Iterators
+    //! @{
+    //==============================================================================================
+    //! Returns an aligned iterator to the beginning
+    EVE_FORCEINLINE auto aligned_begin()  { return aligned_begin_impl(*this); }
+
+    //! Returns an aligned iterator to the beginning
+    EVE_FORCEINLINE auto aligned_begin()  const { return aligned_begin_impl(*this); }
```
DenisYaroshevskiy, line 221: Same here - `begin_aligned`

test/measures.hpp
```diff
@@ -38,6 +39,19 @@ namespace eve
...
+  {
+    std::ostringstream str;
+
+    str << "{ ";
+    for(std::size_t i=0;i<l.size();++i)
+      str << l.get(i) << " ";
+    str << '}';
+
+    return str.str();
+  }
```
DenisYaroshevskiy, line 53: Maybe just `operator<<`?
It's silly that `std::vector` does not have one.

test/unit/memory/soa_vector.cpp
```diff
@@ -0,0 +1,249 @@
...
+
+  // grid2d has some specific default
+  TTS_EQUAL( get<0>(udt_vector.get(0)), +1  );
+  TTS_EQUAL( get<1>(udt_vector.get(0)), -1  );
+}
+
+TTS_CASE("Check soa_vector sized ctor with default value")
+{
+  eve::soa_vector<t_t>          tuple_vector( 69, t_t{13,3.7f,'Z'} );
+  eve::soa_vector<udt::grid2d>  udt_vector  ( 69, udt::grid2d{4,20} );
```
DenisYaroshevskiy, line 46: Nice :)

Not very professional though, I feel this is not a good look for a library to do this types of things.

test/unit/memory/soa_vector.cpp
```diff
@@ -0,0 +1,249 @@
...
+  TTS_EQUAL(udt_vector.size(), 3ULL );
+  TTS_EQUAL(udt_vector.get(0), (udt::grid2d{1,2}));
+  TTS_EQUAL(udt_vector.get(1), (udt::grid2d{3,4}));
+  TTS_EQUAL(udt_vector.get(2), (udt::grid2d{5,8}));
+}
+
+TTS_CASE("Check soa_vector::clear behavior")
+{
+  eve::soa_vector<t_t>          tuple_vector(69);
+  eve::soa_vector<udt::grid2d>  udt_vector  (69);
```
DenisYaroshevskiy, line 85: Same here, let's not

include/eve/memory/soa_vector.hpp
```diff
@@ -0,0 +1,268 @@
...
+
+    //==============================================================================================
+    //! @}
+    //==============================================================================================
+
+    //==============================================================================================
+    //! @name Comparisons and ordering
+    //! @{
+    //==============================================================================================
+    friend auto operator<=>(soa_vector const& lhs, soa_vector const& rhs) = default;
```
DenisYaroshevskiy, line 252: This is poor quality.
Check sizes and then use `eve::algo::equal` for equality and `eve::algo::mismatch` for less. 
Also, I'm not sure this supports custom equality overloads, but the `algo` should.

include/eve/arch/cardinals.hpp
jfalcou: Sure thing

include/eve/memory/soa_vector.hpp
jfalcou: This is * on the `initializer_list`

include/eve/memory/soa_vector.hpp
jfalcou: We decided to have `eve/container`

include/eve/memory/soa_vector.hpp
jfalcou: I'm ok with both so le'ts do that

test/unit/memory/soa_vector.cpp
jfalcou: will change

include/eve/memory/soa_vector.hpp
jfalcou: Noted

include/eve/algo/container/soa_vector.hpp
```diff
@@ -0,0 +1,277 @@
...
+
+#include <eve/algo/zip.hpp>
+#include <eve/detail/kumi.hpp>
+#include <eve/function/read.hpp>
+#include <eve/function/write.hpp>
+#include <eve/memory/aligned_allocator.hpp>
+#include <eve/product_type.hpp>
+#include <vector>
+
+namespace eve
```
DenisYaroshevskiy, line 18: eve::algo

include/eve/algo/container/soa_vector.hpp
```diff
@@ -0,0 +1,277 @@
...
+    //! If the current size is less than `n`, additional elements are appended. Those elements are
+    //! either default-inserted or copied if `value` is provided.
+    //! @param n      new size of the container
+    //! @param value  the value to initialize the new elements with
+    void resize(std::size_t n, value_type value = {})
+    {
+      kumi::for_each( [&](auto& m, auto const& v) { m.resize(n,v); }
+                    , storage, kumi::flatten_all(value)
+                    );
+    }
```
DenisYaroshevskiy, line 145: shrink_to_fit? Maybe capacity stuff when we have a custom storage implementation.

include/eve/algo/container/soa_vector.hpp
```diff
@@ -0,0 +1,277 @@
...
+      os << "{ ";
+      for(std::size_t i=0;i<v.size();++i)
+        os << v.get(i) << " ";
+      os << '}';
+      return os;
+    }
+
+    template <typename Traits, typename Self>
+    requires std::same_as<std::remove_reference_t<Self>, soa_vector>
+    EVE_FORCEINLINE friend auto tagged_dispatch(algo::preprocess_range_, Traits tr, Self& self)
```
DenisYaroshevskiy, line 269: I don't think you tested this for `const`.

DenisYaroshevskiy: I don't think you tested any algos and fix it when I do convert here, np.
I can do that, np.

include/eve/algo/container/soa_vector.hpp
jfalcou: I added shrink_to_fit too. But here, we just want to call resize. We'll be clever as you say with the compact mono-vector implementation later on

DenisYaroshevskiy (approved):

Issue #868, opened 2021-08-22
[FEATURE] compress store for zip


DenisYaroshevskiy:
* allow any logical to be used with any wide.
* use that to implement compress store for zip.

jfalcou:
isn't first part the same as #594 ?

DenisYaroshevskiy:
@jfalcou I meant in compress_store

DenisYaroshevskiy:
Detailed explanation of what needs to be done for the first part.

### Goal

At this point the interface to compress store is:

```
  template<relative_conditional_expr C, real_scalar_value T, typename N, simd_compatible_ptr<wide<T, N>> Ptr>
  EVE_FORCEINLINE
  T* compress_store_(EVE_SUPPORTS(cpu_),
                    C c,
                    unsafe_type,  // can be safe_type too
                    wide<T, N> v,
                    logical<wide<T, N>> mask,
                    Ptr ptr) noexcept
```

And it's called as:

```
out = eve::unsafe(eve::compress_store[c])(x, m, ap_t{actual.begin()});
```

However, what to do with `bundle`?  If we just going to call the scalar version, it will be extra inefficient, since the `mask` needs to be converted for each type and all of the computations, that will be identical for all fields, have to be performed.

So - what should change? We need to accept any mask type with the same cardinal. ( `logical<wide<U, N>> ` instead of `logical<wide<T, N>>`

```
  template<relative_conditional_expr C, real_scalar_value T, real_scalar_value U, typename N, simd_compatible_ptr<wide<T, N>> Ptr>
  EVE_FORCEINLINE
  T* compress_store_(EVE_SUPPORTS(cpu_),
                    C c,
                    unsafe_type,  // can be safe_type too
                    wide<T, N> v,
                    logical<wide<U, N>> mask, 
                    Ptr ptr) noexcept
```

And then, hopefully, the compiler will see through the duplicated computations and only do them once.

### How to get there

`generic/compress_store.hpp`,  `common/compress_store_impl.hpp`
should probably change by just accepting the mask of a different type, everything else should work.
A few special case tests with different mask types will be enough.

https://github.com/jfalcou/eve/blob/c135a786dc9028d357ba6dcdd2e1f5cb8b7e4d87/include/eve/module/real/core/function/regular/generic/compress_store.hpp#L16
[...]

PR #948, opened 2021-09-14
FIX-944: attempt at erase_remove

DenisYaroshevskiy:
* extracted soa_storage. Without it I can't do iterator types, also we'll need it later.
* Implemented `erase` for one element and for a range.
* `compress_store` for bundles unfortunately does not work. https://github.com/jfalcou/eve/issues/868

### DIFF
--- a/test/unit/algo/container/soa_vector.cpp
+++ b/test/unit/algo/container/soa_vector.cpp
@@ -266,3 +266,29 @@ TTS_CASE("Check types")
   TTS_TYPE_IS(v::iterator_aligned,       (eve::algo::converting_iterator<v::pointer_aligned,       T>));
   TTS_TYPE_IS(v::const_iterator_aligned, (eve::algo::converting_iterator<v::const_pointer_aligned, T>));
 }
+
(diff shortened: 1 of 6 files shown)


test/unit/algo/container/soa_vector.cpp
```diff
@@ -247,3 +248,110 @@ TTS_CASE("Check soa_vector::begin/end behavior")
   while(uv_begin != uv_end)
     TTS_EQUAL( eve::read(uv_begin++), uv.get(i++));
 }
+
+TTS_CASE("Check types")
+{
+  using T        = udt::grid2d;
+  using ap       = eve::aligned_ptr<int,       eve::fixed<64 / sizeof(int)>>;
```
jfalcou, line 255: why not using the `cache_line_cardinal` thing ?

DenisYaroshevskiy: It's in detail. I figured testing using detail is not ideal

jfalcou (approved):

PR #958, opened 2021-09-17
Make operator overloading for custom types possible

jfalcou:


include/eve/arch/cpu/wide.hpp
```diff
@@ -346,10 +346,10 @@ namespace eve
     void swap( wide& other ) { std::swap(this->storage(), other.storage()); }
 
     //! Pre-incrementation operator
-    EVE_FORCEINLINE wide &operator++()    noexcept  { return detail::self_add(*this, wide{1}); }
+    EVE_FORCEINLINE wide &operator++()    noexcept  { return *this += wide{1}; }
 
     //! Pre-decrementation operator
-    EVE_FORCEINLINE wide &operator--()    noexcept  { return detail::self_sub(*this, wide{1}); }
+    EVE_FORCEINLINE wide &operator--()    noexcept  { return *this -= wide{1}; }
```
DenisYaroshevskiy, line 352: Well. Hmm. I think that for floats operator++ goes to the next float.
Not sure. If you use operator++ on floats it's on you anyways)

jfalcou: ++ just adds 1

PR #959, opened 2021-09-17
Fix #185 - Implements swap_if

### LINKED ISSUE #185: conditional swap
a  `swap[cond](a,b)` would be great.
It seems that could be a shuffle

jfalcou:


DenisYaroshevskiy (approved):
Should it work for scalar?

PR #962, opened 2021-09-17
FIX-951: complex numbers examples.

DenisYaroshevskiy:


examples/oop/complex_numbers__declaring_an_object_type.cpp
```diff
@@ -0,0 +1,223 @@
...
+  EVE_FORCEINLINE friend auto& operator+=(eve::same_value_type<cmplx> auto& self, eve::same_value_type<cmplx> auto other)
+  {
+    re(self) += re(other);
+    im(self) += im(other);
+    return self;
+  }
+
+  // eve will always use the += but if you want the operation to be provided for scalars
+  // you should provide this.
+  EVE_FORCEINLINE friend auto operator+(eve::same_value_type<cmplx> auto self, eve::same_value_type<cmplx> auto other)
```
jfalcou, line 78: that should be

` EVE_FORCEINLINE friend auto operator+(eve::same_value_type<cmplx> auto self, eve::same_value_type<float> auto other)`

or else we just do the wrong computation

DenisYaroshevskiy: Discussed, it's a different +. fixed.

PR #963, opened 2021-09-18
FIX-888: `like` concept and a CRTP base to ease creating of structs.

DenisYaroshevskiy:
`same_value_type` accepted way too much, it was not comfortable to talk about that in the talk.
Created `eve::like` (as we discussed) that requires an opt in.

Also created a `product_type_base` crtp helper, because explaining where the `operator+` will be generated for you and where it won't was not comfortable.
And it reduces the example size.

### DIFF
--- a/include/eve/product_type.hpp
+++ b/include/eve/product_type.hpp
@@ -69,4 +69,122 @@ namespace eve
 
   template<typename Type>
   inline constexpr bool supports_ordering_v = supports_ordering<Type>::value;
+
+   //================================================================================================
+  //! @addtogroup struct
+  //! @{
+  //!   @struct supports_like
+  //!   @concept like
+  //!
+  //!   `like<Wrapper, T>` is a concept that describes a wrapper for which most of the functionality
+  //!   for T should be applicable. T is always like T.
+  //!
+  //!   `like< Wrapper, T>` is true if Wrapper defines a typedef `using is_like = T` or
+  //!   specialises `eve::supports_like`.
+  //!
+  //!   Example: `eve::wide<T>` is `like<T>`.
+  //!
+  //!   This is used to work with custom structs in eve, see `oop` section in `examples`.
+  //! @}
+  //================================================================================================
+
+
+  template<typename Wrapper, typename Self> struct supports_like : std::false_type {};
+
+  template <typename Wrapper, typename Self>
+  requires requires(Wrapper) { typename Wrapper::is_like; }
+  struct supports_like<Wrapper, Self> :
+  std::bool_constant<std::same_as<Self, typename std::remove_cvref_t<Wrapper>::is_like>>
+  {};
+
+  template<typename Wrapper, typename Self>
+  concept like = std::same_as<std::remove_cvref_t<Wrapper>, Self> ||
+                 supports_like<std::remove_cvref_t<Wrapper>, Self>::value;
+
+  //================================================================================================
+  //! @addtogroup struct
+  //! @{
+  //!   @struct product_type_base<Self, ...Fields>
+  //!   @brief  a helper CRTP base to declare user defined types easier
+  //!
+  //!   It will generate a member wise equality.
+  //!   It will also generate a lexicographical order for your type, unless !supports_ordering_v<Self>
+  //!
+  //!   It generated the not mutating operator version from a mutating one (like operator+ from operator+=),
+  //!   however does not go further than that (like operator-= b from operator+= and operator-)
+  //!   since simd has an instruction for `-` and you want `operator-=` to be declared separately.
+  //!
+  //! @}
+  template <typename Self, typename ... Fields>
+  struct product_type_base;
+
+  namespace detail
+  {
+    std::false_type derived_from_product_type_base_impl(...);
+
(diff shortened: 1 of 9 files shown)


include/eve/product_type.hpp
```diff
@@ -69,4 +69,122 @@ namespace eve
...
+  //!   It will generate a member wise equality.
+  //!   It will also generate a lexicographical order for your type, unless !supports_ordering_v<Self>
+  //!
+  //!   It generated the not mutating operator version from a mutating one (like operator+ from operator+=),
+  //!   however does not go further than that (like operator-= b from operator+= and operator-)
+  //!   since simd has an instruction for `-` and you want `operator-=` to be declared separately.
+  //!
+  //! @}
+  template <typename Self, typename ... Fields>
+  struct product_type_base;
```
jfalcou, line 119: I guess the `Fields...` is because we're are building a type with shortcuts and not adapting an existing type ?

include/eve/product_type.hpp
```diff
@@ -69,4 +69,122 @@ namespace eve
...
+    std::true_type derived_from_product_type_base_impl(product_type_base<Self, Fields...> const&);
+
+    template <typename Type>
+    concept derived_from_product_type_base = decltype(derived_from_product_type_base_impl(Type{}))::value;
+  }
+
+  template <typename Self, typename ... Fields>
+  struct product_type_base : kumi::tuple<Fields...>
+  {
+    using tuple = kumi::tuple<Fields...>;
```
jfalcou, line 135: `tuple_type` ?

include/eve/product_type.hpp
```diff
@@ -69,4 +69,122 @@ namespace eve
...
+      return x;
+    }
+
+    EVE_FORCEINLINE friend auto operator>>(eve::like<Self> auto x, integral_value auto s) requires requires { x >>= s; }
+    {
+      x >>= s;
+      return x;
+    }
+
+    friend auto operator< (eve::like<Self> auto x, eve::like<Self> auto y) requires (!supports_ordering_v<Self>) = delete;
```
jfalcou, line 179: Why do we need this ? IS it for hiding the ordering we inherit from kumi::tuple ?

test/unit/api/udt/product_type.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+template <typename T>
+struct wrapper
+{
+  using is_like = T;
+};
+
+template<> struct eve::supports_like<wrapper<int>, std::int64_t> : std::true_type {};
+
+TTS_CASE("like concet")
```
jfalcou, line 20: concept ?

test/unit/api/udt/product_type.cpp
```diff
@@ -0,0 +1,162 @@
...
+  { operators_test       auto _ = eve::wide<supports_all_ops>{}; (void)_; }
+  { eve_totally_ordered  auto _ = eve::wide<supports_all_ops>{}; (void)_; }
+  TTS_PASS("all types check out");
+
+  { eve::product_type auto _ = supports_no_ops{}; (void)_; }
+  { std::semiregular  auto _ = supports_no_ops{}; (void)_; }
+
+  TTS_CONSTEXPR_EXPECT_NOT(supports_plus_test<supports_no_ops>);
+  TTS_CONSTEXPR_EXPECT_NOT(std::totally_ordered<supports_no_ops>);
+  // TTS_CONSTEXPR_EXPECT_NOT(supports_plus_test<eve::wide<supports_no_ops>>);
```
jfalcou, line 153: why commented?

include/eve/product_type.hpp
jfalcou: Not a big fan of base in the name but this is basically bikeshedding

include/eve/product_type.hpp
jfalcou: Miss the bitwise one

include/eve/product_type.hpp
DenisYaroshevskiy: I'm open to suggestions for names.
...Fields is to specify the types of fields, I don't get your comment tbh.

include/eve/product_type.hpp
DenisYaroshevskiy: Will remove, I don't need it.

include/eve/product_type.hpp
DenisYaroshevskiy: > Why do we need this ? IS it for hiding the ordering we inherit from kumi::tuple ?

Yes. We want to use ordering from tuple. Unless oreding is disabled.

Will do the bitwise ones.

test/unit/api/udt/product_type.cpp
DenisYaroshevskiy: Will do.

test/unit/api/udt/product_type.cpp
DenisYaroshevskiy: Wide operators are not sfinae friendly.
I don't know if we care or not.
For product type base we probably do, since we want the effect of you not declaring them.

jfalcou (approved):

PR #973, opened 2021-09-22
Reinstate aggregate tests

### LINKED ISSUE #511: [FEATURE] avx512 aggregated tests
We have bugs with avx512 aggregations.
I guess we should add a VM for these tests. Should not be too expensive.

jfalcou:
In order to be able to apply small-scale changes to actually fix #511 and related issues, we put the aggregate test back up on AVX512 and locally snuff out the problematic tests by using a secondary list of types selector.

Those changes are meant to be temporary as multiple fixes for different actual issues are being pushed.

### DIFF
--- a/include/eve/module/real/core/function/regular/generic/load.hpp
+++ b/include/eve/module/real/core/function/regular/generic/load.hpp
@@ -125,7 +125,7 @@ namespace eve::detail
                                 , eve::as<Pack> const& tgt, Ptr ptr
                                 ) noexcept
   {
-    if constexpr(sanitizers_are_on && !kumi::product_type<Ptr>)
+    if constexpr(sanitizers_are_on)
     {
       Pack that;
 
--- a/include/eve/module/real/core/function/regular/generic/load.hpp
+++ b/include/eve/module/real/core/function/regular/generic/load.hpp
@@ -125,7 +125,7 @@ namespace eve::detail
                                 , eve::as<Pack> const& tgt, Ptr ptr
                                 ) noexcept
   {
-    if constexpr(sanitizers_are_on)
+    if constexpr(sanitizers_are_on && !kumi::product_type<Ptr>)
     {
       Pack that;
 
--- a/test/unit/module/real/core/rem.cpp
+++ b/test/unit/module/real/core/rem.cpp
@@ -136,9 +136,9 @@ EVE_TEST( "Check behavior of rem on signed types"
   using eve::is_nez;
   using eve::pedantic;
   using eve::detail::map;
-  TTS_ULP_EQUAL( pedantic(rem[is_nez(a2)])(a0, a2), map([](auto e, auto f) {return is_nez(f) ? rem(e, f) : e ; }, a0, a2), 24);
+  TTS_ULP_EQUAL( pedantic(rem[is_nez(a2)])(a0, a2), map([](auto e, auto f) {return is_nez(f) ? rem(e, f) : e ; }, a0, a2), 48);
   a2 = eve::if_else(a2 >= 0, eve::one, a2);
-  TTS_ULP_EQUAL( rem[is_nez(a2)](a0, a2), map([](auto e, auto f) {return rem(e, f); }, a0, a2), 24);
+  TTS_ULP_EQUAL( rem[is_nez(a2)](a0, a2), map([](auto e, auto f) {return rem(e, f); }, a0, a2), 48);
 
   TTS_ULP_EQUAL( rem[a2 > T(64)](a0, a1), map([](auto e, auto f, auto g) {return g > 64 ? rem(e, f) : e ; }, a0, a1, a2), 2);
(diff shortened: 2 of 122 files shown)


include/eve/module/real/core/function/regular/generic/load.hpp
```diff
@@ -125,7 +125,7 @@ namespace eve::detail
                                 , eve::as<Pack> const& tgt, Ptr ptr
                                 ) noexcept
   {
-    if constexpr(sanitizers_are_on)
+    if constexpr(sanitizers_are_on && !kumi::product_type<Ptr>)
```
DenisYaroshevskiy, line 128: Why is this change?

test/unit/module/real/core/rem.cpp
```diff
@@ -73,9 +73,9 @@ EVE_TEST( "Check behavior of rem on wide"
   using eve::rem;
   using eve::pedantic;
   using eve::detail::map;
-  TTS_ULP_EQUAL( pedantic(rem)(a0, a1), map([](auto e, auto f) { return eve::rem(e, f); }, a0, a1), 30);//fma not avail scalar double it seems
+  TTS_ULP_EQUAL( pedantic(rem)(a0, a1), map([](auto e, auto f) { return eve::rem(e, f); }, a0, a1), 32);//fma not avail scalar double it seems
```
DenisYaroshevskiy, line 76: Why is this changing?

test/unit/module/real/core/remdiv.cpp
```diff
@@ -50,8 +50,8 @@ EVE_TEST( "Check behavior of remdiv on wide"
   using eve::toward_zero;
   a1 = eve::if_else(eve::is_eqz(a1), eve::one, a1);
   auto [r, d] = eve::remdiv(a0, a1);
-  TTS_ULP_EQUAL( r, map([](auto e, auto f) { return eve::rem(e, f); }, a0, a1), 30);//fma not avail scalar double it seems
-  TTS_ULP_EQUAL( d, map([](auto e, auto f) { return toward_zero(eve::div)(e, f); }, a0, a1), 30);//fma not avail scalar double it seems
+  TTS_ULP_EQUAL( r, map([](auto e, auto f) { return eve::rem(e, f); }, a0, a1), 32);//fma not avail scalar double it seems
+  TTS_ULP_EQUAL( d, map([](auto e, auto f) { return toward_zero(eve::div)(e, f); }, a0, a1), 32);//fma not avail scalar double it seems
```
DenisYaroshevskiy, line 54: And these?

test/unit/module/real/elliptic/ellint_1.cpp
```diff
@@ -44,6 +44,6 @@ EVE_TEST( "Check behavior of ellint_1 on wide"
   using eve::detail::map;
   using v_t = eve::element_type_t<T>;
 
-  TTS_ULP_EQUAL(eve::ellint_1(k)      , map([](auto e) -> v_t { return boost::math::ellint_1(e); }, k), 10);
-  TTS_ULP_EQUAL(eve::ellint_1(phi, k) , map([](auto e, auto f) -> v_t { return boost::math::ellint_1(e, f); }, k, phi), 10);
+  TTS_ULP_EQUAL(eve::ellint_1(k)      , map([](auto e) -> v_t { return boost::math::ellint_1(e); }, k), 16);
+  TTS_ULP_EQUAL(eve::ellint_1(phi, k) , map([](auto e, auto f) -> v_t { return boost::math::ellint_1(e, f); }, k, phi), 16);
```
DenisYaroshevskiy, line 48: Same for these ones?

DenisYaroshevskiy (commented):
So most of this change is introducing smaller set of types but building everything for 512, right?

What are the other changes?
Like load change and the math tests

test/unit/module/real/core/remdiv.cpp
jfalcou: as the size of the vector changed to includes the new aggregate size for avx512, it appears that one value actually trigger a slightly larger ULP.

test/unit/module/real/elliptic/ellint_1.cpp
jfalcou: as the size of the vector changed to includes the new aggregate size for avx512, it appears that one value actually trigger a slightly larger ULP.

test/unit/module/real/core/rem.cpp
jfalcou: as the size of the vector changed to includes the new aggregate size for avx512, it appears that one value actually trigger a slightly larger ULP.

include/eve/module/real/core/function/regular/generic/load.hpp
jfalcou: I have to check cause I thought I had it removed

DenisYaroshevskiy (approved):

PR #977, opened 2021-09-26
FIX-938: consider types

DenisYaroshevskiy:
Now an iterator can use more than one type to compute cardinal.
This allows to do conversions to a smaller type: `converting_iterator<int*, short>` will have cardinal of int.

We can also pass extra types to consider in a trait.

This:
*  fixes `reduce/inclusive_scan` incorrect type selection issue.
* simplifies `collect_indexes` example.
* opens the road to transform.

### DIFF
--- a/include/eve/algo/concepts/types_to_consider.hpp
+++ b/include/eve/algo/concepts/types_to_consider.hpp
@@ -1,0 +1,31 @@
+//==================================================================================================
+/*
+  EVE - Expressive Vector Engine
+  Copyright : EVE Contributors & Maintainers
+  SPDX-License-Identifier: MIT
+*/
+//==================================================================================================
+#pragma once
+
+#include <eve/algo/concepts/value_type.hpp>
+
+#include <eve/detail/kumi.hpp>
+
+namespace eve::algo
+{
+  template <typename RorI>
+  struct types_to_consider_for
+  {
+    using type = kumi::tuple<value_type_t<RorI>>;
+  };
+
+  template <typename RorI>
+  requires requires(RorI) { typename RorI::types_to_consider; }
+  struct types_to_consider_for<RorI>
+  {
+    using type = typename RorI::types_to_consider;
+  };
+
+  template <typename RorI>
+  using types_to_consider_for_t = typename types_to_consider_for<RorI>::type;
+}
--- a/include/eve/algo/zip.hpp
+++ b/include/eve/algo/zip.hpp
@@ -8,9 +8,10 @@
 #pragma once
 
 #include <eve/algo/as_range.hpp>
-#include <eve/algo/range_ref.hpp>
 #include <eve/algo/concepts/relaxed.hpp>
+#include <eve/algo/concepts/types_to_consider.hpp>
 #include <eve/algo/preprocess_range.hpp>
+#include <eve/algo/range_ref.hpp>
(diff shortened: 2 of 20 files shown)


include/eve/detail/kumi.hpp
```diff
@@ -799,12 +799,12 @@ namespace kumi
 
   namespace result
   {
-    template<product_type T, product_type... Tuples> struct cat
+    template<product_type... Tuples> struct cat
```
jfalcou, line 802: why this change?

include/eve/algo/zip.hpp
```diff
@@ -8,9 +8,10 @@
 #pragma once
 
 #include <eve/algo/as_range.hpp>
-#include <eve/algo/range_ref.hpp>
 #include <eve/algo/concepts/relaxed.hpp>
+#include <eve/algo/concepts/types_to_consider.hpp>
 #include <eve/algo/preprocess_range.hpp>
+#include <eve/algo/range_ref.hpp>
```
jfalcou, line 14: why this include ?

jfalcou (commented):
I see the thing but I fail to grasp where exactly the types to consider get acted upon. 
Could you explain the process  shortly ?

include/eve/detail/kumi.hpp
DenisYaroshevskiy: Because otherwise unpacking didb't work.  I think it's a clang bug.

include/eve/algo/concepts/types_to_consider.hpp
```diff
@@ -0,0 +1,31 @@
+//==================================================================================================
+/*
+  EVE - Expressive Vector Engine
+  Copyright : EVE Contributors & Maintainers
+  SPDX-License-Identifier: MIT
+*/
+//==================================================================================================
+#pragma once
+
+#include <eve/algo/concepts/value_type.hpp>
```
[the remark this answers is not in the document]
DenisYaroshevskiy: We don't have any.

include/eve/algo/zip.hpp
DenisYaroshevskiy: It was here, I just sorted them.

include/eve/algo/concepts/types_to_consider.hpp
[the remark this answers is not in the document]
DenisYaroshevskiy: We also call `settings` `traits`. So at the moment calling this a trait will be confusing.

Perhaps `traits` should become `settings` or `options` and this should be some `meta` or `type_functions` thing. But w/e at the moment.

DenisYaroshevskiy (commented):
> I see the thing but I fail to grasp where exactly the types to consider get acted upon.
Could you explain the process shortly ?

There are two ways to get extra types to consider:
* iterator/range can have some extra types (like converting).
* user passed trait: `consider_types`

The preprocess range (in it's different forms like zip/non-zip) looks at them and decides what cardinal should be.

The iterator/range access one is mostly now necessary for `zip`:

```
eve::algo::traits types_to_consider{ consider_types_key = kumi::result::cat_t<types_to_consider_for_t<Rngs> ...>{} };
```

The main `preprocess_range` gets the cardinal from `iteration_cardinal_t` which takes into account all of the ways extra types can be specified.

jfalcou (approved):

PR #978, opened 2021-09-26
FIX-947:compress 4 elements, sse2

DenisYaroshevskiy:
newplot (25)

We at the very least beat the scalar version.
It' a bit unclear wether the mask shuffle or the switch is better for 4 elements.

Especially on smaller sizes.

However, we know that the switch version is much more susceptible to code alignment and to data.

SSE2:
newplot (27)

SSE4.2:
newplot (28)

We can remove the shuffle mask 4 element version if we decided so later.
We beat the scalar version consistently, that's what's important.

We also improved chars/shorts by accident:

Chars
newplot (29)

Shorts:
newplot (30)

I did not look at them very much, we can probably do substantially better.


include/eve/detail/function/simd/x86/compress_store_impl.hpp
```diff
@@ -14,6 +14,74 @@
...
+  auto compress_store_impl_switch_4(C c,
+                                    wide<T, fixed<4>> v,
+                                    logical<wide<U, fixed<4>>> mask
+                                    )
+  {
+    auto [num, last_set] = compress_store_swizzle_mask_num[c](mask);
+    int count;
+
+    switch (num) {
+      case 0: { count = 0; v = v[pattern<3, 0, 0, 0>]; break; } // 000
```
jfalcou, line 34: can we actually merge both the case and the comment by doing `case 0b001` and so on or is it risky ?

jfalcou (approved):

PR #990, opened 2021-09-29
Fix 974 - slide_left/right fix

jfalcou:


include/eve/detail/function/simd/common/slide_right.hpp
```diff
@@ -85,28 +106,18 @@ namespace eve::detail
...
+      if constexpr( logical_value<Wide> && Wide::abi_type::is_wide_logical)
+      {
+        return bit_cast(slide_right(x.bits(), y.bits(), index<Shift>), as<Wide>{});
+      }
+      else
+      {
+        Wide res{ [&](int i, int size) {
+          if (i < Shift) return x.get(size - Shift + i);
+          else           return y.get(i - Shift);
+        }};
```
DenisYaroshevskiy, line 118: The slide_left call basic swizzle.
This one should too?

PR #992, opened 2021-09-29
Fix #983  - Add eve::construct<product_type>(wide...) to construct wide of UDTs

### LINKED ISSUE #983: [FEATURE] Missing `make_tuple` like utility
Motivational problem: "create vertical lines"

We have starts, and distances, we want to produce lines

```
eve::algo::soa_vector<udt::point2D> starts;
std::vector<std::int32_t> lengths;

eve::algo::soa_vector<udt::line2D> actual(starts.size());

  eve::algo::transform_to(eve::algo::zip(starts, lengths), actual,
    [](auto start_l) {
      auto [start, l] = start_l;
      auto end = start;
      get_y(end) -= l;
      return eve::wide<udt::line2D>{start, end};  // this is the problem
  });
```

As you can see I had to create a `eve::wide<udt::line2D>`. However if there was some type conversion of the cardinal didn't match the default one, I'd have to do smth more tricky.

What I actually need is:

```
auto some_function(wide<T1, N>, wide<T2, N>, ... wide<Tn, N>) -> wide<tuple<T1, ... Tn>>
```

`make_tuple` sounds a bit ify.
`zip` we already have in algo, I'm not sure how much it makes sense.
Something is needed.

jfalcou:


include/eve/function/zip.hpp
```diff
@@ -0,0 +1,73 @@
...
+  requires (sizeof...(Ws) == std::tuple_size<Target>::value)
+  {
+    return wide<Target>{ws...};
+  }
+
+  template<simd_value... Ws>
+  EVE_FORCEINLINE auto zip_(EVE_SUPPORTS(cpu_),  Ws... ws) noexcept
+  {
+    return wide<kumi::tuple<eve::element_type_t<Ws>...>>{ws...};
+  }
```
DenisYaroshevskiy, line 72: This is definetely wrong. It does not compute N.

jfalcou: Forgot to reuse the wide constructor requires here.

DenisYaroshevskiy: You miss the cardinal in the result

DenisYaroshevskiy (approved):

PR #994, opened 2021-09-30
Start here examples

DenisYaroshevskiy:
I think that the first examples that the person sees should be something like these ones.

@jtlap  - can you please check that my math code makes sense?


jfalcou (commented):
I would like the start_here.cpp to be sliced into separated files, one per tests. Otherwise, it's OK.

PR #996, opened 2021-10-01
Refurbish ECS test into DD test using physics

jfalcou:


examples/oop/data_driven_physics.cpp
```diff
@@ -0,0 +1,149 @@
...
+
+    // Update velocity with gravity
+    speed(falling)    -= g*dt;
+
+    // Compute the bounced ball case
+    speed(bounced)    *= -elasticity(bounced);
+    count(bounced)++;
+
+    // Select the proper one based on position
+    return eve::if_else(  count(b) < max_bounce
```
DenisYaroshevskiy, line 72: I don't get this `count` and `max_bounce`.
The speed goes down, right?
So, maybe, just let it go down? and trim below a treshold?

jfalcou: Whenever the position reaches 0, we bounce by reverting the speed vector sign. It'll then get affected by gravity again, then rebound, etc... It will bounce until the speed is close to 0, at which point we stick the particle to the ground.

the count is used to count how many bounces happened for a particule and we stop after a max number of bounces.

DenisYaroshevskiy: Why do we stop after max bounce? It should stop itself, right?
I mean, you can say if the 'speed < treshold`, speed = 0. That's it.

DenisYaroshevskiy (approved):
I don't care that much about that if. You can merge. Shame we don't have a `remove` used

DenisYaroshevskiy (approved):

PR #998, opened 2021-10-02
FIX-989: compress store arm.

DenisYaroshevskiy:
Tagging @aqrit - port of his algorithm to arm.

### extracted `compress_store_swizzle_mask_num_partial`/`compress_store_swizzle_mask_num` for 4 elements.

On x86 we mostly use `compress_store_swizzle_mask_num_partial`: it does not compute the count, just the bool if the last element is set or not (except for 4 doubles, since there is a shuffle of ints which does not allow to store popcount in the mask).

The proper `compress_store_swizzle_mask_num` returns the mask number and the popcount.
On arm we use a sum with mask to get it.

### compress store for 4/8 elements.

We still don't have arm benches but here is the asm.

remove ints, arm-v8

```
.L6:
	ldr	q1, [x2], 16
	cmeq	v0.4s, v1.4s, #0
	bic	v0.16b, v2.16b, v0.16b
	addv	s0, v0.4s
	umov	w0, v0.s[0]
	ubfiz	x3, x0, 4, 4
	asr	w0, w0, 4
	ldr	q0, [x23, x3]
#APP
// 11270 "/usr/lib/gcc-cross/aarch64-linux-gnu/10/include/arm_neon.h" 1
	tbl v1.16b, {v1.16b}, v0.16b
// 0 "" 2
#NO_APP
	str	q1, [x19]
	add	x19, x19, x0, sxtw 2
	cmp	x20, x2
	bne	.L6
```

remove shorts, arm-v8

```
.L6:
	ldr	q2, [x2], 16
	cmeq	v1.8h, v2.8h, #0
	ext	v3.16b, v2.16b, v4.16b, #2
	not	v0.16b, v1.16b
	and	v1.16b, v0.16b, v5.16b
	bsl	v0.16b, v2.16b, v3.16b
	addv	h1, v1.8h
	umov	w0, v1.h[0]
	ubfiz	x3, x0, 4, 5
	ubfx	x0, x0, 7, 9
	ldr	q1, [x23, x3]
#APP
// 11270 "/usr/lib/gcc-cross/aarch64-linux-gnu/10/include/arm_neon.h" 1
	tbl v0.16b, {v0.16b}, v1.16b
// 0 "" 2
#NO_APP
	str	q0, [x19]
	add	x19, x19, x0, lsl 1
	cmp	x20, x2
	bne	.L6
```

Chars have some volume, but seem OK too.

### DIFF
--- a/include/eve/detail/function/simd/common/compress_store_impl.hpp
+++ b/include/eve/detail/function/simd/common/compress_store_impl.hpp
@@ -37,6 +37,40 @@ namespace eve::detail
     };
   }
 
+  // See compress_store_num for explanation
+  template <std::unsigned_integral T>
+  EVE_FORCEINLINE constexpr auto pattern_8_elements(std::array<T, 8> idxs)
+  {
+    using row = std::array<T, 8>;
+
+    std::array<row, 27> res = {};
+
+    for (unsigned i = 0; i != 27; ++i)
+    {
+      unsigned number_of_9s = 0, number_of_3s = 0, number_of_1s = 0;
+      unsigned base_3_value = i;
+
+      if (base_3_value >= 9) ++number_of_9s, base_3_value -= 9;
+      if (base_3_value >= 9) ++number_of_9s, base_3_value -= 9;
+      if (base_3_value >= 3) ++number_of_3s, base_3_value -= 3;
+      if (base_3_value >= 3) ++number_of_3s, base_3_value -= 3;
+      if (base_3_value >= 1) ++number_of_1s, base_3_value -= 1;
+      if (base_3_value >= 1) ++number_of_1s, base_3_value -= 1;
+
+      auto* it = res[i].begin();
+      if (number_of_1s) *it++ = idxs[0], --number_of_1s;
+      if (number_of_1s) *it++ = idxs[1], --number_of_1s;
+      if (number_of_3s) *it++ = idxs[2], --number_of_3s;
+      if (number_of_3s) *it++ = idxs[3], --number_of_3s;
+      if (number_of_9s) *it++ = idxs[4], --number_of_9s;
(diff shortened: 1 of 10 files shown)


include/eve/detail/function/simd/common/compress_store_impl.hpp
```diff
@@ -18,6 +18,94 @@
 
 namespace eve::detail
 {
+  // permutation masks ------------------
```
jfalcou, line 21: shoudln't all of this actually be `consteval`to be sure we don't do any og this at runtime ?

include/eve/module/real/core/function/regular/simd/arm/neon/convert.hpp
```diff
@@ -163,14 +163,15 @@ namespace eve::detail
...
            if constexpr( catou == category::int16x8 ) return vmovl_s8(v0);
+      else if constexpr( catou == category::int16x4 ) return vget_low_s16(vmovl_s8(v0));
       else                                            return convert_(EVE_RETARGET(simd_), v0, tgt);
     }
     //==============================================================================================
     // uint8 -> ?
     else if constexpr( catin == category::uint8x8 )
     {
            if constexpr( catou == category::uint16x8 )  return vmovl_u8(v0);
-      else if constexpr( catou == category::int8x8 )    return vreinterpret_s8_u8(v0);
```
jfalcou, line 173: why is this one convert removed ? It was correct.

DenisYaroshevskiy: Done before in a generic way:

https://github.com/jfalcou/eve/blob/6f8ecb1e2b17c7f4301ba37c5e485f48014bf959/include/eve/module/real/core/function/regular/simd/arm/neon/convert.hpp#L29-L40

jfalcou (approved):

PR #1002, opened 2021-10-04
Improve/trigo

jtlap:


jtlap:
decorators names have still not being modified

Issue #984, opened 2021-09-27
[FEATURE] ECS complicated example


DenisYaroshevskiy:
Started on this but I don't like it.
We need something simpler. And finished.

jfalcou:
Fixed by #1004

PR #1004, opened 2021-10-05
Add remove_if to the Data Driven example

jfalcou:
FInal fix for #984 that put remove_if back.

### DIFF
--- a/examples/oop/data_driven_physics.cpp
+++ b/examples/oop/data_driven_physics.cpp
@@ -67,13 +67,10 @@ struct update
 
     // Compute the bounced ball case
     speed(bounced)    *= -elasticity(bounced);
-    count(bounced)++;
+    ++count(bounced);
 
     // Select the proper one based on position
-    return eve::if_else(  count(b) < max_bounce
-                        , eve::if_else( position(falling) <= 1e-4, bounced, falling)
-                        , b
-                        );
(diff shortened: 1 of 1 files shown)


examples/oop/data_driven_physics.cpp
```diff
@@ -137,10 +138,13 @@ TTS_CASE("Run a ball simulation")
       times.clear();
     }
 
+    // Remove ball that bounced
+    balls.erase(eve::algo::remove_if(balls, [max_bounce](auto b) { return count(b) < max_bounce; }) );
```
DenisYaroshevskiy, line 142: Should be count >= no?

jfalcou: WIll fix

DenisYaroshevskiy (approved):

examples/oop/data_driven_physics.cpp
```diff
@@ -120,10 +121,13 @@ TTS_CASE("Run a ball simulation")
 
     // Display 200 balls
     for(std::size_t i=0;i<200;i++)
```
DenisYaroshevskiy, line 123: just up to the ball size would be simpler but w/e

jfalcou: if you put 10000 balls , it doesn't fit the screen. I just plot the 200 first

Issue #1023, opened 2021-10-15
Adding git tags, and github releases


the-moisrex:
Since I'm (at least trying to) use [CPM.cmake](https://github.com/cpm-cmake/CPM.cmake) in my project and this great project still doesn't have a tag or a github release, I'm afraid it would require a little bit of getting hacky; I can I ask for releases or a git tag with a name like "v1.20.5"?

Not specifying a version doesn't really sets well with CPM;
```cmake
CPMAddPackage(
        NAME eve
        GITHUB_REPOSITORY jfalcou/eve
)
```
The error if you try:
```
CMake Error at eve-subbuild/eve-populate-prefix/tmp/eve-populate-gitclone.cmake:40 (message):
  Failed to checkout tag: 'v'
```

Thanks

jfalcou:
We can do that. We currently had no release as we're still breaking things.
@DenisYaroshevskiy  should we get a a v0.0.1 or something for this ?

DenisYaroshevskiy:
> We can do that. We currently had no release as we're still breaking things.
> @DenisYaroshevskiy should we get a a v0.0.1 or something for this ?

Up to you - whatever needs to be done for deployment. As long as we can reasonably live at head and break ppl's code on a regular basis )

the-moisrex:
I do have [some benchmarks](https://github.com/the-moisrex/webpp/blob/dev/benchmarks/ieuqals/iequals_benchmark.cpp) on this that I needs to re-check as well which requires me to make my project compile-able again 😂; but overall, not having to deal with alignment and manual for loops is great. I'm looking forward to not having to deal with is_less and if_else functions as well if all possible!

the-moisrex:
It's really great; but it's missing 2 optimization features:

1. checking the size (they're not equal if they don't have the same size)
2. the user can provide the information about whether or not the A or B are all in small caps or all capital letters.

And also the to_lower and to_upper could be their own utilities (which I made them [here](https://github.com/the-moisrex/webpp/blob/dev/core/include/webpp/strings/to_case.hpp), with some [benchmark](https://github.com/the-moisrex/webpp/blob/dev/benchmarks/tolower/tolower_benchmark.cpp)). The same goes for is_lower and is_upper.

And all of these are only about ascii and doesn't work for any other locale actually. So far away from complete.

But this is what some of us need; specially in places like parsing HTTP headers or dealing with protocols that only accept ASCII.

[...]

PR #1025, opened 2021-10-16
case-insensitive equal example added

the-moisrex:
After discussing about versioning this library in #1023, we somehow end up [here](https://github.com/jfalcou/eve/issues/1023#issuecomment-944787153), adding a new example for case-insensitive equal with SIMD.

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -7,6 +7,7 @@ make_unit( "examples" start_here.cpp )
 
 make_unit( "examples" algorithms/using_existing/memcmp__two_range_algorithms_interface_specifics.cpp )
 make_unit( "examples" algorithms/using_existing/inclusive_scan_zip__using_zip_with_algorithms.cpp )
+make_unit( "examples" algorithms/using_existing/case_insensitive_equals.cpp )
 
 find_package(Threads REQUIRED)
 make_unit( "examples" algorithms/using_existing/inclusive_scan_par_unseq__using_eve_to_build_both_vectorized_and_parallel_algos.cpp )


DenisYaroshevskiy:
Can you please also move it into: `examples/algorithms/using_existing`?

the-moisrex:
Shouldn't I add it in a CMakeLists.txt somewhere?

jfalcou:
> btw why are you using tests for examples? it's confusing a bit!
It is to make sure we don't change some API or introduce a bug somewhere and we let examples rot only for people to find out the hard way we've been slacking on updating the examples.

the-moisrex:
> I figured I'll do it for you, together with some more testing. Otherwise you'd have to figure out how to build `eve`

Thanks; I did figure out that it's in `examples/CMakeLists.txt` file;
Building it is I guess is like this: `cmake -DCMAKE_BUILD_TYPE=Debug -G "Ninja" -DEVE_BUILD_TEST=ON -DEVE_BUILD_BENCHMARKS=ON -DEVE_BUILD_RANDOM=ON ..`

and then `ninja unit`; but my lack of space on my storage got to me 😅 
```
/tmp/cc1Ca1up.s: Error fatal: test/random/module/real/core/CMakeFiles/random.real.core.is_nlez.regular.simd.exe.dir/__/__/__/__/__/tmp-src/random.real.core.is_nlez/regular.simd.cpp.o: No queda espacio en el dispositivo
```

See if it's in the right place!

DenisYaroshevskiy:
If you insist, here is the doc: https://jfalcou.github.io/eve/html/dev_cmake.html

DenisYaroshevskiy:
Ah, wait - that assumes docker set up

I do like this:

```
cmake .. -G Ninja -DCMAKE_CXX_FLAGS='-march=native --stdlib=libc++ -DEVE_NO_FORCEINLINE'
inja -j 16 examples.exe  && unit/examples.oop.data_driven_physics.exe

PR #1049, opened 2021-11-01
convert: improve u64 => u32 when using AVX2

aguinet:
Preliminary benchmarks (using
https://github.com/aguinet/eve/commit/38599cf6ad77d1bd6dcbb99970ed2332a97a5eaf):

Before patch:
 ```
| ns/element |           element/s |    err% |     total | benchmark
| 0.21       |    4,782,929,097.42 |    0.2% |      0.00 | `convert_downgrade(eve::avx_abi_v0::wide<long unsigned int, eve::fixed<4> >)`
 ```

After patch:
 ```
| ns/element |           element/s |    err% |     total | benchmark
| 0.16       |    6,402,093,057.96 |    0.1% |      0.00 | `convert_downgrade(eve::avx_abi_v0::wide<long unsigned int, eve::fixed<4> >)`
 ```

### DIFF
--- a/include/eve/module/real/core/function/regular/simd/x86/convert_256.hpp
+++ b/include/eve/module/real/core/function/regular/simd/x86/convert_256.hpp
@@ -16,7 +16,7 @@
 namespace eve::detail
 {
   //================================================================================================
-  // 256 bits ==> 256 bits
+  // 256 bits ==> 256/128 bits
   // We need to care only for AVX-sized input, which means we'll only convert by using a
   // Split/Compute/Combine skeletons (my PHD director would be proud :p).
   // We don't call aggregate cause aggregate wants EVERYTHING to be sliceable.
@@ -29,6 +29,14 @@ namespace eve::detail
     if constexpr( std::is_same_v<In, Out> )
     {
       return v0;
+    }
+    else if constexpr( current_api >= avx2 && (N::value == 4)
+                       && std::is_integral_v<In> && (sizeof(In) == 8)
+                       && std::is_integral_v<Out> && (sizeof(Out) == 4) )
+    {
+      // u64 to u32
+      const auto p = _mm256_permutevar8x32_epi32(v0, _mm256_set_epi32(7, 5, 3, 1, 6, 4, 2, 0));
+      return _mm256_extractf128_si256(p, 0);
     }
     else
     {
--- a/include/eve/module/real/core/function/regular/simd/x86/convert_256.hpp
+++ b/include/eve/module/real/core/function/regular/simd/x86/convert_256.hpp
@@ -31,11 +31,11 @@ namespace eve::detail
       return v0;
     }
     else if constexpr( current_api >= avx2 && (N::value == 4)
-                       && std::is_integral_v<In> && (sizeof(In) == 8)
-                       && std::is_integral_v<Out> && (sizeof(Out) == 4) )
+                       && std::integral<In> && (sizeof(In) == 8)
+                       && std::integral<Out> && (sizeof(Out) == 4) )
     {
       // u64 to u32
-      const auto p = _mm256_permutevar8x32_epi32(v0, _mm256_set_epi32(7, 5, 3, 1, 6, 4, 2, 0));
+      auto const p = _mm256_permutevar8x32_epi32(v0, _mm256_set_epi32(7, 5, 3, 1, 6, 4, 2, 0));
       return _mm256_extractf128_si256(p, 0);
     }
     else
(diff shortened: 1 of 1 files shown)


include/eve/module/real/core/function/regular/simd/x86/convert_256.hpp
```diff
@@ -30,6 +30,14 @@ namespace eve::detail
     {
       return v0;
     }
+    else if constexpr( current_api >= avx2 && (N::value == 4)
+                       && std::is_integral_v<In> && (sizeof(In) == 8)
```
jfalcou, line 34: Matter of style and compile-time, prefer using `std::integral` concept instead of the traits.

include/eve/module/real/core/function/regular/simd/x86/convert_256.hpp
```diff
@@ -30,6 +30,14 @@ namespace eve::detail
...
+                       && std::is_integral_v<Out> && (sizeof(Out) == 4) )
+    {
+      // u64 to u32
+      const auto p = _mm256_permutevar8x32_epi32(v0, _mm256_set_epi32(7, 5, 3, 1, 6, 4, 2, 0));
```
jfalcou, line 38: `auto const` :p

jfalcou (commented):
Some style comments

include/eve/module/real/core/function/regular/simd/x86/convert_256.hpp
```diff
@@ -30,6 +30,14 @@ namespace eve::detail
     {
       return v0;
     }
+    else if constexpr( current_api >= avx2 && (N::value == 4)
+                       && std::is_integral_v<In> && (sizeof(In) == 8)
+                       && std::is_integral_v<Out> && (sizeof(Out) == 4) )
```
DenisYaroshevskiy, line 35: Hmm, what about signed unsigned?

Should probably be OK, I dunno what we have in terms of tests here.

This covers I guess: https://github.com/jfalcou/eve/blob/d1885fd6e0849899ab070ab35bb313aafbf187a1/test/unit/module/real/core/convert/to_int32.cpp#L60-L71

I did check - the intrinsics, I think this shuffle is your best bet unfortunatly.

include/eve/module/real/core/function/regular/simd/x86/convert_256.hpp
```diff
@@ -30,6 +30,14 @@ namespace eve::detail
...
+    {
+      // u64 to u32
+      const auto p = _mm256_permutevar8x32_epi32(v0, _mm256_set_epi32(7, 5, 3, 1, 6, 4, 2, 0));
+      return _mm256_extractf128_si256(p, 0);
```
DenisYaroshevskiy, line 39: Please use: `_mm256_castsi256_si128 `
I'd hope compiler optimises this but it's not harder.

DenisYaroshevskiy (commented):
Small nit.

include/eve/module/real/core/function/regular/simd/x86/convert_256.hpp
aguinet: IMO signed and unsigned are both covered by this case. Indeed, for me a downcast of s64 => s32 is the same as u64 => u32, thanks to the two complements representation.

include/eve/module/real/core/function/regular/simd/x86/convert_256.hpp
aguinet: `_mm256_extractf128_si256` already returns a `__m128i` by design : https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm256_extractf128_si256&ig_expand=2989

include/eve/module/real/core/function/regular/simd/x86/convert_256.hpp
aguinet: Fixed in 3ca5c9d3a2ad7b252bb269f11d6dbd14d3418520

include/eve/module/real/core/function/regular/simd/x86/convert_256.hpp
aguinet: For instance, LLVM has a `zext` and `sext` operation, but only a `trunc` one: https://llvm.org/docs/LangRef.html#conversion-operations

include/eve/module/real/core/function/regular/simd/x86/convert_256.hpp
DenisYaroshevskiy: That's not what I meant.

You don't need extract. Extract has a latency 3. Cast does the same thing but with latency 0.

Llvm can optimize that but I don't know about MSVC.

Generally we prefer not to rely on optimizer if it doesn't cost us anything

DenisYaroshevskiy: So put `_mm256_castsi256_si128` instead please

aguinetqb: Done in https://github.com/jfalcou/eve/pull/1049/commits/6a6887de53bb7b78be9159c4dbdf642d5e92de73

DenisYaroshevskiy:
Lgtm, thanks for the contribution!

Will wait for Joel to sign off

jfalcou (approved):

PR #1050, opened 2021-11-02
Fix #1047 - G++ 11 issue in -O3

### LINKED ISSUE #1047: [BUG] g++ 11.1 aliasing bugs in -O2/-O3
For some reasons, g++ has aliasing issue that clang don't in -O3. More than just failing on some complex cases, it spuriously fails in simple stuff like constant generation.

Investigation shows that the culprit is the common implementation for get/srt that use reinterpret_cast in a too liberal way

Potential fix is to change to using memcpy over a local array of scalar than accessing the array. The codegen is correct but we need to check it fixes everything

jfalcou:


DenisYaroshevskiy (approved):
Dunno if you are done but looks good so far

PR #1051, opened 2021-11-04
Fix/pow abs

jtlap:
compile time reduced for pow_abs and gamm_p_inv tests


include/eve/module/real/math/function/regular/generic/exp.hpp
```diff
@@ -32,12 +32,19 @@
...
 #include <eve/module/real/core/detail/generic/horn.hpp>
 #include <type_traits>
+#include <cmath>
 
 namespace eve::detail
 {
   template<floating_real_value T>
   EVE_FORCEINLINE constexpr T exp_(EVE_SUPPORTS(cpu_), T x) noexcept
   {
+//     if constexpr(scalar_value<T>)
```
jfalcou, line 42: delet don't comment

PR #1058, opened 2021-11-09
Replace std::forward by EVE_FWD

jfalcou:
For compile time purpose

### DIFF
--- a/include/eve/detail/abi.hpp
+++ b/include/eve/detail/abi.hpp
@@ -7,6 +7,9 @@
 //==================================================================================================
 #pragma once
 
+// Faster than std::forward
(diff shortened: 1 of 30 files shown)


include/eve/detail/abi.hpp
```diff
@@ -7,6 +7,9 @@
 //==================================================================================================
 #pragma once
 
+// Faster than std::forward
+#define EVE_FWD(...) static_cast<decltype(__VA_ARGS__) &&>(__VA_ARGS__)
+
```
DenisYaroshevskiy, line 12: Is there any reason it's variadic?

DenisYaroshevskiy (approved):

include/eve/detail/abi.hpp
jfalcou: So we can call on expression containing comma

PR #1065, opened 2021-11-16
FIX-1060: reverse_copy

DenisYaroshevskiy:
* next_aligned
* reverse_conditional
* views::reverse 
* reverse_copy


include/eve/conditional.hpp
```diff
@@ -533,4 +544,17 @@ namespace eve
     if constexpr (!C::has_alternative) return c;
     else                               return c.map_alternative(map);
   }
+
+  template <eve::relative_conditional_expr C, typename T>
+  constexpr auto reverse_conditional(C c, eve::as<T> tgt)
+  {
+         if constexpr ( C::is_complete                  ) return c;
```
jfalcou, line 551: I was about to comment on this to be too closed for extension but we agree we have all the ignore* we need anyway right ?

PR #1067, opened 2021-11-17
FIX-1066: reverse

DenisYaroshevskiy:
* swap_ranges
* reverse

Measurements results.

* unrolling does not help.
* gcc can do reverse in simd by itself but clang can't

### Win over scalar for 1000 bytes

newplot (43)
 9 times for chars, 4 times for shorts, 6.7 times for ints.
 
###  Win over scalar for 10'000 bytes

newplot (44)

16.4 times for chars, 8.2 times for shorts, 6.53 times for ints.

### Can we do better?

My experimental implementation of reverse used a custom iteration pattern.
Here we just use `swap_ranges`.
Custom iteration pattern is better, especially on smaller sizes even though it makes things significantly more difficult to implement. I suspect this is due to masked loads/stores, current implementation requires more of them.

### Custom iteration pattern, 1000 bytes
 
newplot (45)

Almost two times faster (13ns vs 26 ns) for chars and shorts, 1ns diff for ints, probably noise.

### Custom iteration pattern, 1000 bytes

newplot (46)

Same 13 ns difference for chars and shorts. 

### Conclusion. 

Measure without aligning, maybe the difference will stop being significant, maybe it won't.
If not - implement and test custom aligning in eve.

### DIFF
--- a/include/eve/algo/reverse.hpp
+++ b/include/eve/algo/reverse.hpp
@@ -7,11 +7,50 @@
 //==================================================================================================
 #pragma once
 
+#include <eve/algo/as_range.hpp>
 #include <eve/algo/copy.hpp>
+#include <eve/algo/preprocess_range.hpp>
+#include <eve/algo/swap_ranges.hpp>
 #include <eve/algo/views/reverse.hpp>
 
 namespace eve::algo
 {
+  //================================================================================================
+  //! @addtogroup eve.algo
+  //! @{
+  //!  @var reverse
+  //!
+  //!  @brief version of std::reverse
+  //!    * default unrolling is 1.
+  //!    * will align by default.
+  //!
+  //!   **Required header:** `#include <eve/algo/reverse.hpp>`
+  //!
+  //! @}
+  //================================================================================================
+
+  template <typename TraitsSupport>
+  struct reverse_ : TraitsSupport
+  {
+    template <relaxed_range Rng>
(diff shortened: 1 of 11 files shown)


include/eve/algo/reverse.hpp
```diff
@@ -7,11 +7,50 @@
...
+
+      // To utilize extra information from range prerprocessing.
+      auto processed = eve::algo::preprocess_range(TraitsSupport::get_traits(), EVE_FWD(rng));
+      auto f = processed.begin();
+      auto l = processed.end();
+
+      std::ptrdiff_t n = l - f;
+      auto m = unalign(f) + n / 2;
+
+      swap_ranges[processed.traits()](as_range(f, m), views::reverse(l));
```
DenisYaroshevskiy, line 48: Hmm.

I think this is not quite right.
This will pass down divisible by cardinal, but halve is not necessary divisible by cardinal.
I don't think I should pass traits down.

This means not sufficient testing.

DenisYaroshevskiy: If `n` is divisible by cardinal then `n / 2` is divible by cardinal unless n == cardinal. And that's the missing test.

DenisYaroshevskiy: That's not true. 3 n is divisible. I was not reading the logs well.

jfalcou (approved):

PR #1071, opened 2021-11-24
Interleave shuffle

jfalcou:


DenisYaroshevskiy (approved):

jfalcou:
The AVX512 logical codegen is bad, I need to write a bit interleave version

PR #1078, opened 2021-11-29
Improve convert u64=>u32 using AVX2 + clang

aguinet:
See inline comment as to how & why


jfalcou (approved):

DenisYaroshevskiy:
Dunno why this wasn't merged

Issue #815, opened 2021-07-17
[BUG] Testing for non zero values in the beginning of a small wide.


DenisYaroshevskiy:
Sometime back we decided that for wides that are smaller then expected cardinal, we will have any garbage in the top values.
I suspect there are issues because of this.
These bugs are very subtle, this is UB.
At least we need to adjust the input generation, so that the values before the beginning are some garbage.

-----------------------------

Proposed resolution

Sub Goal:
 - [x] Provide traits like `fundamental_cardinal` or something that gives the smallest cardinal before having to deal with unused lanes (#1024)
 - [x] Modify the value generator to specifically insert recognizable garbage in unused lanes (#1084)
 - [x] Fix wide and logical constructors to not advertently let lanes uninitialized with garbage (#1092)
 - [x] Add tests in fundamental function (shuffle, reduction, conditional operations) so that we know we don't touch the garbage (#1092)

jfalcou:
This can be done at the test level.

jfalcou:
Idea: force substandard wide constructor that may introduce spurrious values to check for size and bcast a 0 if required before.
Example.

Before:

```c++
    template<std::invocable<size_type,size_type> Generator>
    EVE_FORCEINLINE wide(Generator &&g) noexcept
                  : storage_base( detail::fill(eve::as<wide>{}, std::forward<Generator>(g)) )
    {}
```

After:


```c++
    template<std::invocable<size_type,size_type> Generator>
    EVE_FORCEINLINE wide(Generator &&g) noexcept requires( has_expected_size<Wide> )
                  : storage_base( detail::fill(eve::as<wide>{}, std::forward<Generator>(g)) )
    {}

    template<std::invocable<size_type,size_type> Generator>
    EVE_FORCEINLINE wide(Generator &&g) noexcept requires( !has_expected_size<Wide> )
                  : storage_base( 0 )
    {
       storage() = detail::fill(eve::as<wide>{}, std::forward<Generator>(g)) ;
    }
```

jfalcou:
Sub Goal:
 - [ ] Provide traits like `fundamental_cardinal` or something that gives the smallest cardinal before having to deal with unused lanes
 - [ ] Generalize to all architecture the `arm_cleanup` introduced in the sum PR so we have a centralized cleanup system
 - [ ] Modify the value generator to specifically insert recognizable garbage in unused lanes
[...]

PR #1084, opened 2021-12-08
#815 - Poison data in test to find issues with sub-sized wide

jfalcou:


test/generator.hpp
```diff
@@ -33,6 +35,38 @@ namespace eve::test
...
+      using ftype = as_wide_t<v_t, eve::fundamental_cardinal_t<v_t>>;
+
+      ftype these(data.storage());
+
+      // Compute a recognizable filler
+      auto filler = eve::Constant<p_t, static_cast<p_t>(0xDEADBEEFC007BABE)>();
+      for(std::ptrdiff_t i=data.size();i<these.size();++i)
+      {
+        these.set(i, filler);
+      }
```
DenisYaroshevskiy, line 60: Can we use different values?

jfalcou: I'll try to mix the position of the data in the lower bits

jfalcou: Done

DenisYaroshevskiy: Is the top bit 1? So that it affects movemask?

DenisYaroshevskiy (approved):

PR #1090, opened 2021-12-11
Export Bessel functions as a pseudo-module

jfalcou:
As per [this discussion](https://github.com/jfalcou/eve/discussions/1085), this is a first step to extract existing functions into separate module-like entities.

This is a transitional state before actual modules are usable on most compilers.


DenisYaroshevskiy (approved):
I don't understand the logic of moving things around but it's OK.

PR #1091, opened 2021-12-11
Fix #1079 - Implements `has_aggregated_component`

### LINKED ISSUE #1079: [FEATURE] `has_aggregated_component`
A `concept` that tells wether a `bundle` wide has at least one aggregated component.
I would also like this to return `true` for just plain `aggregated_abi_`

jfalcou:


DenisYaroshevskiy (commented):
Missing tests for multi-layer

DenisYaroshevskiy (approved):

PR #1093, opened 2021-12-11
Remove remnant of restricted_types

jfalcou:
Some tests were still using the old AVX512 fixed test types. This removes the restricted:: namespace and fi the tests.

### DIFF
--- a/test/unit/api/regular/shuffle/slide_left.cpp
+++ b/test/unit/api/regular/shuffle/slide_left.cpp
@@ -6,19 +6,16 @@
 **/
 //==================================================================================================
 #include "test.hpp"
-
 #include <eve/function/slide_left.hpp>
-
 
 //==================================================================================================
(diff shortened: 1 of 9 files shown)


test/unit/api/regular/shuffle/slide_left.cpp
```diff
@@ -6,19 +6,16 @@
...
 //==================================================================================================
 // slide_left test
 //==================================================================================================
-
-
 EVE_TEST_TYPES("Check behavior of slide_left shuffle", eve::test::simd::all_types)
 <typename T>(eve::as<T>)
 {
-  T x{[](int i, int) { return i; }};
+  using v_t = eve::element_type_t<T>;
```
DenisYaroshevskiy, line 17: We usually call this e_t but whatever

DenisYaroshevskiy (approved):

PR #1115, opened 2021-12-25
FIX-946: convert from non-const to const for iterators

DenisYaroshevskiy:
Had to make a few changes in `aligned_ptr` - construction from pointer has to be `explicit`, conversion to `pointer` - implicit.
This is to allow for `equality_comparible_with`.

Also - due to conversion to `pointer` from `aligned_pointer` now being implicit, had to remove `operator[]` since it somehow conflicted with a built in `operator[]` on gcc for arm-v7 (and only there).

The functionality still works.
I can disable with a macro if you prefer.


jfalcou (approved):
I'm OK with all this, the operator[] on aligned_ptr was anyway probably a bad idea.

PR #1125, opened 2021-12-31
First jab at emulating std::experimental::simd types

jfalcou:


include/eve/std.hpp
```diff
@@ -0,0 +1,32 @@
...
+
+  template<typename T>
+  using simd_mask = eve::logical<eve::wide<T, eve::fixed<16/sizeof(T)>>>;
+
+  template<typename T, int N >
+  using fixed_size_simd_mask = eve::logical<eve::wide<T, eve::fixed<N>>>;
+
+  template<typename T>
+  using native_simd_mask = eve::experimental::simd_mask<T>;
+}
```
DenisYaroshevskiy, line 32: New line

include/eve/std.hpp
```diff
@@ -0,0 +1,32 @@
...
+//==================================================================================================
+#pragma once
+
+#include <eve/wide.hpp>
+#include <eve/logical.hpp>
+
+namespace eve::experimental
+{
+  template<typename T>
+  using simd = eve::wide<T, eve::fixed<16/sizeof(T)>>;
```
DenisYaroshevskiy, line 16: Rly? Always 16?

DenisYaroshevskiy (commented):
Seems reasonable

include/eve/std.hpp
jfalcou: I dunno, clang and g++ gives different size. I purposedly let this and I await the failure to update accordingly.

include/eve/std.hpp
jfalcou: OK

include/eve/std.hpp
DenisYaroshevskiy: OK. Let's do that.

jfalcou:
Looks like it passes.
@DenisYaroshevskiy your opinion ?

DenisYaroshevskiy (approved):

Issue #1126, opened 2022-01-03
[BUG] The dependency target "..." of target "..." does not exist.


the-moisrex:
**Describe the bug**
Disabling building of tests, benchmarks, and documentations will cause this error while using CPM.cmake:

````
CMake Error at /home/$USER/.cache/CPM/eve/0473c468ff0d44c6aa8d466a61593cbb877720ec/cmake/target/targets.cmake:29 (add_dependencies):
The dependency target "doc.exe" of target "unit.exe" does not exist.
Call Stack (most recent call first):
/home/$USER/.cache/CPM/eve/0473c468ff0d44c6aa8d466a61593cbb877720ec/CMakeLists.txt:41 (include)


CMake Error at /home/$USER/.cache/CPM/eve/0473c468ff0d44c6aa8d466a61593cbb877720ec/cmake/target/targets.cmake:38 (add_dependencies):
The dependency target "unit.api.exe" of target "unit.basic.exe" does not
exist.
Call Stack (most recent call first):
/home/$USER/.cache/CPM/eve/0473c468ff0d44c6aa8d466a61593cbb877720ec/CMakeLists.txt:41 (include)


CMake Error at /home/moisrex/.cache/CPM/eve/0473c468ff0d44c6aa8d466a61593cbb877720ec/cmake/target/targets.cmake:37 (add_dependencies):
The dependency target "unit.arch.exe" of target "unit.basic.exe" does not
exist.
Call Stack (most recent call first):
/home/$USER/.cache/CPM/eve/0473c468ff0d44c6aa8d466a61593cbb877720ec/CMakeLists.txt:41 (include)


CMake Error at /home/$USER/.cache/CPM/eve/0473c468ff0d44c6aa8d466a61593cbb877720ec/cmake/target/targets.cmake:39 (add_dependencies):
The dependency target "unit.meta.exe" of target "unit.basic.exe" does not
exist.
Call Stack (most recent call first):
/home/$USER/.cache/CPM/eve/0473c468ff0d44c6aa8d466a61593cbb877720ec/CMakeLists.txt:41 (include)
````

CPM code:

````cmake
CPMAddPackage(
        NAME eve
        URL https://github.com/jfalcou/eve/archive/refs/tags/v2021.10.0.zip
        VERSION 2021.10.0
        OPTIONS
                "EVE_BUILD_TEST OFF"
                "EVE_BUILD_BENCHMARKS OFF"
                "EVE_BUILD_DOCUMENTATION OFF"
)

[...]

PR #1124, opened 2021-12-30
Revamped docs to add basic 101 tutorials

jfalcou:
Tutorials now goes basic wide function -> basic algorithms -> advanced algorithms -> SIMD tuple/UDT
Added an advanced tutorial on manual loop

PR #1127, opened 2022-01-03
fixing disabling tests bug (#1126)

the-moisrex:
Fixing #1126.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -36,13 +36,13 @@ endif()
 ## =================================================================================================
 ## Test and Benchmarks target
 ## =================================================================================================
-include(CTest)
-include(target/add_parent_target)
-include(config/compiler)
-include(config/types)
-include(config/pch)
-
 if( EVE_BUILD_TEST )
+  include(CTest)
+  include(target/add_parent_target)
+  include(config/compiler)
+  include(config/types)
+  include(config/pch)
+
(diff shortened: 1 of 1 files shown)


jfalcou:
I think this is exactly what is needed.

jfalcou (approved):

CMakeLists.txt
```diff
@@ -36,13 +36,13 @@ endif()
...
-include(config/compiler)
-include(config/types)
-include(config/pch)
-
 if( EVE_BUILD_TEST )
+  include(CTest)
+  include(target/add_parent_target)
+  include(config/compiler)
+  include(config/types)
+  include(config/pch)
```
the-moisrex, line 44: This file which is included here adds "documentation" related executable while the if above will make sure that executable won't be created.

Meaning that I think we're breaking things for someone who disables "tests" and enables "documentation" here.

jfalcou: As said in the issue, this documentation source thing is gonna be removed by the doc PR anyway so only test and bench options will remain.

jfalcou: See https://github.com/jfalcou/eve/pull/1124/files?file-filters%5B%5D=.cmake&hide-deleted-files=true

the-moisrex: Oh, that's okay then. thanks

PR #1134, opened 2022-01-07
Some more documentation

jfalcou:


include/eve/literals.hpp
```diff
@@ -29,27 +29,88 @@ namespace eve::detail
...
+  //! @groupheader{Wide literals}
+  //! Suffixes are provided to construct a eve::wide instance from a literal constant.
+  //!
+  //! @code
+  //! using namespace eve::literal;
+  //! // x is a eve::wide<short>
+  //! auto x = 45_wi16;
+  //!
+  //! // x is a eve::wide<float>
+  //! auto y = 2.635_wf;
```
DenisYaroshevskiy, line 68: Omg rly? :)
Why would we need this?

include/eve/platform.hpp
```diff
@@ -12,28 +12,43 @@
...
+  //! @brief  Platform specific constexpr information
+  //!
+  //! Provides constexpr information about the status (supported or not) of various platform
+  //! computing capability.
+  //! @}
+  //================================================================================================
   struct platform
   {
+    //! @brief Evaluates to `true` if current platform supports NaN values
 #if defined(EVE_NO_NANS)
```
DenisYaroshevskiy, line 28: Is this rly a thing? Where?

include/eve/platform.hpp
```diff
@@ -12,28 +12,43 @@
...
     static constexpr bool supports_nans = false;
 #else
     static constexpr bool supports_nans = true;
 #endif
 
+    //! @brief Evaluates to `true` if current platform supports infinites values
```
DenisYaroshevskiy, line 34: Same here? Who doesn't do IEEE?

DenisYaroshevskiy (approved):
I'm OK with this. I'd just remove most of it though.

include/eve/literals.hpp
jfalcou: It's an old piece of code I forgot existed and I documented it.
I am not attached to it so if you want we can remove it.

include/eve/platform.hpp
jfalcou: A32 ARM doesn't support denormal values in SIMD register  for example. ARM code and tests use this supports_denormals cause if not, they fail.

The others are also meant to be passed by users in order to globally control the fact our SIMD functions care or do not care of some corner cases in FP to win some speed globally. 

I should add doc for the macro or if not this struct looks strange.

include/eve/platform.hpp
jfalcou: See above

include/eve/literals.hpp
jfalcou: Let's remove them

PR #1142, opened 2022-01-08
FIX-1013: remove detail::value type

DenisYaroshevskiy:
We would like `eve::value_type` to have ranges semantics and use `eve::element_type` for maybe going from wide to scalar.

This required:
a few concept changes
removing sign_of, since I couldn't make it work for wide in meta and it was hacked anyways. It's used only in one place, so I just implemented a version there.
A lot of small changes.

You can review by commits if you want to.


include/eve/module/proba/regular/impl/lognormal_distribution.hpp
```diff
@@ -270,7 +270,7 @@ namespace eve
         return median(d)*exp(-sqr(d.s));
       else
       {
-        value_type inve = I(0.36787944117144232159552377016146086744581113103176);
+        auto inve = I(0.36787944117144232159552377016146086744581113103176);
```
jfalcou, line 273: this value_type is a local typedef, I think it should stay the way it was.

include/eve/module/real/algorithm/function/regular/generic/reduce.hpp
```diff
@@ -54,26 +54,26 @@ namespace eve::detail
     return op(s, v);
   }
 
-  template<real_scalar_value T, typename N, typename Callable>
+  template<scalar_value T, typename N, typename Callable>
```
jfalcou, line 57: Why this change ? I guess tests on tuple-like types ?

include/eve/traits/as_integer.hpp
```diff
@@ -9,15 +9,26 @@
...
 #include <eve/forward.hpp>
 #include <eve/detail/meta.hpp>
+#include <eve/traits/element_type.hpp>
 
 namespace eve
 {
   namespace detail
   {
-    template<typename T> struct sign_of<logical<T>> : sign_of<T> {};
+    template<typename T>
```
jfalcou, line 18: if it works here, why not put it back in meta.hpp as sign_of<T> ?

jfalcou (changes requested):

include/eve/module/real/algorithm/function/regular/generic/reduce.hpp
DenisYaroshevskiy: Well reduce works on bundles. And somehow concepts restrictions here didn't work.

include/eve/traits/as_integer.hpp
DenisYaroshevskiy: I can't use `element_type_t` in `meta` - it depends on `forward` which depends on `meta`.
Also `meta` does not know about `eve` types - this is why you had the hack here for logicals.

include/eve/module/proba/regular/impl/lognormal_distribution.hpp
DenisYaroshevskiy: It's not a local typedef, it's likely dead/untested code.

There is no `value_type` defined. It was just never compiled.
And now there's no `value_type` struct, so even in a not-instantiated form it's not working.

include/eve/module/real/algorithm/function/regular/generic/reduce.hpp
jfalcou: Good

include/eve/traits/as_integer.hpp
jfalcou: OK good

jfalcou (approved):

PR #1146, opened 2022-01-09
first cleaning of distribs

jtlap:


include/eve/module/proba/regular/impl/ev_distribution.hpp
```diff
@@ -288,10 +288,12 @@ namespace eve
     EVE_FORCEINLINE  auto mode_(EVE_SUPPORTS(cpu_)
                                , ev_distribution<T,U,I> const & d) noexcept
     {
-      if constexpr (floating_value<T>)
-        return d.m;
-      return
-        zero(as<I>());
+      if constexpr (std::same_as<eve::callable_zero_, T>)
```
jfalcou, line 291: What about:

```
 if constexpr (floating_value<T>)  return d.m;
else                               return T{}(as<typename ev_distribution<T,U,I>::value_type>());
```

?

jtlap: si tu veux, mais ce n'est ni +court ni + clair

PR #1147, opened 2022-01-10
Infra/arithmetic2

jtlap:


jfalcou:
Still needed :
 * Move geommean to math
 * Move div_180 to math
 * Move dist to arithmetic
 * Move deg_inrad etc to math
 * Eliminate ffs/fls
 * Remove old .cpp from core

Issue #1103, opened 2021-12-18
[FEATURE] Remove x[pattern] and just use shuffle() function instead


jfalcou:
The `wide::operator[]` is massively abused and it leads to strange dependency.
We're going back to a proper shuffle function that does the optimization of patterns.

- [x] Move all specific swizzle to function/ (4ff00c0)
- [x] Move lookup to function (2c5cac4)
- [x] Add a shuffle function that perform pattern recognition + calls (fc06095)
- [x] Remove the operator[] performing swizzle/lookup from wide/logical

jfalcou:
- [ ] Move all specific swizzle to function/
- [ ] Move lookup to function
- [ ] Add a shuffle function that perform pattern recognition + calls
- [ ] Remove the operator[] performing swizzle/lookup from wide/logical

PR #1149, opened 2022-01-11
#1103 - Task 4/4 : Remove operator[] for shuffling and replace it by shuffle

jfalcou:


include/eve/function/shuffle.hpp
```diff
@@ -7,28 +7,30 @@
...
-#include <eve/detail/function/swizzle.hpp>
+#include <eve/module/real/core/detail/swizzle.hpp>
 
 namespace eve
 {
   EVE_MAKE_CALLABLE(shuffle_, shuffle);
+}
+
+
+#include <eve/module/real/core/detail/generic/patterns.hpp>
```
DenisYaroshevskiy, line 18: Why are patterns in detail?

DenisYaroshevskiy: Ah.

Seems like 'patterns' maybe should be renamed as 'find_optimized_pattern`? Or `find_optimized_shuffle_pattern`?

include/eve/module/real/core/detail/generic/swizzle.hpp
```diff
@@ -35,7 +37,7 @@ namespace eve::detail
                                       ) noexcept
   {
     constexpr auto sz = Pattern::size();
-    return bit_cast( v.mask()[p], as<logical<wide<T,fixed<sz>>>>() );
+    return bit_cast( shuffle(v.mask(),p), as<logical<wide<T,fixed<sz>>>>() );
```
DenisYaroshevskiy, line 40: Do we want to keep the name `swizzle` even if we use the `shuffle` function?

DenisYaroshevskiy (approved):
A bit complaining about names, but this is an improvement - so it can go as is.

include/eve/function/shuffle.hpp
jfalcou: Yes I was thinking about that to prevent confusion

include/eve/module/real/core/detail/generic/swizzle.hpp
jfalcou: swizzle is the bare bone of shuffle. I just moved it to where it belongs.
We can rename it to basic_shuffle

DenisYaroshevskiy (approved):
still looks good

PR #1155, opened 2022-01-13
No need to support XOP/FMA4

jfalcou:


include/eve/arch/x86/detection.hpp
```diff
@@ -29,7 +29,5 @@ namespace eve
     else                                            return false;
   }
 
-  inline bool is_supported(xop_ const &)  noexcept  { return detail::cpuid_states.supports_xop(); }
   inline bool is_supported(fma3_ const &) noexcept  { return detail::cpuid_states.supports_fma3();}
```
DenisYaroshevskiy, line 32: Is fma3 still a thing? Are we testing this?

jfalcou: FMA3 is the FMA for AVX/AVX2.
We can easily add it to the test matrix.

DenisYaroshevskiy (approved):

PR #1157, opened 2022-01-14
Fix 1013: intoduce soa ptr

DenisYaroshevskiy:


jfalcou:
Why not `soa_ptr<int, char const, double>`  instead of `soa_ptr<int*, char const*, double*>`  ?

jfalcou (approved):

PR #1178, opened 2022-01-20
FIX-1170: deinterleave_groups_shuffle

DenisYaroshevskiy:
Implementation of `deinterleave_groups_shuffle` (just for 2 wides).
Did not do power since we just do what should go in the basic shuffle (when we have one).

Did not test `neon_struct_to_wide.hpp` - we'll test it with what we touch it with I guess.

I also left tests for G == 1 duplicated, because they helped me debug and they are easier inspectible.

(fixed a couple of issues in other shuffles that I saw).


include/eve/module/real/core/function/regular/simd/x86/deinterleave_groups_shuffle.hpp
```diff
@@ -0,0 +1,219 @@
...
+    constexpr auto g_sz = sizeof(T) * G;
+    constexpr auto n  = N() / G;
+
+    // we start with 4, 2 groups is a no op
+    if constexpr ( n == 4 )
+    {
+      // g_sz 16
+      // I can only do in 2 heavy instructions, or rely on a mask in a register.
+      // I picked register - that's done by the basic shuffle.
+
```
jfalcou, line 28: where is the code then ?

test/unit/api/regular/shuffle/deinterleave_groups_shuffle.cpp
```diff
@@ -0,0 +1,109 @@
...
+  eve::wide<eve::element_type_t<T>, eve::fixed<T::size() * 2>>
+   expected { [](int i, int size) {
+    if (i < size / 2 || size == 1) return 0xA0 | (i & 7);
+    i -= size / 2;
+    return 0xB0 | (i & 7);
+  }};
+
+  auto actual = eve::deinterleave_groups_shuffle(a, b, eve::lane<1>);
+
+#if 0
```
jfalcou, line 40: if it's not needed, just delete this maybe ?

include/eve/module/real/core/function/regular/simd/x86/deinterleave_groups_shuffle.hpp
DenisYaroshevskiy: In the very very last else.

test/unit/api/regular/shuffle/deinterleave_groups_shuffle.cpp
DenisYaroshevskiy: It's very useful for debugging. I don't think we have 'print if fails in TTS' or the ability to attach to error msgs more stuff.

I like the Catch2 approach for that if you feel like doing that at some point btw 🙂

PR #1200, opened 2022-01-24
FIX-1170: deinterleave groups for 2,4 wides + emulation

DenisYaroshevskiy:
2 and 4 wides will work well.
Test from 1 to 5.

### DIFF
--- a/include/eve/detail/function/simd/common/deinterleave_groups.hpp
+++ b/include/eve/detail/function/simd/common/deinterleave_groups.hpp
@@ -20,7 +20,7 @@ namespace eve::detail
     EVE_FORCEINLINE auto select_every_step_from_tuple(T t)
     {
       return [&]<std::size_t ...i>(std::index_sequence<i...>) {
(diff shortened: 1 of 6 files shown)


include/eve/detail/function/simd/common/deinterleave_groups.hpp
```diff
@@ -0,0 +1,114 @@
...
+
+namespace eve::detail
+{
+  namespace _deinterleave_groups
+  {
+    template <std::size_t first, std::size_t step, kumi::product_type T>
+    EVE_FORCEINLINE auto select_every_step_from_tuple(T t)
+    {
+      return [&]<std::size_t ...i>(std::index_sequence<i...>) {
+        return kumi::make_tuple(get<first + i * step>(t) ...);
```
jfalcou, line 23: `return kumi::reorder<(first + i * step)...>(t)` seems simpler

DenisYaroshevskiy: Done

jfalcou (approved):

Issue #1068, opened 2021-11-19
[FEATURE] reverse perf improvements


DenisYaroshevskiy:
See comments from https://github.com/jfalcou/eve/pull/1067

DenisYaroshevskiy:
Well, I tried the `no_aligning`, it does not give good results
![newplot (47)](https://user-images.githubusercontent.com/11256077/145902228-bfca321c-9a1f-417a-8df6-2846c82e9f73.png)

So for newer arches for sure we need to do custom iteration.
Older arches is a good question, because sse2 will give different perf on my machine and on a machine where it was the actual architecture.

But I suspect that aligning one of the reads/writes wouldn't be valuable there too.

DenisYaroshevskiy:
![newplot (41)](https://user-images.githubusercontent.com/11256077/151467882-6f5d00a2-086f-43c6-ac21-10bdcbd56671.png)

https://github.com/jfalcou/eve/pull/1211

PR #1211, opened 2022-01-27
FIX-1068: reverse custom loop

DenisYaroshevskiy:
Let's not merge until I measure.

See what are we doing in the issue: https://github.com/jfalcou/eve/issues/1068

### DIFF
--- a/include/eve/algo/transform.hpp
+++ b/include/eve/algo/transform.hpp
@@ -36,7 +36,7 @@ namespace eve::algo
         using out_t = eve::element_type_t<decltype(processed)>;
         auto cvt_and_store_it = views::convert(store_it, as<out_t>{});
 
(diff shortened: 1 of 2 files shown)


include/eve/algo/transform.hpp
```diff
@@ -36,7 +36,7 @@ namespace eve::algo
         using out_t = eve::element_type_t<decltype(processed)>;
         auto cvt_and_store_it = views::convert(store_it, as<out_t>{});
 
-        eve::store[ignore](op(xs), cvt_and_store_it);
+        eve::store[ignore](processed, cvt_and_store_it);
```
jfalcou, line 39: why this change ?

DenisYaroshevskiy: Because I called `op` twice - now just call once.

jfalcou (approved):

PR #1223, opened 2022-02-01
FIX-899: map view, map convert view.

DenisYaroshevskiy:
This needs more testing but we'll do that a bit later.
Maybe when the kumi issue is fixed.

Anyways - for most part should work.


include/eve/algo/concepts/nothing.hpp
```diff
@@ -0,0 +1,20 @@
+//==================================================================================================
```
jfalcou, line 1: I think this can just lives in eve/detail/meta.hpp no ?
Or is it part of the public API ?

include/eve/conditional.hpp
```diff
@@ -38,22 +38,17 @@ namespace eve
 
   namespace detail
   {
+    template <logical_simd_value Logical>
+    Logical to_logical(eve::top_bits<Logical> mmask);
+
     template<typename T, relative_conditional_expr C>
```
jfalcou, line 44: Shouldn't it just be an overload of to_logical over relative_conditional_expr ?

include/eve/algo/views/map.hpp
```diff
@@ -0,0 +1,385 @@
...
+
+    template<relative_conditional_expr C, decorator S>
+      requires iterator<I> && (!std::same_as<LoadOp, nothing_t>)
+    EVE_FORCEINLINE friend auto tagged_dispatch(eve::tag::load_,
+                                                C c,
+                                                S s,
+                                                eve::as<vw_type>,
+                                                map_iterator self)
+    {
+      auto loaded =  self.load_op(eve::load(drop_alternative(c), s,
```
jfalcou, line 341: Why do we need to drop the alternative then repalce_ignore afterward instead of just loading with an alternative?
Is it due to how we handle some traits over map?

include/eve/algo/concepts/nothing.hpp
DenisYaroshevskiy: Well, it could.
Here is my thinking:
it should be public, because the user will see it sometimes.

Like:

`map_range<std::vector<int const>, my_lamda, eve::algo::nothing_t>`

I feel in this case it should not be in detail

include/eve/algo/views/map.hpp
DenisYaroshevskiy: Well. Our alternative is provided for the mapped to type.

Let's say:

```
auto r = eve::algo::map(ints, [](auto v) { return eve::zip(v, v + v); });
// ...

eve::load[ignore_first(1).else_(kumi::tuple{0, 0}](it);
```

So, in order to load with alternative from the underlying we need to apply the inverse of load_op to the provided alternative and then apply the load_op again.
We can't quite do that - so just `replace_ignored`

include/eve/conditional.hpp
DenisYaroshevskiy: `to_logical` for `relative_conditional_expr` is called `mask` and it's a member function.

In its implementation it uses top_bits, same as before.

include/eve/algo/concepts/nothing.hpp
jfalcou: OK

include/eve/conditional.hpp
jfalcou: OK

jfalcou (approved):

Issue #1037, opened 2021-10-29
[FEATURE] avx512 aggregation for compress_store and algo


DenisYaroshevskiy:

DenisYaroshevskiy:
https://github.com/jfalcou/eve/pull/1232

PR #1232, opened 2022-02-09
FIX-1013: compress as algo 2

DenisYaroshevskiy:
The only thing that's left customising `compress_store` is zip.
That breaks in too many places.

Introduced `store_equivalent`.
It's better like this over purely customising `store` - because we want, for example, to convert the value once.

store customisation is left in `zip`, `reverse` and `backward` views.
In zip it will stay for sure. reverse and backward - maybe smth will be done.

Also https://github.com/jfalcou/eve/issues/1037 should be fixed with this PR.


include/eve/detail/compress/simd/common/compress_using_masks.hpp
```diff
@@ -118,7 +118,7 @@ namespace eve::detail
   {
     if constexpr ( kumi::product_type<T> )
     {
-      return wide<T, N>{ kumi::map(compress_using_masks_shuffle, v, num) };
+      return wide<T, N>{ kumi::map([num](auto v_) { return compress_using_masks_shuffle(v_, num); }, v) };
```
jfalcou, line 121: shouldn't it be a function object for inlining purpose ?

include/eve/traits/value_type.hpp
```diff
@@ -67,5 +67,6 @@ namespace eve
   //================================================================================================
 
   template <typename T>
+    requires (!value<T>)
```
jfalcou, line 70: why this change ?

include/eve/detail/compress/simd/common/compress_using_masks.hpp
DenisYaroshevskiy: Yeah it should. Let me do that in the next pr

include/eve/traits/value_type.hpp
DenisYaroshevskiy: It tries to get the value of doible and that results in a hard error, not sfinae error.

I don't have a concept that'd distinguish value and pointer here. So I did !value

jfalcou: OK

include/eve/detail/compress/simd/common/compress_using_masks.hpp
jfalcou: Perfect

jfalcou (approved):

PR #1233, opened 2022-02-09
First batch of fixes for MSVC support

jfalcou:
- Make some constructs more explicit
- `IN` and `OUT` are not good names
- fixed warning for converting int <-> `ptrdiff_t` in some tests
- Other small scale fixes demanded by MSVC front-end

### DIFF
--- a/include/eve/conditional.hpp
+++ b/include/eve/conditional.hpp
@@ -293,7 +293,8 @@ namespace eve
       }
       else
       {
-        using i_t   = typename as_integer<typename type::mask_type>::type;
+        using m_t = typename type::mask_type;
+        using i_t = as_integer_t<m_t>;
         if constexpr(eve::use_complete_storage<type>)
         {
           return bit_cast(detail::linear_ramp(eve::as<i_t>()) < i_t(count_), as<type>());
@@ -428,7 +429,8 @@ namespace eve
       }
       else
       {
-        using i_t   = typename as_integer<typename type::mask_type>::type;
+        using m_t = typename type::mask_type;
+        using i_t = as_integer_t<m_t>;
         constexpr std::ptrdiff_t card = cardinal_v<T>;
         if constexpr(eve::use_complete_storage<type>)
         {
@@ -568,7 +570,8 @@ namespace eve
       }
       else
       {
-        using i_t   = typename as_integer<typename type::mask_type>::type;
+        using m_t = typename type::mask_type;
+        using i_t = as_integer_t<m_t>;
         if constexpr(eve::use_complete_storage<type>)
         {
           auto const i = detail::linear_ramp(eve::as<i_t>());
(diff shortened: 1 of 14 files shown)


include/eve/conditional.hpp
```diff
@@ -293,7 +293,7 @@ namespace eve
       }
       else
       {
-        using i_t   = as_integer_t<typename type::mask_type>;
+        using i_t   = typename as_integer<typename type::mask_type>::type;
```
DenisYaroshevskiy, line 296: Are you saying that `_t` does not work on MSVC? Can you recheck? This is bizzare.

include/eve/detail/abi.hpp
```diff
@@ -25,9 +25,7 @@
 
 // Force a lambda function to be inline
 #if !defined(EVE_LAMBDA_FORCEINLINE)
-#  if defined(_MSC_VER)
-#    define EVE_LAMBDA_FORCEINLINE __forceinline
-#  elif defined(__GNUC__) && __GNUC__ > 3
+#  if defined(__GNUC__) && __GNUC__ > 3
```
DenisYaroshevskiy, line 28: This is used in one place. We should delete it.

include/eve/memory/aligned_allocator.hpp
```diff
@@ -63,7 +68,14 @@ namespace eve
     value_type *allocate(std::size_t n)
     {
       auto nbelem = align(n * sizeof(value_type), over{alignment()});
-      return static_cast<value_type *>( std::aligned_alloc(alignment(), nbelem) );
+
+      #if defined(SPY_COMPILER_IS_MSVC)
+      void *result = _aligned_malloc(nbelem,alignment());
```
DenisYaroshevskiy, line 73: [aligned free](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/aligned-free?view=msvc-170) is also a thing

include/eve/module/core/regular/impl/load.hpp
```diff
@@ -110,7 +110,7 @@ namespace eve::detail
 
 #if defined(SPY_COMPILER_IS_CLANG) or defined(SPY_COMPILER_IS_GCC)
 #define DISABLE_SANITIZERS __attribute__((no_sanitize_address)) __attribute__((no_sanitize_thread))
-#elif defined(SPY_COMPILER_IS_MSVC)
+#elif defined(SPY_COMPILER_IS_MSVC) and !defined(__clang__)
```
DenisYaroshevskiy, line 113: How is `defined(__clang__)` but not SPY_COMPILER_IS_CLANG ?
If it's clang cl, it should probably use the first DISABLE_SANITIZERS then?

include/eve/detail/abi.hpp
jfalcou: Done

include/eve/module/core/regular/impl/load.hpp
jfalcou: changed the ordering to be simpler

include/eve/memory/aligned_allocator.hpp
jfalcou: Fixed

PR #1249, opened 2022-02-20
WASM: minimum toolchain

DenisYaroshevskiy:
I hooked up wasm and was able to build something.
Unfortunatly the libstdc++ we have does not have bit cast.

emscripten moved to libc++ I think: https://github.com/emscripten-core/emscripten/blob/4c70173bf44c885801b13a12ce1777152ab4f68e/system/lib/libcxx/include/bit#L53

but that does not have bit_cast either (somehow!)

So I think updating emscrpiten at the moment won't solve the issue.
I don't think we won't to hack around `bit_cast` - so I'd say - let's merge the toolchain so we don't have to start from scratch and wait.


jfalcou (approved):

seanbaxter:
The latest libstdc++ in the Ubuntu 20.04 channel is 10.3, and that doesn't have bit_cast. I'd definitely prefer eve not having that dependency so it will build on these common machines.

jfalcou:
> The latest libstdc++ in the Ubuntu 20.04 channel is 10.3, and that doesn't have bit_cast. I'd definitely prefer eve not having that dependency so it will build on these common machines.

The issue is not only bit_cast. We use a bunch of other utilities from <bit>.
Let's move this to Discussions ok ?

jfalcou:
https://github.com/jfalcou/eve/discussions/1251

PR #1266, opened 2022-03-08
iota.hpp

jtlap:
just one file change to make scaled_iota use fma.
This does not change the codegen if fma is not hard-wired and improve it if it is available

### DIFF
--- a/include/eve/algo/views/iota.hpp
+++ b/include/eve/algo/views/iota.hpp
@@ -50,8 +50,7 @@ namespace eve::algo::views
       i(0)
     {
       wide_cur = wv_type {[&](int j, int) { return j; }};
-      wide_cur *= scale;
-      wide_cur += base;
+      wide_cur = fam(base, wide_cur, scale);
       operator+=(i_);
     }
 
@@ -71,7 +70,7 @@ namespace eve::algo::views
     base_plus_offset_iterator& operator+=(std::ptrdiff_t n)
     {
       i += n;
-      wide_cur += scale * n;
+      wide_cur =  fam(wide_cur, scale, n);
       return *this;
     }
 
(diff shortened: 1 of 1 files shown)


jtlap:
scaled_iota does not seem to me a good name see discussion #1268

include/eve/algo/views/iota.hpp
```diff
@@ -71,7 +70,7 @@ namespace eve::algo::views
     base_plus_offset_iterator& operator+=(std::ptrdiff_t n)
     {
       i += n;
-      wide_cur += scale * n;
+      wide_cur =  fam(wide_cur, scale, n);
```
DenisYaroshevskiy, line 73: The reason this was not done with fam is that `scale * n` is optimised to a constant (if you iterate with the same step which most of the time you are

DenisYaroshevskiy: Please compare asm

DenisYaroshevskiy:
@jfalcou - what did you see?\

basically - operator + should not generate fma instruction. I t should generate +. 
I don't think that fma is the good use-case here.

include/eve/algo/views/iota.hpp
DenisYaroshevskiy: You can see that this is wrong, right? We should multiply once - outside of the loop. And them just do add.
using fma prevents the correct codegen

DenisYaroshevskiy (commented):
loop always goes with the same step (or like 2 different steps)
Doing separately add and multiply allos the compiler to hoist the multiplication outside of the loop.
Using one intrinsic prevents the compiler from doing this.

`vfmadd231ps     ymm0, ymm2, ymm1` - is bad codegen
it should be 

`vfadd`

include/eve/algo/views/iota.hpp
jtlap: you are right, fma brings nothing as s*n is constant

jfalcou:
We all agree so PR closed.

PR #1270, opened 2022-03-09
renaming iota_scaled to iota_with_step

jtlap:


include/eve/algo/views/iota.hpp
```diff
@@ -36,21 +36,21 @@ namespace eve::algo::views
     using wv_type    = eve::wide<value_type, N>;
 
     value_type      base;
-    value_type      scale;
```
DenisYaroshevskiy, line 39: probably the class should be renamed in `iota_with_step_iterator`

include/eve/algo/views/iota.hpp
```diff
@@ -66,46 +66,46 @@ namespace eve::algo::views
...
-      wide_cur += scale * n;
+      wide_cur += step * n;
       return *this;
     }
 
     friend std::ptrdiff_t operator-(base_plus_offset_iterator x, base_plus_offset_iterator y)
     {
       EVE_ASSERT(x.base == y.base,   "different base");
-      EVE_ASSERT(x.scale == y.scale, "different scale");
+      EVE_ASSERT(x.step == y.step, "different step");
```
DenisYaroshevskiy, line 81: please align with the top one if this one is changed.
Same for other assertions.

jtlap: ok I'll do that.
For the alignment too, but remember that before release we intend to pass all files to clang-format-14

DenisYaroshevskiy: We didn't set up clang format yet, did we?

@jfalcou

jfalcou: we need to complete it. The last patch in clangformat14 fixed the egregious support for if constexpr. There is still
hick ups on requires clause but they are manageable.

We need to home in a final setup then proceed by part to convert existing file to new format.

include/eve/algo/views/iota.hpp
```diff
@@ -18,94 +18,94 @@ namespace eve::algo::views
...
     std::ptrdiff_t  i;
     wv_type         wide_cur;
 
-    base_plus_offset_iterator() = default;
+    iota_with_step_iterator() = default;
 
-    EVE_FORCEINLINE base_plus_offset_iterator(value_type base,
-                                              value_type scale,
+    EVE_FORCEINLINE iota_with_step_iterator(value_type base,
+                                              value_type step,
```
DenisYaroshevskiy, line 46: bad formatting :(

DenisYaroshevskiy (approved):

PR #1289, opened 2022-04-29
First batch of complex numbers support

jfalcou:
Only contains the structure, some functions and couple of operators so we can iterate on the layout and design.


include/eve/module/complex/complex.hpp
```diff
@@ -0,0 +1,147 @@
...
+
+    EVE_FORCEINLINE friend auto& operator+=(like<complex> auto& self, real_value auto o) noexcept
+    {
+      real(self) += o;
+      return self;
+    }
+
+    template<like<complex> Z1, typename Z2>
+    EVE_FORCEINLINE friend  auto operator+(Z1 const& lhs, Z2 const& rhs) noexcept
+    requires(!like<Z2,complex>)
```
DenisYaroshevskiy, line 125: It's real right? The second one? Just say so.
Or even just Type.

include/eve/module/complex/complex.hpp
```diff
@@ -0,0 +1,145 @@
...
+{
+  //================================================================================================
+  //! @addtogroup simd_types
+  //! @{
+  //================================================================================================
+  //! @brief SIMD-compatible representation of complex numbers
+  //!
+  //! **Required header:** `#include <eve/module/complex.hpp>`
+  //!
+  //! eve::complex is structure representing complex number and mean to be used in conjunction with
```
DenisYaroshevskiy, line 31: is meant

include/eve/module/complex/complex.hpp
```diff
@@ -0,0 +1,145 @@
...
+    using value_type = Type;
+
+    /// Default constructor
+    explicit  complex(Type r = 0)     noexcept : parent{r,0} {}
+              complex(Type r, Type i) noexcept : parent{r,i} {}
+
+    /// Stream insertion operator
+    friend std::ostream& operator<<(std::ostream& os, like<complex> auto const& z)
+    {
+      return os << real(z) << std::showpos << imag(z) << "i" << std::noshowpos;
```
DenisYaroshevskiy, line 52: I don't know how much we care, but this will break the state of os if that had showpos. Why do you need the modifier

include/eve/module/complex/complex.hpp
```diff
@@ -0,0 +1,145 @@
...
+    EVE_FORCEINLINE friend  auto operator+(Z1 const& lhs, Z2 const& rhs) noexcept
+    requires(requires(as_wide_as_t<Z1,Z2> t) { t += rhs; })
+    {
+      as_wide_as_t<Z1,Z2> that{lhs};
+      return that += rhs;
+    }
+
+    template<real_value Z1, like<complex> Z2>
+    EVE_FORCEINLINE friend  auto operator+(Z1 const& lhs, Z2 const& rhs) noexcept
+    requires(requires(as_wide_as_t<Z2,Z1> t) { t += lhs; })
```
DenisYaroshevskiy, line 132: just use rhs, lhs?

include/eve/module/complex/complex.hpp
```diff
@@ -0,0 +1,145 @@
...
+    {
+      return rhs + lhs;
```
DenisYaroshevskiy, line 134: I prefer x and y after Stepanov

include/eve/module/complex/regular/detail/arithmetic.hpp
```diff
@@ -0,0 +1,47 @@
...
+
+#include <eve/module/core.hpp>
+#include <eve/module/math.hpp>
+#include <eve/module/complex/regular/traits.hpp>
+#include <eve/module/complex/regular/conj.hpp>
+#include <eve/module/complex/regular/proj.hpp>
+
+namespace eve::detail
+{
+  EVE_FORCEINLINE auto complex_unary_dispatch(eve::tag::arg_, auto const& z) noexcept
```
DenisYaroshevskiy, line 18: Why unary/binary? I mean, we have perfect forwarding.

DenisYaroshevskiy: Do you have tests for functions that should work as default?
Like `eve::zip`?

include/eve/module/complex/regular/proj.hpp
```diff
@@ -0,0 +1,67 @@
...
+
+  EVE_MAKE_CALLABLE(proj_, proj);
+
+  namespace detail
+  {
+    template<floating_value V>
+    EVE_FORCEINLINE auto proj_(EVE_SUPPORTS(cpu_), V const & v) noexcept
+    {
+      using c_t = eve::as_complex_t<V>;
+      return if_else(is_infinite(v), c_t{inf(as(v)), zero(as(v))}, c_t(v,0));
```
DenisYaroshevskiy, line 64: This is_infine check seems useless. v, 0 should just work, no?

include/eve/module/complex/regular/traits.hpp
```diff
@@ -0,0 +1,58 @@
...
+
+  template<floating_scalar_value T> struct as_complex<complex<T>>
+  {
+    using type = eve::complex<T>;
+  };
+
+  template<floating_simd_value T> struct as_complex<T>
+  {
+    using type = eve::wide<eve::complex<element_type_t<T>>, cardinal_t<T>>;
+  };
```
DenisYaroshevskiy, line 31: Do we need these?

include/eve/module/core/regular/impl/diff_of_prod.hpp
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  template<floating_real_value T>
+  EVE_FORCEINLINE auto diff_of_prod_(EVE_SUPPORTS(cpu_)
+                                   , const T& a
+                                   , const T& b
+                                   , const T& c
+                                   , const T& d
+                                   ) noexcept
+  requires(has_native_abi_v<T>)
```
DenisYaroshevskiy, line 34: Why? Aggregated won't work?

include/eve/module/core/regular/impl/sum_of_prod.hpp
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  template<floating_real_value T>
+  EVE_FORCEINLINE auto sum_of_prod_(EVE_SUPPORTS(cpu_)
+                                   , const T& a
+                                   , const T& b
+                                   , const T& c
+                                   , const T& d
+                                   ) noexcept
+  requires(has_native_abi_v<T>)
```
DenisYaroshevskiy, line 34: Again, aggregated

include/eve/product_type.hpp
```diff
@@ -156,10 +157,19 @@ namespace eve
...
-    EVE_FORCEINLINE friend auto operator+(eve::like<Self> auto x, eve::like<Self> auto y) requires requires { x += y; }
+    template<like<Self> Z1, like<Self> Z2>
+    EVE_FORCEINLINE friend auto operator+(Z1 x, Z2 y) requires requires { x += x; }
     {
-      x += y;
-      return x;
+      if constexpr(scalar_value<Z1> && simd_value<Z2>)
+      {
+        as_wide_t<Z1> that{x};
+        return that += y;
```
DenisYaroshevskiy, line 166: Why not just y += x;

include/eve/module/complex/complex.hpp
jfalcou: so I don't have to do a if to know if I need to display 3 + 4i or 3 - 4i
I can put a if instead of I need to delve into ostream flag state conservation which is also a thing.

include/eve/module/complex/complex.hpp
jfalcou: I can't ask for real += complex so I ask for <real turned into proper wide> += lhs

include/eve/module/complex/complex.hpp
jfalcou: instead of lhs/rhs ?

include/eve/module/complex/regular/detail/arithmetic.hpp
jfalcou: 1./ I stumbled upon an ambiguous case for functions that may have 1 or 2 arguments AND supports decorator. SPlitting them this way was the easiest way I found to prevent those ambiguities. I can have another look if needed.

2./ Good point, will add one. I tested it locally but didn't write one.

include/eve/module/complex/regular/proj.hpp
jfalcou: It has to return +Infinite regardless of the sign of the inial value.

include/eve/module/complex/regular/traits.hpp
jfalcou: For a given generic type T that you want to turn into a complex, complex<T> can't work if T is wide<X,N>

include/eve/product_type.hpp
jfalcou: Good point

include/eve/module/core/regular/impl/sum_of_prod.hpp
jfalcou: Have to check, that's old jtlap's code

jfalcou: OK, the aggregated case is handled by the previous overload.
We don't want to call aggregated directly on multi-operators expression, so we call the arithmetic_call skeleton instead.
Now, we usually do that with a if constexpr in all the other code so it requires to be streamlined

include/eve/module/complex/complex.hpp
DenisYaroshevskiy: Wait, I suspect all of this is wrong. You will print (wide) (wide) and you are suppose to print by pairs I think.

include/eve/module/complex/regular/proj.hpp
jfalcou: From the std

> For most z, std::proj(z)==z, but all complex infinities, even the numbers where one component is infinite and the other is NaN, become positive real infinity.

include/eve/module/complex/complex.hpp
DenisYaroshevskiy: yep. I did that in a few places. You can stick with lhs, rhs if you want. Noone is ever consistent about this.

include/eve/product_type.hpp
jfalcou: Note that it will be needed for - and / that are no symmetric

include/eve/module/complex/complex.hpp
DenisYaroshevskiy: Ah, it's common type. Why not require rhs + lhs?

include/eve/module/core/regular/impl/sum_of_prod.hpp
jfalcou: Well, nope, the other overload takes care of both aggregation AND multi types variant.
So no this code is correct and it pass the aggregation tests already

include/eve/module/complex/regular/detail/arithmetic.hpp
DenisYaroshevskiy: 1. Ok - makes sense.

include/eve/module/core/regular/impl/sum_of_prod.hpp
DenisYaroshevskiy: Never used the "arithmetic_call" if it does what's needed, OK.
Removing overloads would be nice I think

jfalcou: The multi-type cases is needed to handle wide/scalar mixture.
I would say to let it this way for now

include/eve/module/complex/complex.hpp
jfalcou: Will do

include/eve/module/complex/complex.hpp
jfalcou: rhs can be `complex<T>` and lhs can be `wide<T>` so it won't work

DenisYaroshevskiy (approved):

Issue #1144, opened 2022-01-09
[PROJECT] Cleanup basic concepts & tests


DenisYaroshevskiy:
We need a good clean up of basic concepts. And have tests for subsumption.

# Phase 1
- [x] built_in_scalar_value (name is questionable - ints and floats)
- [x] arithmetic_scalar_value ( + product types)
- [x] predeclare wide <arithmetic_scalar_value>

# Phase 2
- [x] arithmetic_simd_value = instance_of_wide
- [x] builtin_simd_value = arithmetic_simd_value && builtin_scalar_value<element_type>
- [x] predeclare logical<built_in_scalar_value>
- [x] predeclare logical<wide<built_in_scalar_value>>
- [x] logical_scalar_value = instance_of_logical && the T inside is built_in_scalar_value
- [x] logical_simd_value   = instance_of_logical && the T inside is built_in_simd_value

# Phase 3

## Updated plan
By trying to remove the complicated and rather redundant `common_compatible` and `common_compatible_value`, @jtlap found out we can get rid of them with a simpler traits and couple of concepts. THe experience seems safer and more controlalble than just doing s/real//. 

 - [x] Use a single `common_value` traits
 - [x] real_* is in fact just requiring < to exist, adapt to make it obvious
 - [x] Remove all real_*
 - [x] Remove all real_*_value concepts as they are useless
 - [x] scalar_value = arithmetic_scalar_value || logical_scalar_value
 - [x] simd_value   = arithmetic_simd_value || logical_simd_value
 - [x] value = scalar_value || simd_value

jfalcou:
Note for later: we need to have product_type types somewhere so we can ask for SIMD product types proper.
This is a use case for complex for ex.

jfalcou:
Merged in 40e4615

PR #1306, opened 2022-06-02
#1144 - Part 1 : Basic concept and forward constraining wide

jfalcou:
The basic.hpp concept file defines : 
 - `plain_scalar` i.e float, ints etc
 -  `scalar_product_type ` which is a `kumi::product_type` of `plain_scalar` 
 - `scalar` which is `plain_scalar` or `scalar_product_type `

Added tests and made wide constrained on vectorizable.
Took the opportunity to make empty tuple non-vectorizable.


include/eve/concept/basic.hpp
```diff
@@ -0,0 +1,50 @@
...
+#include <type_traits>
+#include <cstdint>
+#include <cstddef>
+
+namespace eve
+{
+  template<typename T>
+  concept builtin_vectorizable  =     std::is_arithmetic_v<T>
+                                  && !std::is_same_v<T,bool>
+                                  &&  sizeof(T) <= 8;
```
DenisYaroshevskiy, line 20: I don't like the name tbh. 

Given that is_integral is already a list, do we maybe want to just list everything?

https://github.com/llvm/llvm-project/blob/50de7f1e7787bf327d1d04c4daab97c1c180b58c/libcxx/include/__type_traits/is_integral.h#L22

DenisYaroshevskiy (commented):
How about `plain_scalar` and `scalar`?

include/eve/concept/basic.hpp
jfalcou: I fear we may miss some std::intx_t on some architecture if we use a list

DenisYaroshevskiy: We should miss it. We don't support char for example. We have a very specific list.

jfalcou: OK

include/eve/concept/scalar.hpp
```diff
@@ -0,0 +1,52 @@
...
+
+#include <eve/detail/kumi.hpp>
+#include <type_traits>
+#include <cstdint>
+#include <cstddef>
+
+namespace eve
+{
+  template<typename T>
+  concept plain_scalar  = std::is_same_v<T,float>         ||  std::is_same_v<T,double>
```
DenisYaroshevskiy, line 18: a) we should have one_of concept
b) we should use std::same_as concept

include/eve/concept/scalar.hpp
```diff
@@ -0,0 +1,52 @@
...
+        {
+          using flatten_t = kumi::result::flatten_all_t<T>;
+          return []<std::size_t... I>( std::index_sequence<I...> )
+          {
+            return (plain_scalar<kumi::element_t<I,flatten_t>> && ... && true);
+          }(std::make_index_sequence<kumi::size<flatten_t>::value>{});
+        }
+        else
+        {
+          return false;
```
DenisYaroshevskiy, line 40: Is no elelements a problem

include/eve/concept/scalar.hpp
jfalcou: a) we have detaill:is_on_of<T>( types<U...>)  available

include/eve/concept/scalar.hpp
```diff
@@ -0,0 +1,56 @@
...
+#include <eve/detail/kumi.hpp>
+#include <eve/detail/meta.hpp>
+#include <type_traits>
+#include <cstdint>
+#include <cstddef>
+
+namespace eve
+{
+  template<typename T, typename... Ts>
+  concept one_of = (std::same_as<T,Ts> || ... || false);
```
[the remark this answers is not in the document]
jfalcou: it helps if Ts... is empty

DenisYaroshevskiy (approved):

PR #1307, opened 2022-06-03
Issue #1144 - Phase 2 : SIMD related concept + logical constraints

jfalcou:


include/eve/concept/simd.hpp
```diff
@@ -0,0 +1,40 @@
...
+  namespace detail
+  {
+    template<typename T> struct is_wide { static constexpr bool value = false; };
+
+    template<scalar T, typename N>
+    struct is_wide<wide<T,N>> { static constexpr bool value = true; };
+  }
+
+  template<typename T>
+  concept simd = detail::is_wide<T>::value;
```
DenisYaroshevskiy, line 30: Should we just move instance of to main detail? We have it in algo, we needed it for aligned pointer, now here?

DenisYaroshevskiy: Wait. Is logical<simd> - simd?

include/eve/concept/simd.hpp
```diff
@@ -0,0 +1,40 @@
...
+
+  template<typename T>
+  concept plain_simd = simd<T> && plain_scalar<element_type_t<T>>;
+
+  template<typename T>
+  concept product_simd  = simd<T> && kumi::product_type<element_type_t<T>>;
+
+  template<typename T>
+  concept plain_value = plain_scalar<T> || plain_simd<T>;
```
DenisYaroshevskiy, line 39: Do the other way around please:

```
concept plain_value = plain_scalar<element_type_t<T>>;
concept plain_simd..
```

DenisYaroshevskiy (commented):
I am not wild about "simd" as a concept name. 
Why not simd_value?

jfalcou:
> I am not wild about "simd" as a concept name. Why not simd_value?

As per #1144 discussion and decision, `simd_value` will be used for something else.

On e potential solution is to rename scalar -> vectorizable and simd -> vectorized AFTER we changed all the other concepts.
We can keep simd for now as a transition phase.

include/eve/concept/simd.hpp
jfalcou: Per #1144, logical<wide< is a simd_value but not simd
simd is basically the builtin stuff+product type as we decided in the #1144 discussion

include/eve/concept/simd.hpp
jfalcou: `plain_value` is either `plain_simd` or `plain_scalar` so I don't get what you mean

include/eve/concept/simd.hpp
jfalcou: I'll move instance_of into meta.hpp

include/eve/concept/simd.hpp
DenisYaroshevskiy: Hmm. I wanted `plain_value` to be `plain<element_type_t<T>` but this leads to issues.

ok fair enough.

DenisYaroshevskiy:
Discussion called `instance_of_wide` `arithmetic_simd_value`.
ANd it called `scalar_value` `arithmetic_scalar_value `.

We never agreed on `simd` concept that's for sure.

DenisYaroshevskiy (approved):

PR #1313, opened 2022-06-17
Update tests and infrastructure to use latest TTS to speedup compile times

jfalcou:


include/eve/module/core/regular/impl/simd/x86/slide_left.hpp
```diff
@@ -27,6 +27,15 @@ namespace eve::detail
...
         using i_t = as_integer_t<wide<T,N>, unsigned>;
+        using ec_t = expected_cardinal_t<T,abi_t<T, N>>;
+
+        if constexpr(N::value < 16)
+        {
+          v = bit_cast( bit_cast(v,as<wide<T,ec_t>>())
+                      & as_logical_t<wide<T,ec_t>>([](auto i, auto) { return i<N::value; }).mask()
+                      , as(v)
+                      );
+        }
```
DenisYaroshevskiy, line 38: This seems like a leftover

PR #1315, opened 2022-06-30
Better cleanup for functions sliding data in

### LINKED ISSUE #667: [FEATURE] clean up with known at compile time ignore
At the moment in a couple of places we clean up with the following code:

```
result &= keep_first(N::value-Shift).mask(as(result)).bits();
```

Since we know at compile time what we need to clear, we can do it more efficiently

jfalcou:
Some functions used a bad cleanup and slide_left somehow forgot to get cleaned up properly.
This fix #667 and clear up the issue raised by the latest tests rewrite on unary slide_left.


jfalcou:
@DenisYaroshevskiy current implementation generates a logical mask from `constexpr` values.
Even if the code generation looks OK. the alternative is to use `slide_right` which doesn't require additional constant generation.
However `slide_right` is in core, and `remove_garbage` lives in detail so I am a bit reluctant to use it. 

What's your opinion?

jfalcou:
> Owner



> @DenisYaroshevskiy current implementation generates a logical mask from `constexpr` values. Even if the code generation looks OK. the alternative is to use `slide_right` which doesn't require additional constant generation. However `slide_right` is in core, and `remove_garbage` lives in detail so I am a bit reluctant to use it.
> 
> What's your opinion?

After discussion, we can use slide_right

DenisYaroshevskiy (approved):

PR #1320, opened 2022-07-02
temptative clang format setup

jfalcou:


include/eve/concept/properly_convertible.hpp
```diff
@@ -19,29 +20,30 @@ namespace eve
   {
     // Note: this traits is customizable for other types we may need later on (complex,etc...)
     template<typename T, typename U>
-    struct  has_no_loss_convertion
-          : std::bool_constant<!(std::is_floating_point_v<T> && std::is_integral_v<U>)>
-    {};
+    struct has_no_loss_convertion
+        : std::bool_constant<!(std::is_floating_point_v<T> && std::is_integral_v<U>)>
+    {
+    };
```
DenisYaroshevskiy, line 26: Can we not do the 2 line braces? There should be an option.

include/eve/detail/function/simd/x86/friends.hpp
```diff
@@ -9,500 +9,485 @@
...
+  constexpr auto c = categorize<wide<T, N>>();
+  constexpr auto f = to_integer(cmp_flt::eq_oq);
 
-  //================================================================================================
-  template<real_value T, typename N>
-  EVE_FORCEINLINE as_logical_t<wide<T, N>> self_eq(wide<T,N> v, wide<T,N> w) noexcept
-      requires x86_abi<abi_t<T, N>>
+  if constexpr( current_api >= avx512 )
+  {
+    if constexpr( c == category::float32x16 ) return mask16 {_mm512_cmp_ps_mask(v, w, f)};
```
DenisYaroshevskiy, line 75: I think there was a setting to line up conditions, wasn't there?

include/eve/concept/properly_convertible.hpp
jfalcou: Fixed

include/eve/detail/function/simd/x86/friends.hpp
jfalcou: Can't find it. The one we waited for what that if constexpr was treated as If with respect to else.

include/eve/arch/cpu/wide.hpp
```diff
@@ -344,49 +340,56 @@ namespace eve
...
+    }
 
     //! Post-decrementation operator
-    EVE_FORCEINLINE wide operator--(int)  noexcept  { auto that(*this); operator--(); return that; }
+    EVE_FORCEINLINE wide operator--(int) noexcept
+    {
+      auto that(*this);
+           operator--();
+      return that;
+    }
```
DenisYaroshevskiy, line 372: What happened here? Is this by design?

DenisYaroshevskiy (commented):
I think ifs are less than ideal. But it's your call

DenisYaroshevskiy (approved):
I guess it's going to stay "ify". We will just have to deal.

include/eve/detail/function/simd/x86/friends.hpp
jfalcou: We will remove the else whenever we run over some to guarantee alignment

include/eve/arch/cpu/wide.hpp
jfalcou: Fixed

PR #1335, opened 2022-07-16
FIX-1334: transform_reduce

DenisYaroshevskiy:
One algorithm to be used as an alternative to `algo::reduce(views::map)`
Because it allows for non template mappin operation, had been done separately.


include/eve/algo/reduce.hpp
```diff
@@ -20,6 +20,40 @@
 
 namespace eve::algo
 {
+  //================================================================================================
+  //! @addtogroup algorithms
+  //! @{
+  //!  @var reduce
+  //!
+  //!  @brief SIMD version of std::reduce
```
jfalcou, line 28: I would appreciate if the new documentation for algo follows the schema from the copy.hpp file with respect to :
 - sub header
 - global location for information
 - use of doxygen @note or @warning accordingly

 It'll svae time later down the road.

include/eve/algo/transform_reduce.hpp
```diff
@@ -0,0 +1,142 @@
...
+#include <utility>
+
+namespace eve::algo
+{
+//================================================================================================
+//! @addtogroup algorithms
+//! @{
+//!  @var transform_reduce
+//!
+//!  @brief SIMD version of std::transform_reduce for a single range
```
jfalcou, line 28: Same than for reduce

test/unit/algo/algorithm/reduce_generic.h
```diff
@@ -0,0 +1,115 @@
+//==================================================================================================
```
jfalcou, line 1: Why not .hpp like usual ?

test/unit/algo/algorithm/reduce_generic.h
```diff
@@ -0,0 +1,115 @@
...
+  {
+    auto alg = basic_reduce[eve::algo::unroll<1>][eve::algo::force_cardinal<T::size()>];
+    reduce_generic_test_page_ends(tgt, e_t{10}, alg);
+  }
+
+  // unroll 3
+  {
+    reduce_generic_test_page_ends(native_tgt, e_t{10}, basic_reduce[eve::algo::unroll<3>]);
+  }
+}
```
jfalcou, line 115: Could we add test for stateful reduction operation ?

test/unit/algo/algorithm/sums_special_cases.cpp
```diff
@@ -69,21 +66,57 @@ TTS_CASE("eve.algo.inclusive_scan complex numbers")
...
   TTS_EQUAL(expected_real, real);
-  TTS_EQUAL(expected_img,  img);
+  TTS_EQUAL(expected_img, img);
 
   TTS_EQUAL(expected_real, real_copy);
-  TTS_EQUAL(expected_img,  img_copy);
+  TTS_EQUAL(expected_img, img_copy);
+};
+
+TTS_CASE("eve.algo.transform_reduce_simple")
```
jfalcou, line 88: or should the stateful reduction happen here?

test/doc/algo/reduce.cpp
```diff
@@ -1,5 +1,6 @@
 #include <eve/module/core.hpp>
 #include <eve/algo.hpp>
+#include <numeric>
 #include <iostream>
 #include <vector>
 #include "print.hpp"
```
jfalcou, line 6: due to how doc tests are able to be opened in Godbolt, we should not use print.hpp in those doc tests.
What I think is OK is to reproduce of the std cppreference samples, rise the size of the data and be happy.

Again, check how copy.cpp has been modified

test/doc/algo/transform_reduce.cpp
```diff
@@ -0,0 +1,26 @@
...
+#include <numeric>
+#include <vector>
+#include "print.hpp"
+
+int main()
+{
+  std::vector<float>  v = {1.0f, 2.0f, 3.0f, 4.0f};
+
+  std::cout << " -> v                                                                                               = ";
+  doc_utils::print(v);
```
jfalcou, line 13: Same. Avoid print

test/unit/algo/algorithm/sums_special_cases.cpp
DenisYaroshevskiy: I don't know a good example on stateful reduction. Stateful map added.

include/eve/algo/reduce.hpp
DenisYaroshevskiy: Please recheck.

test/unit/algo/algorithm/sums_special_cases.cpp
```diff
@@ -69,21 +66,91 @@ TTS_CASE("eve.algo.inclusive_scan complex numbers")
...
+
+TTS_CASE("eve.algo.reduce, stateful reduce")
+{
+  std::vector v {1, 2, 3};
+
+  TTS_EQUAL(6,
+            eve::algo::reduce(v,
+                              std::pair {[_ = 3](auto x, auto y)
+                                         {
+                                           (void)_;
```
jfalcou, line 132:
```
[[maybe_unused]]_;
```

maybe ?

DenisYaroshevskiy: doesn't work

jfalcou (approved):

PR #1336, opened 2022-07-21
Refactor EVE's exported CMake target and installation

justend29:
Hello!

While using EVE for the first time via CMake, I noticed a few points of friction. Instead of creating a GitHub issue, I thought it would be most beneficial to arrive with solutions instead of problems. As this is my first PR into EVE, I've only made a draft PR.

## The CMake issues were as follows
- EVE installs its CMake config-files into a non-default CMake path, directly in `/usr/lib`, preventing consuming libraries, including those installed by my system package manager, from finding it without explicitly specifying its path. CMake's search paths are specified in their documentation, [here](https://cmake.org/cmake/help/latest/command/find_package.html#config-mode-search-procedure)
- The exported target for eve, *eve_lib*, is different than that created in the build-tree, *eve::eve*.
- Furthermore, exported CMake targets are routinely exported with a namespace that ends in double colons (::). These are unacceptable in a path name, and will therefore limit CMake to finding CMake targets, with all of their interface requirements, instead of the compiled library or arbitrary library with the same name, *eve_lib* in this case.

Furthermore, I noticed a [comment in EVE's installation CMake module](https://github.com/jfalcou/eve/blob/main/cmake/config/eve-install.cmake#L21) that mentioned it was installing EVE to versioned directories. It wasn't actually doing this

## Changes
EVE....
- installs its config-file into a path within CMake's search path (specifically, a versioned path)
- exports a target named *eve::eve*, to align with the internal target and to retain the `::`.
- installs all files into versioned paths, allowing multiple simultaneously installed EVE versions
- has new usage documentation to reflect usage through CMake
    - the sample CMake configure+build commands were also updated to be build-system agnostic
    - documentation was built ([a6c3634](https://github.com/jfalcou/eve/pull/1336/commits/a6c36346cbb91d6281b75554f41b0ce8c3ed3b80) contains all of the doxygen artifacts and can therefore be easily removed if there's an official procedure for this)
- 's integration test has been updated for the new target & install
- installs its LICENSE.md (+1)
    
## Testing
I've tested installing EVE in both default and custom installation locations. In both of these, a separate project and the integration test within EVE were able to find, link, and use EVE with the CMake language commands added to the documentation.

I've also gone through the procedure of building and running the `unit.meta.exe` unit test in the CI docker container detailed in the [internal documentation](https://github.com/jfalcou/eve/tree/main/doc/internals). This was mostly a smoke test to ensure I didn't completely bork EVE. Running the full CI suite will be a better indication of this.


jfalcou:
Hi!

Thanks for this very thorough PR! We've been eye-balling our CMake install process for a while and I am very happy someone with proper skills took a look at it.

Some points to go forward:
 - we usually don't PR the HTML output of the doc until the very end. We also require to use trunk Doxygen cause some of our C++ needs a really recent Doxygen version. The best is to leave it out and we'll take care of updating the docs ourselves.
 - the interesting tests should be the ones in tests/integration as they "play along" faking an installation/fetch from GA+FetchContent etc... 
 
I think that once you clean up the docs/ part, we can run that as an actual PR and see how it goes.
I'll add more comments directly in the PR code.

DenisYaroshevskiy:
Awesome that you took the time, thanks. @jfalcou will do the review - I don't know any of this, just wanted to say thanks.

justend29:
@jfalcou @DenisYaroshevskiy Thanks for the quick look over!

As for the Doxygen documentation, I have removed all of the generated Doxygen artifacts from this PR and let you handle generation in the most appropriate manner.

I'll get back to the FetchContent and CPM integration tests after work.

jfalcou:
The CPM/FetchContent fails are probably due to the fact I try to fetch the <branch> from our repo. So whenever a PR from outside is made, it can't find the branch on jfalcou/eve. I probably need to grab the proper git ref for the PR and pull that.

justend29:
I've made the PR official so we can review the code in the meantime, and then we can address the CI checks after.

doc/page01_info.md
```diff
@@ -32,17 +32,19 @@ The following instructions are tentatively supported (ie code is incomplete and
 # Retrieving the source
 
 ## Github
+
 **EVE** is available on GitHub and can be retrieved via the following command:
 <br/>
 
 @verbatim
-$ git clone  https://github.com/jfalcou/eve.git
+$ git clone  <https://github.com/jfalcou/eve.git>
```
jfalcou, line 40: why the < > ? Is it stylistic ?

justend29: Oops. Definitely an accident. Too many keyboard shortcuts to accidentally hit. Will remove.

doc/page01_info.md
```diff
@@ -94,55 +99,60 @@ cmake
...
 Once **CMake** completes, you can use the `install` target to build and install **EVE**. By default,
-the library will be installed in the `/usr/local`directory, thus requiring root privileges on Linux.
+the library will be installed in the `/usr/local` directory, thus requiring root privileges on Linux.
 
-@verbatim
-$ ninja install
-@endverbatim
+```bash
+cmake --install build                # cmake >=3.15
+cmake --build build --target install # cmake <3.15
```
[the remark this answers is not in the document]
justend29: I was following EVE's [top-level CMake](https://github.com/jfalcou/eve/blob/main/CMakeLists.txt#L6), which indicated 3.2.
We can definitely simplify the documentation to just using the *--install* command, but I would recommend at least requiring 3.15, or 3.19, in the top level to avoid misuse.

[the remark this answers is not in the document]
jfalcou: Feel free to add it there. This 3.2 makes little sense.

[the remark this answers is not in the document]
justend29: Since CMake 3.19 commands are used elsewhere, I'll ensure to update the minimum required CMake version to 3.19 as part of this PR.

Just to be transparent, the *--install* command won't invoke the *install* target but will instead run the installation rules generated by the CMake configure step. There's nothing to build with this header-only library, so an intermediate *--build* command isn't needed. Building the *install* target builds everything in *all*, too. I believe that *--install* is preferred because it doesn't override the default install prefix, CMAKE_INSTALL_PREFIX, that CMake deduced through its system introspection, and allows the prefix changes w/o re-configuring + re-building.

The reason I'm mentioning this is that I will also slightly simplify the wording of the above text to no longer mention the *install* target or the building, and wanted to run this by you before I make another push and burn more of your CI credits. I can post a diff or just make the push, too.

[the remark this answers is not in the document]
jfalcou: Don't hesitate to fire anytime, the CI credit is not an issue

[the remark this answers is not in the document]
justend29: Okay. I increased the minimum required CMake version to 3.19 and updated docs to reflect those newer commands.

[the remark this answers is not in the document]
jfalcou: Guess I was wrong XD 3.18 it is. Sorry for the mishmash.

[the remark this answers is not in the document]
justend29: changed to 3.18. Any changes you'd like or can this conversation be resolved?

jfalcou:
So yeah the Fetch and CPM tests fail because we do stuff like

```
FetchContent_Declare(eve
  GIT_REPOSITORY https://github.com/jfalcou/eve.git
  GIT_TAG        ${EVE_SHA1}
  )
```

`EVE_SHA1` beign passed by the Github Action as '${{ github.head_ref }}'
So I guess we need to pass that + another macro to grab the actual repo the PR is coming from. 
`github.repositoryUrl` maybe but I fear it'll resolve as jfalcou/eve anyway.

doc/tutorial/mutli_arch.hpp
```diff
@@ -97,20 +100,21 @@ suffix. For each suffixes, the corresponding options will be used from [OPTIONS]
...
+cmake_minimum_required(VERSION 3.18)
 project(eve-multi-arch LANGUAGES CXX)
 
-find_package(eve)
+find_package(eve CONFIG REQIURED)
 
 add_library(setup INTERFACE)
-target_link_libraries(setup INTERFACE ${EVE_LIBRARIES})
+target_link_libraries(setup INTERFACE eve::eve)
 target_compile_options(setup INTERFACE "-O3")
```
justend29, line 118: @jfalcou In these cases, would it be better to let the user select the optimization level through cmake's `CMAKE_BUILD_TYPE`?

justend29:
@jfalcou I've rebased this branch on #1336 to combine the integration tests updates, specifically for the CI integration tests for this forked PR. Although those are now working, all of the other GitHub Actions CI checks are now failing due to a weird permission error within the CI environment.

Whenever you get a chance, could you take a peek at their details to see if you recognize the issue?

jfalcou:
> Whenever you happen to be free, could you please take a peek at their details to see if you recognize the issue?

Oh something probably stopped our runner. I'll check if they need to be rebooted or something. That's unrelated to what you did.

.github/actions/run_docker/Dockerfile
```diff
@@ -1,6 +1,7 @@
 # Container image that runs your code
 FROM jfalcou/compilers:v3
 
+
```
jfalcou, line 4: let's be pedantic for 0.5ns.
Can we remove that noisy commit? I guess this new line is accidental.

doc/page01_info.md
```diff
@@ -185,19 +191,37 @@ $ ./output
...
+configure and build your **CMake** project.
+
+```cmake
+add_executable(use-eve main.cpp)
+find_package(eve CONFIG REQUIRED)
+target_link_libraries(use-eve PRIVATE eve::eve)
+```
+
+> If a custom installation prefix was used, ensure your **EVE** installation is within **CMake's**
+  search path with the use of the **CMake** variables **eve_ROOT** (CMake 3.12), **eve_DIR**, or
```
jfalcou, line 207: Should we remove the 3.12 remark ?

doc/page01_info.md
```diff
@@ -185,19 +191,37 @@ $ ./output
...
 |--------|:-------------|:---------------------|
-| `EVE_BUILD_TEST`      | Enable unit tests for **EVE** (`ON` by default).       | `ninja unit`        |
-| `EVE_BUILD_BENCHMARKS`| Enable benchmark tests for **EVE** (`OFF` by default). | `ninja benchmarks`  |
-| `EVE_BUILD_RANDOM`    | Enable random tests for **EVE** (`OFF` by default).    | `ninja random`      |
-
-There is currently over 2000 tests, so compiling all unit tests may require a large machine or
-some time. We recommend compiling in parallel using `-j`.
+| `EVE_BUILD_TEST`      | Enable unit tests for **EVE** (`ON` by default).       | `unit`        |
+| `EVE_BUILD_BENCHMARKS`| Enable benchmark tests for **EVE** (`OFF` by default). | `benchmarks`  |
+| `EVE_BUILD_RANDOM`    | Enable random tests for **EVE** (`OFF` by default).    | `random`      |
```
jfalcou, line 219: Didn't we had a `EVE_BUILD_INTEGRATION` entry at some point or did I dreamt too deep ?

jfalcou (commented):
Very last comment before merging

.github/actions/run_docker/Dockerfile
justend29: Being pedantic is good.

I actually placed that line there purposefully to change a Dockerfile, causing rebuilding of the docker image. I did this when I was getting those mysterious CI errors. It didn't work.

I will clean up!

doc/page01_info.md
justend29: Since we are now saying 3.19 is the minimum, yes. Thanks!

doc/page01_info.md
justend29: Yes, that option was added as part of #1336 to enable the integration tests that are now registered with CTest. Thanks for catching my not updating this documentation section to match! I'll do that.

doc/tutorial/mutli_arch.hpp
jfalcou: Maybe, I don't know. The build_variant system is made so that you can apply an INTERFACE to your builds which may be more precise than a blanket set of flags. You tell me.

Now, this can be addressed in another PR maybe?

justend29: I'll just resolve this for now because it's only in a little tutorial and is very insignificant. We can open something later if it's worth it.

doc/page01_info.md
justend29: I added these docs and made an aggregated `integration` target to run the integration tests. This target was really just to have a target to put in the table, though ;)

jfalcou:
Thanks a lot for this amazing contribution!

DenisYaroshevskiy:
Good work, @justend29 
Thanks for taking the time

justend29:
Happy to help, guys! Thanks for the amazing project!

PR #1337, opened 2022-07-22
Fix #766 - Implement predicated min/max

### LINKED ISSUE #766: [FEATURE] making min/max callable with predicates
**Problem**

We need to write `min_element`, `max_element`.

For both of these we need both `min` (or `max) operation and `less` operation.
For something like `sort` we need all 3 `min`, `max` and `less`.

Even if we wait for compilers to optimise `if_else(less(y, x), y, x)` to `min` (they don't do it reliably enough) we still have a problem that `reduce(min)` should be specialised and it's hard to expect compilers to get all the way there.

The solution is:
move current min/max in detail.
make exposed min/max callables that can select a proper implementation. 

The `min_element` will look smth like this (except for traits and ranges and stuff):

```
template <eve::algo::iterator I, eve::algo::sentinel_for<I> S, typename P>
std::optional<value_type<I>> min_value(I f, S l, P p)
{
  if (f == l) return std::nullopt;
  return algo::reduce(f, l, min(p));
}

template <eve::algo::iterator I, eve::algo::sentinel_for<I> S, typename P>
I min_element(I f, S l, P p)
{
  std::optional v = min_value(f, l, p);
  if (!v) return f;
  return algo::find_if_not(f, l, [v, p](auto x) { p(v, x); });
}
```

** Details **

```
min(is_less) -> detail::min
min(is_greater) -> detail::max
min(a, b) -> min(eve::is_less)(a, b)
max(a, b) -> max(eve::is_less)(a, b)
```

jfalcou:
min and max can now be customized by a callable to change their predicate.
In this form, no other decorator nor masking can be applied.

### DIFF
--- a/include/eve/module/core/regular/impl/max.hpp
+++ b/include/eve/module/core/regular/impl/max.hpp
@@ -75,7 +75,7 @@ EVE_FORCEINLINE auto
 max_(EVE_SUPPORTS(cpu_), Callable f)
 {
   if constexpr( std::same_as<Callable, callable_is_less_> ) return eve::max;
-  else return [f](auto x, auto y) { return if_else(f(x, y), y, x); };
+  else return [f](auto x, auto y) { return if_else(f(y, x), x, y); };
 }
 
 }
--- a/include/eve/module/core/regular/impl/max.hpp
+++ b/include/eve/module/core/regular/impl/max.hpp
@@ -14,6 +14,7 @@
 #include <eve/module/core/regular/all.hpp>
 #include <eve/module/core/regular/if_else.hpp>
 #include <eve/module/core/regular/is_less.hpp>
+#include <eve/module/core/regular/is_greater.hpp>
 #include <eve/traits/common_compatible.hpp>
 
 namespace eve::detail
@@ -75,6 +76,7 @@ EVE_FORCEINLINE auto
 max_(EVE_SUPPORTS(cpu_), Callable f)
 {
   if constexpr( std::same_as<Callable, callable_is_less_> ) return eve::max;
+  else if constexpr( std::same_as<Callable, callable_is_greater_> ) return eve::min;
   else return [f](auto x, auto y) { return if_else(f(y, x), x, y); };
 }
 
--- a/include/eve/module/core/regular/impl/max.hpp
+++ b/include/eve/module/core/regular/impl/max.hpp
@@ -7,63 +7,75 @@
 //==================================================================================================
 #pragma once
 
-#include <eve/module/core/regular/if_else.hpp>
-#include <eve/module/core/regular/all.hpp>
-#include <eve/detail/implementation.hpp>
-#include <eve/module/core/regular/is_less.hpp>
-#include <eve/concept/value.hpp>
 #include <eve/concept/compatible.hpp>
+#include <eve/concept/value.hpp>
 #include <eve/detail/apply_over.hpp>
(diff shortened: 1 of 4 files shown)


include/eve/module/core/regular/impl/max.hpp
```diff
@@ -7,63 +7,75 @@
...
+max_(EVE_SUPPORTS(cpu_), T const& a, U const& b) noexcept requires compatible_values<T, U>
+{
+  return arithmetic_call(max, a, b);
+}
+
+template<real_scalar_value T>
+EVE_FORCEINLINE auto
+max_(EVE_SUPPORTS(cpu_), T const& a, T const& b) noexcept
+{
+  return a < b ? b : a;
```
DenisYaroshevskiy, line 34: If equal - return b.
So `b < a ? a : b`

include/eve/module/core/regular/impl/max.hpp
```diff
@@ -7,63 +7,75 @@
...
-  }
+//================================================================================================
+// Predicate case
+//================================================================================================
+template<typename Callable>
+EVE_FORCEINLINE auto
+max_(EVE_SUPPORTS(cpu_), Callable f)
+{
+  if constexpr( std::same_as<Callable, callable_is_less_> ) return eve::max;
+  else return [f](auto x, auto y) { return if_else(f(x, y), y, x); };
```
DenisYaroshevskiy, line 78: I think `max(is_greater)` => `min` is correct.

include/eve/module/core/regular/impl/min.hpp
```diff
@@ -65,4 +66,16 @@ min_(EVE_SUPPORTS(cpu_), T0 a0, T1 a1, Ts... args)
...
+
+//================================================================================================
+// Predicate case
+//================================================================================================
+template<typename Callable>
+EVE_FORCEINLINE auto
+min_(EVE_SUPPORTS(cpu_), Callable f)
+{
+  if constexpr( std::same_as<Callable, callable_is_less_> ) return eve::min;
+  else return [f](auto x, auto y) { return if_else(f(y, x), y, x); };
```
DenisYaroshevskiy, line 78: I think min(is_greater) returning max is correct.

include/eve/module/core/regular/impl/max.hpp
```diff
@@ -7,63 +7,75 @@
...
```
DenisYaroshevskiy, line 78: Same as b < a, a, b. PLease add a test.

include/eve/module/core/regular/impl/min.hpp
```diff
@@ -30,7 +31,7 @@ template<real_scalar_value T>
 EVE_FORCEINLINE auto
 min_(EVE_SUPPORTS(cpu_), T const& a, T const& b) noexcept
 {
-  return if_else(b < a, b, a);
+  return b < a ? b : a;
```
DenisYaroshevskiy, line 34: Wait. How is this instead of if_else

jfalcou: This is for scalar.

include/eve/module/core/regular/impl/max.hpp
jfalcou: Ok

jfalcou: Done

DenisYaroshevskiy (approved):

PR #1338, opened 2022-07-22
Automate integration tests and correct their fetches.

justend29:
## Issue
Hard-coded paths to `jfalcou/eve` within EVE's [CPM](https://github.com/jfalcou/eve/blob/main/test/integration/cpm-test/CMakeLists.txt#L16) and [FetchContent](https://github.com/jfalcou/eve/blob/main/test/integration/fetch-test/CMakeLists.txt#L16) integration tests cause these tests to always fail for `pull_request` workflows from third-party repositories, i.e. project forks. This issue was found in [this PR regarding CMake installation updates](https://github.com/jfalcou/eve/pull/1336#issuecomment-1192845010).

Instead, these integration tests should fetch the merged results of the HEAD and the target branch.

## Background

GitHub Actions already evaluates the `pull_request` workflows on the merged results of the HEAD and target branch. This is done by creating a commit of the merged results onto an intermediate, temporary branch. Then each *checkout* action in the CI workflow checks out this commit. This [Stack Overflow response](
https://stackoverflow.com/questions/57683943/github-actions-how-to-build-a-pull-request-as-if-it-were-merged) and some additional testing with the [GitHub Environment Variables](https://docs.github.com/en/actions/learn-github-actions/environment-variables) proved helpful in this

## Approach

Rather than instructing these integration tests to redundantly fetch the merged code again, CPM and FetchContent instead reference the **EVE** root project directory. With this, they always use whatever code exists in the currently checked-out git working tree.

## Changes

- These integration tests now reference the CMake variable, `eve_SOURCE_DIR`, instead of the `jfalcou/eve` 
- All of the integration tests were also registered with CTest, allowing them to be directly run as the other automated tests. In an attempt to follow project conventions, these new tests are:
  - **integration.install-eve.exe** - installs the EVE project to a local installation prefix. This tests the installation procedure and allows future tests to find it with `find_package()`
  - **integration.install.eve** - configures, builds, and tests `test/integration/install-test`, finding the result from the above installation.
  - **integration.cpm.exe** - configures, builds, and tests `test/integration/cpm-test`, fetching the current working tree
  - **integration.fetch.exe** - configures, builds, and tests `test/integration/fetch-test`, also fetching the current working tree
- the executables built within each of these subprojects, all called *test_eve*, are now registered with CTest such that running the new integration tests automatically runs their simple test program. The return code of these test programs will validate or invalidate their enclosing integration test, thereby supporting more rigorous internal assertions.
- A new CMake option, `EVE_BUILD_INTEGRATION`, is used to enable or disable the configuration of these integration tests independent of `EVE_BUILD_TEST`.
- EVE's CI entrypoint for the integration tests was simplified to reflect the more automated nature of these integration tests.

## Future Work

Once this PR is merged, I will rebase #1336, then we can merge it.


justend29:
@jfalcou I'm assuming the workflows awaiting approval are those handled by GitHub Actions; those which contain the pertinent integration tests. Whenever you get a chance, could you please enable this to continue with the tests? Thanks :)

jfalcou:
You need to make it a PR. Drafts are not run.

justend29:
Yes, the GitHub Actions are being triggered. Thank you.

cmake/config/options.cmake
```diff
@@ -10,37 +10,45 @@
 option( EVE_USE_PCH             "Build unit tests with precompield headers"     ON  )
 option( EVE_BUILD_TEST          "Build unit tests for eve"                      ON  )
 option( EVE_BUILD_RANDOM        "Build random precision tests for eve"          OFF )
+option( EVE_BUILD_INTEGRATION   "Build CMake integration tests for eve"         OFF )
 option( EVE_BUILD_BENCHMARKS    "Build benchmarks for eve"                      OFF )
 option( EVE_BUILD_DOCUMENTATION "Compile Doxygen documentation"                 ON  )
 
 if( EVE_USE_PCH )
-  set(pch_status "ON")
+  set(pch_status "ON ")
```
jfalcou, line 18: why the space ?

test/integration/cpm-test/CMakeLists.txt
```diff
@@ -4,17 +4,18 @@
 ##  SPDX-License-Identifier: MIT
 ##==================================================================================================
 cmake_minimum_required(VERSION 3.19)
```
jfalcou, line 6: should we put the same requirements as the main CMakeLists.txt?

jfalcou (commented):
Wonderful patch 👍

cmake/config/options.cmake
justend29: The word *OFF* has three characters, so the additional space on "ON " aligns the leading parenthesis on the following console messages.

justend29:
> Wonderful patch +1

I'm glad you like it! I just updated the main PR comment to provide context and explanation for some of the changes.

test/integration/cpm-test/CMakeLists.txt
justend29: Done.

justend29: @jfalcou It looks like the CMake version on some of the CI environments is 3.18.4, which are now failing as a result of this update.

I can look into what version is absolutely mandatory. Maybe we can get away with 3.18.4 - maybe not.
 We could also just upgrade the CMake version. Thoughts?

jfalcou: I was more thinking setting up everybody to 3.18 as we established I the other PR

justend29: Oh, I understand what you were indicating. Will do.

cmake/config/options.cmake
jfalcou: wont it affect the way we test for such option or did I accindentally wrote the correct thing ?

justend29: They didn't appear to be used anywhere outside of `options.cmake`.

jfalcou: Yeah noticed right after.

PR #1340, opened 2022-07-23
FIX-1339: last_true

DenisYaroshevskiy:


test/unit/module/core/first_true.cpp
```diff
@@ -103,7 +103,7 @@ TTS_CASE_TPL("Check eve::first_true", eve::test::simd::all_types)
     }
 
     x = true;
-    TTS_EQUAL(eve::first_true[eve::ignore_first(T::size())](x), std::nullopt);
+    TTS_EQUAL(eve::first_true[eve::ignore_last(T::size())](x), std::nullopt);
```
jfalcou, line 106: why the change ? Was the test wrong since before, why didn't we found out ?

DenisYaroshevskiy: It was created as a copy paste. This one I originally forgot to change

include/eve/module/core/regular/impl/last_true.hpp
```diff
@@ -0,0 +1,106 @@
...
+      return {};
+    }
+    // This is pretty good for aggreagted as well.
+    else if constexpr ( !top_bits<T>::is_cheap )
+    {
+      // No ignore, we might luck out even if some elements should not be counted.
+      if ( !eve::any(v) ) return {};
+
+      top_bits mmask{v, cond};
+      if constexpr ( C::is_complete ) return *last_true(mmask); // optimizes a check on clang
```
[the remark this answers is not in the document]
DenisYaroshevskiy: The !0 check. We already know it's !0 if the condition is complete.

*first_true will be ub in case of 0 check and clang can remove it

jfalcou (approved):

PR #1341, opened 2022-07-23
Basic setup and start of SVE support

jfalcou:
+ Setup CI for SVE using cross-compilers
 + Basic setup

This is to be completed.


include/eve/arch/arm/tags.hpp
```diff
@@ -33,23 +33,37 @@ namespace eve
...
 
   //================================================================================================
   // ARM ABI concept
   //================================================================================================
-  template<typename T> concept arm_abi = detail::is_one_of<T>(detail::types<arm_64_,arm_128_> {});
+  template<typename T> concept arm_abi = detail::is_one_of<T> ( detail::types<arm_64_,arm_128_> {});
+  template<typename T> concept sve_abi = detail::is_one_of<T> ( detail::types < arm_sve_128_
+                                                                              , arm_sve_256_
+                                                                              , arm_sve_512_
+                                                                              > {}
```
DenisYaroshevskiy, line 67: Wait - arm_abi should include sve. Same as x86 includes 512. Am I wrong?

cmake/toolchain/gcc.sve512.cmake
```diff
@@ -0,0 +1,13 @@
...
+##  SPDX-License-Identifier: MIT
+##==================================================================================================
+set(CMAKE_SYSTEM_NAME Linux     )
+set(CMAKE_SYSTEM_PROCESSOR arm  )
+
+set(CMAKE_C_COMPILER    aarch64-linux-gnu-gcc-11  )
+set(CMAKE_CXX_COMPILER  aarch64-linux-gnu-g++-11  )
+set(CMAKE_CXX_FLAGS     "-Wno-psabi -DEVE_NO_FORCEINLINE ${EVE_OPTIONS} -march=armv8-a+sve -msve-vector-bits=512" )
+
+set(CMAKE_CROSSCOMPILING_CMD qemu-aarch64)
```
DenisYaroshevskiy, line 13: All 3 running with the same emulator is weird - how does qemu know the instruction width.

DenisYaroshevskiy (commented):
I am not sure I understand the rest but we will figure it out.

cmake/toolchain/gcc.sve512.cmake
jfalcou: THis needs investigation yes

include/eve/arch/arm/tags.hpp
jfalcou: Considering the register storage is different from ARM NEON and the intrinsics works in a different way, I wanted to split it up. We can discuss that.

DenisYaroshevskiy (approved):

include/eve/arch/arm/as_register.hpp
```diff
@@ -17,7 +17,99 @@ namespace eve
...
+      }
+    }
+
+    using type = decltype(find());
+    static_assert( !std::is_void_v<type>, "[eve arm sve] - Type is not usable in a SIMD register");
+  };
+}
+#endif
+
+#if defined(EVE_HW_ARM) || defined(EVE_HW_ARM_SVE)
```
DenisYaroshevskiy, line 112: Wait, this is confusing.

I thought we decided either arm or arm sve.

Why is this an if?

include/eve/arch/arm/tags.hpp
```diff
@@ -33,23 +33,37 @@ namespace eve
...
-  struct neon128_ : simd_api<simd_   , spy::neon_>  {};
-  struct asimd_   : simd_api<neon128_, spy::asimd_> {};
+  struct neon128_ : simd_api<simd_   , spy::neon_>      {};
+  struct asimd_   : simd_api<neon128_, spy::asimd_>     {};
+  struct sve128_  : simd_api<simd_   , spy::fixed_sve_> {};
+  struct sve256_  : simd_api<simd_   , spy::fixed_sve_> {};
+  struct sve512_  : simd_api<simd_   , spy::fixed_sve_> {};
 
   //================================================================================================
   // NEON extension tag objects
```
DenisYaroshevskiy, line 52: Not neon

DenisYaroshevskiy (commented):
Let's figure out the mess with neon/sve a bit.

a) let's split neon/sve where there are not split
b) how does one opt out of sve if they want to just use neon?

include/eve/arch/arm/sve/tags.hpp
```diff
@@ -0,0 +1,61 @@
...
+#include <eve/arch/arm/predef.hpp>
+#include <eve/arch/cpu/tags.hpp>
+#include <eve/detail/meta.hpp>
+
+namespace eve
+{
+  //================================================================================================
+  // ABI tags for all ARM bits SIMD registers
+  //================================================================================================
+  template<std::size_t Size, bool Logical> struct sve_abi_
```
DenisYaroshevskiy, line 19: probably bool is not really needed?

include/eve/arch/cpu/wide.hpp
```diff
@@ -404,7 +375,6 @@ namespace eve
 #if !defined(EVE_DOXYGEN_INVOKED)
```
DenisYaroshevskiy, line 375: Probably should be removed here too.

DenisYaroshevskiy (approved):

PR #1347, opened 2022-07-24
find backward

DenisYaroshevskiy:


test/doc/algo/find.cpp
```diff
@@ -6,20 +6,31 @@
 
 int main()
 {
```
jfalcou, line 8: I would prefer to have a separate .cpp for backward as if not, the doc is a bit loong to read.

jfalcou (approved):

PR #1348, opened 2022-07-24
Setup for WASM

jfalcou:


.github/actions/run_docker/entrypoint.sh
```diff
@@ -47,7 +49,10 @@ test_targets()
   return 0;
 }
 
-set -e
+# Temporary fix for ARM runtime
+rm /lib/ld-linux-armhf.so.3
+ln -s /usr/arm-linux-gnueabihf/lib/ld-linux-armhf.so.3 /lib/ld-linux-armhf.so.3
```
DenisYaroshevskiy, line 54: What is happening and how is WASM related to arm runtime?

.github/workflows/main.yml
```diff
@@ -42,6 +42,7 @@ jobs:
         - { comp: gcc         , arch: sve128  , opts: -Wno-psabi, status: 0 }
         - { comp: gcc         , arch: sve256  , opts: -Wno-psabi, status: 0 }
         - { comp: gcc         , arch: sve512  , opts: -Wno-psabi, status: 0 }
+        - { comp: clang       , arch: wasm    , opts: -Wno-psabi, status: 0 }
```
DenisYaroshevskiy, line 45: Does wasm also needs -Wno-psabi

DenisYaroshevskiy (commented):
are you adding the registers in this pr? Or is that a separate pr?

.github/actions/run_docker/entrypoint.sh
jfalcou: I had to move to the v4 of our Docker. It appears to have a bug on how the ARM 32bits works now and need to fix it.
To not have spurrious fail, I added that. This is totally temporary.

.github/workflows/main.yml
jfalcou: good catch

jfalcou:
> are you adding the registers in this pr? Or is that a separate pr?

In this PR yes, will do this afternoon

include/eve/concept/scalar.hpp
```diff
@@ -28,7 +28,9 @@ concept plain_scalar_value = detail::one_of<T,
                                             std::uint8_t,
                                             std::uint16_t,
                                             std::uint32_t,
-                                            std::uint64_t>;
+                                            std::uint64_t,
+                                            std::size_t,
+                                            std::ptrdiff_t>;
```
DenisYaroshevskiy, line 33: I think this is the wrong decision here, they are not plain_scalar. Didn't we discuss an alternative approach?

include/eve/module/math/regular/impl/geommean.hpp
```diff
@@ -60,6 +59,6 @@ namespace eve::detail
 
     };
     ((that = next(that,args)),...);
-    return if_else(is_even(sizeof...(args)) && is_ltz(sgn),  eve::allbits, sgn*that);
+    return if_else((sizeof...(args) % 2 == 0) && is_ltz(sgn),  eve::allbits, sgn*that);
```
DenisYaroshevskiy, line 62: What's wrong with is_even btw?

include/eve/concept/scalar.hpp
jfalcou: Yes but it's elsewhere. SO I'll probably extract this exact fix and put it there.

include/eve/module/math/regular/impl/geommean.hpp
jfalcou: I'm investigating it, I fixed it this way so we can check the rest of the behaviour

include/eve/arch/abi_of.hpp
```diff
@@ -7,85 +7,84 @@
...
+      {
+        if constexpr( !f64 && width <= 16 ) return ppc_ {};
+        else return emulated_ {};
+      }
+      else if constexpr( spy::simd_instruction_set == spy::fixed_sve_ )
+      {
+        if constexpr( spy::simd_instruction_set.width == 128 ) return arm_sve_128_ {};
+        else if constexpr( spy::simd_instruction_set.width == 256 ) return arm_sve_256_ {};
+        else if constexpr( spy::simd_instruction_set.width == 512 ) return arm_sve_512_ {};
+        else return emulated_ {};
```
DenisYaroshevskiy, line 56: Ok but if we are not supporting bigger sve for now, shouldn't we go to neon here?

include/eve/arch/wasm/as_register.hpp
```diff
@@ -0,0 +1,48 @@
...
+template<typename Type, typename Size> struct as_register<Type, Size, eve::wasm_128_>
+{
+  static constexpr auto find()
+  {
+    constexpr auto width = sizeof(Type) * Size::value;
+    if constexpr( width <= 16 )
+    {
+      if constexpr( std::is_same_v<Type, double> ) return v128_t {};
+      else if constexpr( std::is_same_v<Type, float> ) return v128_t {};
+      else if constexpr( std::is_integral_v<Type> ) return v128_t {};
```
DenisYaroshevskiy, line 34: Why if constexpr if  the result is always the same? To do the assert? Maybe then just use the `plain` concept?

include/eve/arch/wasm/tags.hpp
```diff
@@ -0,0 +1,53 @@
...
+};
+
+struct wasm_128_ : wasm_abi_<128, true>
+{};
+
+//================================================================================================
+// Dispatching tag for WASM SIMD implementation
+//================================================================================================
+struct simd128_ : simd_api<simd_, spy::simd128_>
+{};
```
DenisYaroshevskiy, line 41: I am confused as to what this is.

include/eve/concept/scalar.hpp
DenisYaroshevskiy: Is it staying? I'd like it to go away if possible.

include/eve/detail/function/simd/wasm/arithmetic_compounds.hpp
```diff
@@ -0,0 +1,136 @@
...
+    constexpr auto c = categorize<type>();
+
+    if constexpr( c == category::float64x2 ) self = wasm_f64x2_div(self, other);
+    else if constexpr( c == category::float32x4 ) self = wasm_f32x4_div(self, other);
+    else
+    {
+      auto           s    = self;
+      constexpr auto sdiv = [](auto a, auto b) { return a /= b; };
+      self                = map(sdiv, s, other);
+    }
```
DenisYaroshevskiy, line 131: There is no integer division??? O_o
Yeah seems like you right.

include/eve/detail/function/simd/wasm/bit_compounds.hpp
```diff
@@ -0,0 +1,123 @@
...
+  using enum category;
+
+  if constexpr( c == int64x2 ) v = wasm_i64x2_shr(v, s);
+  if constexpr( c == int32x4 ) v = wasm_i32x4_shr(v, s);
+  if constexpr( c == int16x8 ) v = wasm_u16x8_shr(v, s);
+  if constexpr( c == int8x16 ) v = wasm_i8x16_shr(v, s);
+  if constexpr( c == uint64x2 ) v = wasm_i64x2_shr(v, s);
+  if constexpr( c == uint32x4 ) v = wasm_i32x4_shr(v, s);
+  if constexpr( c == uint16x8 ) v = wasm_i16x8_shr(v, s);
+  if constexpr( c == uint8x16 ) v = wasm_i8x16_shr(v, s);
```
DenisYaroshevskiy, line 57: Eh... Are you sure. This is identical to shr

include/eve/detail/function/simd/wasm/bit_compounds.hpp
```diff
@@ -0,0 +1,123 @@
...
+template<scalar_value T, value U, typename N>
+    EVE_FORCEINLINE decltype(auto)
+    self_bitxor(wide<T, N>& self, U const& other) noexcept
+    requires((sizeof(wide<T, N>) == sizeof(U)) || (sizeof(T) == sizeof(U)))
+    && wasm_abi<abi_t<T, N>>
+{
+  if constexpr( scalar_value<U> )
+  {
+    using type = wide<T, N>;
+    return self ^= type {bit_cast(other, as<T> {})};
```
DenisYaroshevskiy, line 115: What's happening? Is it like ^ between an int and a float? Is that a thing?

include/eve/detail/function/simd/wasm/bit_compounds.hpp
DenisYaroshevskiy: No elses. I don't mind but we do elses in other places

include/eve/detail/function/simd/wasm/friends.hpp
```diff
@@ -0,0 +1,165 @@
...
+{
+  return bit_cast(~v.bits(), as(v));
+}
+
+//==================================================================================================
+template<real_value T, typename N>
+EVE_FORCEINLINE auto
+self_eq(wide<T, N> v, wide<T, N> w) noexcept requires wasm_abi<abi_t<T, N>>
+{
+  using type       = as_logical_t<wide<T, N>>;
```
DenisYaroshevskiy, line 31: Or you just write this as a return type and code shrinks

include/eve/detail/function/simd/wasm/friends.hpp
```diff
@@ -0,0 +1,165 @@
...
+{
+  return bit_cast(v.bits() == w.bits(), as(v));
+}
+
+//==================================================================================================
+template<real_value T, typename N>
+EVE_FORCEINLINE auto
+self_neq(wide<T, N> v, wide<T, N> w) noexcept requires wasm_abi<abi_t<T, N>>
+{
+  using type       = as_logical_t<wide<T, N>>;
```
DenisYaroshevskiy, line 59: Same here.

include/eve/detail/function/simd/wasm/slice.hpp
```diff
@@ -0,0 +1,57 @@
...
+  template<typename T, typename N, typename Slice>
+  EVE_FORCEINLINE wide<T, typename N::split_type>
+  slice(wide<T,N> const &a, Slice const &) noexcept
+      requires wasm_abi<abi_t<T, N>>
+  {
+    using sub_t = as_wide_t<T, typename N::split_type>;
+
+    auto eval = [&](auto... I) { return sub_t {a.get(I + (Slice::value * sub_t::size()))...}; };
+    return apply<N::value / 2>(eval);
+  }
```
DenisYaroshevskiy, line 27: This is not CPU? Why not?

include/eve/detail/function/simd/wasm/slice.hpp
```diff
@@ -0,0 +1,57 @@
...
+
+  template<typename T, typename N, typename Slice>
+  EVE_FORCEINLINE logical<wide<T, typename N::split_type>>
+  slice(logical<wide<T,N>> const &a, Slice const &) noexcept
+      requires wasm_abi<abi_t<T, N>>
+  {
+    using l_t = logical<wide<T, typename N::split_type>>;
+    using s_t = typename l_t::storage_type;
+    return l_t( bit_cast(a.mask().slice(Slice{}).storage(), as<s_t>()) );
+  }
```
DenisYaroshevskiy, line 37: Same here - seems like CPU.

include/eve/detail/function/simd/wasm/slice.hpp
```diff
@@ -0,0 +1,57 @@
...
+    return that;
+  }
+
+  template<typename T, typename N>
+  EVE_FORCEINLINE auto slice(logical<wide<T,N>> const &a) noexcept
+      requires wasm_abi<abi_t<T, N>>
+  {
+    std::array<logical<wide<T, typename N::split_type>>,2> that{slice(a, lower_), slice(a, upper_)};
+    return that;
+  }
```
DenisYaroshevskiy, line 56: CPU?

cmake/toolchain/sve128.json
```diff
@@ -0,0 +1,3 @@
+{
+  "unit.meta.exe":      "^unit.meta\\..*\\.exe"
```
DenisYaroshevskiy, line 2: I think it's reasonable to expect that we will have one sve config.

include/eve/detail/function/simd/wasm/make.hpp
```diff
@@ -0,0 +1,120 @@
...
+  if constexpr( S::value == 2 )
+  {
+    if constexpr( c == float64x2 ) return wasm_f64x2_make(vs...);
+    if constexpr( c == int64x2 ) return wasm_i64x2_make(vs...);
+    if constexpr( c == uint64x2 ) return wasm_u64x2_make(vs...);
+    if constexpr( c == float32x4 ) return wasm_f32x4_make(vs..., 0.f, 0.f);
+    if constexpr( c == int32x4 ) return wasm_i32x4_make(vs..., 0, 0);
+    if constexpr( c == uint32x4 ) return wasm_u32x4_make(vs..., 0, 0);
+    if constexpr( c == int16x8 ) return wasm_i16x8_make(vs..., 0, 0, 0, 0, 0, 0);
+    if constexpr( c == uint16x8 ) return wasm_u16x8_make(vs..., 0, 0, 0, 0, 0, 0);
```
DenisYaroshevskiy, line 40: if you wrap these `u16x8_make` into

```
namespace wsm {
inline auto wasm_u32x4_make(std::uint32_t v0,  std::uint32_t v1, std::uint32_t v2, std::uint32_t v3) {
   return wasm_u32x4_make(v0, v1, v2, v3);
} 
}
```

You will be able to use variadics here

include/eve/detail/function/simd/wasm/bit_compounds.hpp
jfalcou: we support bitwise between all scalar since day 1 as it is a basic SIMD optimisation. all bit_* operations treat their second aprameter as a mere bit mask w/r to its type.

cmake/toolchain/sve128.json
jfalcou: True

PR #1352, opened 2022-07-25
try to build a few things sve

DenisYaroshevskiy:
This compiled more things then compiled now. Not that a single new test can be enabled, mind you.
I don't know if we want to merge or drop this - I am very much struggling with new ideas here.


DenisYaroshevskiy:
I will withdraw this one and try again from scratch later

PR #1354, opened 2022-07-26
pedantic mul and div for complex

jtlap:
changes the behaviour of sum/diff_of_prod according the future doc


include/eve/module/complex/complex.hpp
```diff
@@ -251,6 +251,59 @@ namespace eve
       return y * x;
     }
 
+    template<like<complex> Z1, like<complex> Z2>
+    friend auto
+    tagged_dispatch(eve::tag::mul_, pedantic_type const &, Z1 const& z1, Z2  const& z2) noexcept
```
DenisYaroshevskiy, line 256: Are you sure that pedantic mul should me in the main file?

include/eve/module/complex/complex.hpp
```diff
@@ -214,9 +214,9 @@ namespace eve
     {
       auto [a, b] = self;
       auto [c, d] = o;
-      real(self) = diff_of_prod(a, c, b, d);
-      imag(self) = sum_of_prod(a, d, b, c);
-      return self;
+      real(self) = a*c-b*d;
+      imag(self) = a*d+b*c;
```
DenisYaroshevskiy, line 218: Curious: what happened with those functions?

DenisYaroshevskiy (commented):
drive by

include/eve/module/complex/complex.hpp
jtlap: Have you another proposal ?

include/eve/module/complex/complex.hpp
jtlap: in fact the purpose of all that is to change the behaviour of sum_of_prod and diff_of_prod to always call fma functions and not the naive calls. This is done.
But it implies that the compl test/unit/module/complex/mul.cpp
ex multiplication is in all case 2 times (at least) slower than the naive call.
So the naive call is used in complex multiplication and sum of prod diff of prod in pedantic.
The use of these functions implies  slower execution but also better accuracy. The part added after is any(is_not_finite...
are to provide some amelioration in corner cases as (1+inf i)*(1+0i)

include/eve/module/complex/complex.hpp
jtlap: up to now, in complex the decorated version of the functions are in the same file as the functions, but I can put them in details arithmetic. Not a great difference.
Tell me your pro and con

DenisYaroshevskiy: I think maybe our design principles changed here maybe.
Originally we wanted to do "include what you use" model, so that user can save compile time.
Meaning, in this case  the would need to include the header for pedantic multiplication.

Now we have these big headers that we want users to include, assuming that modules will be here soon. So now this is less impactful.

never the less, I'd probably go like we do everywhere else:

`regular/mul` `pedantic/mul` and then include one of both of them, something like that.

include/eve/module/complex/complex.hpp
DenisYaroshevskiy: I didn't get all of that but:

Does this mean that `sum_of_prod` and `diff_of_prod` are pedantic by default?
This seems weird, since most things we have a fast by default, no?

As a user I'd expect that I should always call a function with a proper name, not that I shlould do it by hand in some cases.

DenisYaroshevskiy: Basically, if `diff_of_prod` is worse than `a*c-b*d` - we have a problem.

jtlap: the problem with diff_of_prod is that contrarily to fma if fma is not available it is lot worse than ab-cd. 
If hardware fma is there the worse factor is only 2 and it presents accuracy advantage (no more than 3 ulps from correct roundng)
So diff_of prod must be used only if needed. 
With the solution implemented there, up to now it is  used once or twice in huge special functions. And no more in complex  
The new version of the documentation will put its description  in a section "accuracy helpers" that will also contains functions as two_prod and two_add that are meant to help emulate fma when the hardare lacks.

jtlap: accuracy and speed are two separate but related problems.
Perhaps a compilation option style -DACCURATE_MATH could be used, but it seems more complicated.

DenisYaroshevskiy: No - the switches are not the solution I think.

I see your point but I feel ify about it, I would think that if there is a function "diff_of_prod" I should use it. And then have settings around it to do what I want.

But the notion of "accuracy helpers" also makes sense. I will leave this to you and @jfalcou

include/eve/module/complex/complex.hpp
jtlap: all the way the use of complex is to include all complex functions with no details. Talk with Joel.

jfalcou: We changed when we went for the modular approach. Include time of functions is mostly the same this way. 
For complex, we need a lot of them anyway and we wanted to mimic std so includign eve/complex.hpp gives you everything proper out of the box.

include/eve/module/complex/complex.hpp
jfalcou: we **may** actually have a raw(diff_of_prod) that just do a*b-c*d and have the other do the proper thing.
Then complex mul use raw(diff_of_prod) the pedantic use the regular.

THe cost of actual diff_of_prod with FMA semantic is high so we can't just use it willy nilly or our speed compared to auto vectorized complex multiplication is bad.

include/eve/module/core/regular/impl/sum_of_prod.hpp
```diff
@@ -33,6 +37,16 @@ namespace eve::detail
                                    ) noexcept
   requires(has_native_abi_v<T>)
   {
-    return a*b+c*d;
+    if constexpr(spy::supports::fma_ || std::same_as<T, float>)
```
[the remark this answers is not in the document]
jtlap: ok
je croyais que c'était deja parti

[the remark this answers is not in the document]
jtlap: done

PR #1357, opened 2022-07-29
sve firsts

DenisYaroshevskiy:
implementaion of enumerating make on sve


test/unit/arch/new_arch_first_tests.cpp
```diff
@@ -0,0 +1,71 @@
...
+// need to pass.
+//
+// Comment out 1 by 1.
+
+TTS_CASE("Implement basic make function for wide<int>")
+{
+  eve::wide<int> x {[](int i, int) { return i; }};
+
+  (void)x;
+  TTS_PASS();
```
jfalcou, line 24: We have TTS_COMPILE now for that, look up TTS doc.

DenisYaroshevskiy: Thanks - I tried, doesn't seem like a good fit.

It is a good tool for sfinae things and for less error messages.
Here it doesn't help much I think

PR #1363, opened 2022-08-01
Upgrade to latest Docker

jfalcou:
g++ 12 is on

### DIFF
--- a/.github/actions/integration/entrypoint.sh
+++ b/.github/actions/integration/entrypoint.sh
@@ -1,6 +1,9 @@
 #!/bin/sh -l
 set -e
 
+# Setup github/worspace as safe
+git config --global --add safe.directory /github/workspace
+
 # ID for various tests
 INSTALL_TEST=0
 FETCH_TEST=1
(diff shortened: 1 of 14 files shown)


cmake/toolchain/random.targets.json
```diff
@@ -0,0 +1,3 @@
+{
+  "random.exe": "^random\\..*\\.exe"
```
DenisYaroshevskiy, line 2: What's this?

.github/actions/integration/entrypoint.sh
```diff
@@ -1,6 +1,9 @@
 #!/bin/sh -l
 set -e
 
+# Setup github/worspace as safe
+git config --global --add safe.directory /github/workspace
```
DenisYaroshevskiy, line 5: What's this?

DenisYaroshevskiy (approved):
Couple of curious questions

.github/actions/integration/entrypoint.sh
jfalcou: Recent got update regarding to security require this in our situation

PR #1364, opened 2022-08-01
sve: new arch basic tests

DenisYaroshevskiy:
Some basic tests

### DIFF
--- a/include/eve/arch/arm/sve/sve_true.hpp
+++ b/include/eve/arch/arm/sve/sve_true.hpp
@@ -1,0 +1,28 @@
+//==================================================================================================
+/*
+  EVE - Expressive Vector Engine
+  Copyright : EVE Contributors & Maintainers
+  SPDX-License-Identifier: MIT
+*/
+//==================================================================================================
+#pragma once
+
+#include <eve/arch.hpp>
+
+namespace eve::detail
+{
+
+// Returns clear sve_true for type
+// Calling svptrue_b8 is OK for most cases, where you OK with
+// 1s in insignificant bits but sometimes you are not.
+template<typename T>
+svbool_t
+sve_true()
+{
+  if constexpr( sizeof(T) == 1 ) return svptrue_b8();
+  else if constexpr( sizeof(T) == 2 ) return svptrue_b16();
+  else if constexpr( sizeof(T) == 4 ) return svptrue_b32();
+  else if constexpr( sizeof(T) == 8 ) return svptrue_b64();
+}
+
+} // namespace eve::detail
--- a/include/eve/arch/arm/sve/tags.hpp
+++ b/include/eve/arch/arm/sve/tags.hpp
@@ -39,16 +39,20 @@ namespace eve
   //================================================================================================
   // Dispatching tag for ARM SIMD implementation
   //================================================================================================
-  struct sve128_  : simd_api<simd_   , spy::fixed_sve_> {};
-  struct sve256_  : simd_api<simd_   , spy::fixed_sve_> {};
-  struct sve512_  : simd_api<simd_   , spy::fixed_sve_> {};
+  struct sve_     : simd_api<simd_   , spy::sve_> {};
+  struct sve_vls_ : simd_api<sve_    , spy::fixed_sve_> {};
+  struct sve128_  : simd_api<sve_vls_, spy::fixed_sve_> {};
+  struct sve256_  : simd_api<sve_vls_, spy::fixed_sve_> {};
+  struct sve512_  : simd_api<sve_vls_, spy::fixed_sve_> {};
 
   //================================================================================================
   // SVE extensions tag objects
   //================================================================================================
-  inline constexpr sve128_  sve128  = {};
-  inline constexpr sve256_  sve256  = {};
-  inline constexpr sve512_  sve512  = {};
+  inline constexpr sve_     sve       = {};
+  inline constexpr sve_vls_ sve_vls   = {};
+  inline constexpr sve128_  sve128    = {};
+  inline constexpr sve256_  sve256    = {};
+  inline constexpr sve512_  sve512    = {};
 
   //================================================================================================
   // ARM SVE ABI concept
--- a/include/eve/detail/function/simd/arm/sve/load.hpp
+++ b/include/eve/detail/function/simd/arm/sve/load.hpp
@@ -1,0 +1,35 @@
+//==================================================================================================
+/*
+  EVE - Expressive Vector Engine
+  Copyright : EVE Contributors & Maintainers
+  SPDX-License-Identifier: MIT
+*/
+//==================================================================================================
+#pragma once
+
+#include <eve/as.hpp>
+#include <eve/concept/vectorizable.hpp>
+#include <eve/detail/category.hpp>
+#include <eve/detail/implementation.hpp>
+#include <eve/detail/spy.hpp>
+#include <eve/module/core/regular/unalign.hpp>
+#include <eve/arch/arm/sve/sve_true.hpp>
+
+namespace eve::detail
+{
+
+template<real_scalar_value T, typename N, simd_compatible_ptr<wide<T, N>> Ptr>
+EVE_FORCEINLINE wide<T, N>
+                load_(EVE_SUPPORTS(sve_),
+                      ignore_none_ const&,
+                      safe_type const&,
+                      eve::as<wide<T, N>> const&,
+                      Ptr p) requires(sve_abi<abi_t<T, N>>)
+{
+  auto ptr = unalign(p);
+
+  if constexpr( N() == expected_cardinal_v<T> ) return svld1(sve_true<T>(), ptr);
+  else return svldff1(sve_true<T>(), ptr);
+}
+
+}
--- a/include/eve/detail/function/simd/arm/sve/make.hpp
+++ b/include/eve/detail/function/simd/arm/sve/make.hpp
@@ -1,0 +1,114 @@
+//==================================================================================================
+/*
+  EVE - Expressive Vector Engine
+  Copyright : EVE Contributors & Maintainers
+  SPDX-License-Identifier: MIT
+*/
+//==================================================================================================
+#pragma once
+
+#include <eve/arch/fundamental_cardinal.hpp>
+#include <eve/as.hpp>
+#include <eve/detail/abi.hpp>
+#include <eve/detail/function/load.hpp>
+#include <eve/detail/meta.hpp>
+#include <eve/traits/as_integer.hpp>
+#include <eve/arch/arm/sve/sve_true.hpp>
+
+namespace eve::detail
+{
+//============================s====================================================================
+// Enumerated make
+//================================================================================================
+template<real_scalar_value T, typename N, typename... Vs>
+EVE_FORCEINLINE auto
+make(eve::as<wide<T, N>>, Vs... vs) noexcept requires sve_abi<abi_t<T, N>>
+{
+  static_assert(sizeof...(Vs) == N::value, "[eve::make] - Invalid number of arguments");
+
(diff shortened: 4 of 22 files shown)


include/eve/detail/function/simd/arm/sve/make.hpp
```diff
@@ -0,0 +1,115 @@
...
+#pragma once
+
+#include <eve/arch/fundamental_cardinal.hpp>
+#include <eve/as.hpp>
+#include <eve/detail/abi.hpp>
+#include <eve/detail/function/load.hpp>
+#include <eve/detail/meta.hpp>
+#include <eve/traits/as_integer.hpp>
+
+namespace eve::detail
```
DenisYaroshevskiy, line 17: I am not happy about this code but we should see what gcc actually generates.

include/eve/detail/function/simd/arm/sve/load.hpp
```diff
@@ -0,0 +1,39 @@
...
+                      eve::as<wide<T, N>> const&,
+                      Ptr p) requires(sve_abi<abi_t<T, N>>)
+{
+  auto ptr = unalign(p);
+
+  if constexpr( N() == expected_cardinal_v<T> ) return svld1(svptrue_b8(), ptr);
+  else
+  {
+    typename wide<T, N>::storage_type that {};
+    std::memcpy(&that, ptr, N::value * sizeof(T));
```
DenisYaroshevskiy, line 34: maybe this should be falt tolerant load

include/eve/arch/arm/sve/sve_true.hpp
```diff
@@ -0,0 +1,28 @@
...
+
+namespace eve::detail
+{
+
+// Returns clear sve_true for type
+// Calling svptrue_b8 is OK for most cases, where you OK with
+// 1s in insignificant bits but sometimes you are not.
+template<typename T>
+svbool_t
+sve_true()
```
DenisYaroshevskiy, line 20: very much on the fence with respect to where we should use typed or non-typed bool. We'll see I guess.

include/eve/arch/arm/sve/tags.hpp
```diff
@@ -39,16 +39,18 @@ namespace eve
   //================================================================================================
   // Dispatching tag for ARM SIMD implementation
   //================================================================================================
-  struct sve128_  : simd_api<simd_   , spy::fixed_sve_> {};
-  struct sve256_  : simd_api<simd_   , spy::fixed_sve_> {};
-  struct sve512_  : simd_api<simd_   , spy::fixed_sve_> {};
+  struct sve_fixed_ : simd_api<simd_      , spy::fixed_sve_> {};
```
DenisYaroshevskiy, line 42: Should be renamed but later

include/eve/arch/arm/sve/tags.hpp
```diff
@@ -39,16 +39,20 @@ namespace eve
   //================================================================================================
   // Dispatching tag for ARM SIMD implementation
   //================================================================================================
-  struct sve128_  : simd_api<simd_   , spy::fixed_sve_> {};
-  struct sve256_  : simd_api<simd_   , spy::fixed_sve_> {};
-  struct sve512_  : simd_api<simd_   , spy::fixed_sve_> {};
+  struct sve_     : simd_api<simd_   , spy::sve_> {};
+  struct sve_vls_ : simd_api<sve_    , spy::fixed_sve_> {};
```
DenisYaroshevskiy, line 43: Here is my thinking on `sve_`, `sve_vls_`

* we need a common tag type
* we can for now use `sve_` and `sve_vls_` interchangeably - obviously we don't support an unsized sve_ but if we feel like this function is unsized, we can use `sve_` tag. 

Then if/when we actually do the unsized sve, we will have less work (hopefully).

include/eve/detail/function/simd/arm/sve/load.hpp
```diff
@@ -0,0 +1,34 @@
...
+                load_(EVE_SUPPORTS(sve_),
+                      ignore_none_ const&,
+                      safe_type const&,
+                      eve::as<wide<T, N>> const&,
+                      Ptr p) requires(sve_abi<abi_t<T, N>>)
+{
+  auto ptr = unalign(p);
+
+  if constexpr( N() == expected_cardinal_v<T> ) return svld1(svptrue_b8(), ptr);
+  else return svldff1(svptrue_b8(), ptr);
```
DenisYaroshevskiy, line 31: I am using a fault tolerant load. 
The correct case will work. But maybe we should ensure that the asan will trigger on accessing garbage memory? I mean - this is a safe load.

Well - maybe later

include/eve/detail/function/simd/arm/sve/slice.hpp
```diff
@@ -0,0 +1,57 @@
...
+
+template<typename T, typename N, typename Slice>
+EVE_FORCEINLINE logical<wide<T, typename N::split_type>>
+                slice(logical<wide<T, N>> a, Slice) noexcept requires sve_abi<abi_t<T, N>>
+{
+  if constexpr( !Slice::value ) return a.storage();
+  else
+  {
+    auto bits = a.bitmask();
+    return svcmpne(svptrue_b8(), slice(bits, Slice{}), 0);
```
DenisYaroshevskiy, line 34: Couldn't come up with anything. They gave us nothing to move bits around

include/eve/detail/function/simd/arm/sve/subscript.hpp
```diff
@@ -0,0 +1,68 @@
...
+template<typename T, typename N>
+EVE_FORCEINLINE T
+insert(wide<T, N>& v, std::size_t i, T x) noexcept requires sve_abi<abi_t<T, N>>
+{
+  return v.storage()[i] = x;
+}
+
+template<typename T, typename N>
+EVE_FORCEINLINE logical<T>
+extract(logical<wide<T, N>> const& v, std::size_t i) noexcept requires sve_abi<abi_t<T, N>>
```
DenisYaroshevskiy, line 31: the logical ones are quite bad

include/eve/arch/arm/sve/tags.hpp
jfalcou: let's have (per ARM nomenclature) :
 - `sve_`  
 - `sve_vls_` inheriting from `sve_`  
 - `sve_vla_` inheriting from `sve_`
 
 ?

DenisYaroshevskiy: vla_ is `lengh agnostic`? then `vls_` should inherit `vla_`.
I just named `sve_` length agnostic one.

Can we do that in the next pr if this is the only change.

jfalcou: Sure, can wait next PR.
I think I agree with you on the vsl/ vsa thing.

.github/workflows/main.yml
```diff
@@ -39,9 +39,9 @@ jobs:
         - { comp: gcc         , arch: aarch64 , opts: -Wno-psabi, status: 1 }
         - { comp: gcc         , arch: arm     , opts: -Wno-psabi, status: 1 }
         - { comp: gcc         , arch: ppc64   , opts: -Wno-psabi, status: 1 }
-        - { comp: gcc         , arch: sve128  , opts: -Wno-psabi, status: 0 }
-        - { comp: gcc         , arch: sve256  , opts: -Wno-psabi, status: 0 }
-        - { comp: gcc         , arch: sve512  , opts: -Wno-psabi, status: 0 }
+        - { comp: gcc         , arch: sve128  , opts: -Wno-psabi -DEVE_USE_PCH=OFF, status: 0 }
```
jfalcou, line 42: I thought you wanted that in the toolchain files?

jfalcou: or even remove that since we have g++12 no?

include/eve/arch/arm/sve/sve_true.hpp
jfalcou: they have to be typed. Or the 1's won't be where they should in the internal representation.
I guess this detail function can also be used for the True() constant

include/eve/detail/function/simd/arm/sve/load.hpp
jfalcou: Sounds OK for now

include/eve/detail/function/simd/arm/sve/make.hpp
```diff
@@ -0,0 +1,115 @@
...
+make(as<logical<wide<T, N>>>, Vs... vs) noexcept requires sve_abi<abi_t<T, N>>
+{
+  using bits_type = typename logical<wide<T, N>>::bits_type;
+  using e_t       = element_type_t<bits_type>;
+
+  auto bits = make(as<bits_type> {}, (vs ? (e_t)-1 : 0)...);
+  return svcmpne(svptrue_b8(), bits, (e_t)0);
+}
+
+// svdup_n_bNN
```
jfalcou, line 90: spurious comment ?

include/eve/detail/function/simd/arm/sve/slice.hpp
jfalcou: Can you clarify what is going on ?

test/unit/arch/new_arch_first_tests.cpp
```diff
@@ -0,0 +1,198 @@
...
+// need to pass.
+//
+// Comment out 1 by 1.
+
+TTS_CASE("Implement basic make function for wide<int>")
+{
+  eve::wide<int> x {[](int i, int) { return i; }};
+
+  (void)x;
+  TTS_PASS();
```
jfalcou, line 24: can't this test be merged with the next one ?

test/unit/arch/new_arch_first_tests.cpp
```diff
@@ -0,0 +1,198 @@
...
+    for( int i = 0; i != full_wide.size(); ++i )
+    {
+      e_t expected = i < T::size() ? 3 : 0;
+      TTS_EQUAL(expected, full_wide.get(i));
+    }
+  }
+};
+
+TTS_CASE("Slice, wide<int>")
+{
```
jfalcou, line 108: why the duplicated test with the all ones ?

include/eve/detail/function/simd/arm/sve/slice.hpp
DenisYaroshevskiy: I convert to bits, slice and convert back?

include/eve/arch/arm/sve/sve_true.hpp
DenisYaroshevskiy: They don't always have to be typed. In some contexts, like in load, you can use the b8 one.

The question is: do we always use typed or do we use b8 when we can.
The answer is basically how many bool constants will we have.

Also also probably we will need one constant for each type involved regardless of anything, so using the appropriate one might save us having a b8

.github/workflows/main.yml
DenisYaroshevskiy: Unfortunately, still crashes. Yeah, let me move to toolchain.

include/eve/detail/function/simd/arm/sve/slice.hpp
jfalcou: OK, I got distracted byt the raw svcmpne in the middle. Got it now

.github/workflows/main.yml
jfalcou: OK

test/unit/arch/new_arch_first_tests.cpp
DenisYaroshevskiy: I duplicate tests to reduce on horrible horrible compiler messages first.
All of these are unnessary  - we have this stuff tested already

jfalcou (approved):

PR #1376, opened 2022-08-06
preparing sve top bits

DenisYaroshevskiy:
* ripped out avx512 top bits.
* redid top bits for avx512 as a logical inside.
* added a few sve operators
* fixed the testing for logicals


include/eve/detail/function/simd/arm/sve/make.hpp
```diff
@@ -98,7 +98,7 @@ make(eve::as<logical<wide<T, N>>>, V x) noexcept requires sve_abi<abi_t<T, N>> &
   {
     return [&]<std::size_t... i>(std::index_sequence<i...> const&)
     {
-      return make(as<logical<wide<T, f_t>>> {}, (i < N::value ? x : false)...);
+      return make(as<logical<wide<T, f_t>>> {}, (i < N::value ? (bool)x : false)...);
```
jfalcou, line 101: I would prefer `static_cast`.

include/eve/arch/cpu/top_bits.hpp
```diff
@@ -0,0 +1,488 @@
...
+  } // namespace eve
+
+// ---------------------------------------------------------------------------------
+// to_logical(top_bits)
+//
+
+namespace eve::detail {
+
+template <logical_simd_value Logical>
+EVE_FORCEINLINE Logical to_logical(eve::top_bits<Logical> mmask) noexcept
```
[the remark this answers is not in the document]
DenisYaroshevskiy: the problem is - predeclaration of to_logical
We need then to predeclare to separate versions: just for top_bits and for top_bits on x86

We should always use callables. Or do some other trick. For now let's leave this.

include/eve/detail/function/simd/arm/sve/make.hpp
DenisYaroshevskiy: If you ok with the other thing, let's do in the next patch

include/eve/arch/cpu/top_bits.hpp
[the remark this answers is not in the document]
jfalcou: OK I see

include/eve/detail/function/simd/arm/sve/make.hpp
jfalcou: Fine

jfalcou (approved):

PR #1384, opened 2022-08-16
Masked

jtlap:
more masked callable with tests
two pbs remain
the checked callables as shl shr
and the kind of sincos(x): it could be possible to make a masked returning {x,x} in the false case


include/eve/module/core/regular/binarize.hpp
```diff
@@ -53,7 +53,8 @@ namespace eve
 //!  @godbolt{doc/core/regular/binarize.cpp}
 //! @}
 //================================================================================================
-EVE_MAKE_CALLABLE(binarize_, binarize);
+
```
jfalcou, line 56: `binarize_not` has its masking de-activated. Why not `binarize` ?

include/eve/module/core/regular/impl/negmaxabs.hpp
```diff
@@ -9,63 +9,25 @@
...
-//================================================================================================
-template<decorator D, conditional_expr C, real_value U, real_value V>
+template<conditional_expr C, typename  ...Ts>
 EVE_FORCEINLINE auto
-negmaxabs_(EVE_SUPPORTS(cpu_), C const& cond, D const&, U const& t, V const& f) noexcept requires
-    compatible_values<U, V>
+negmaxabs_(EVE_SUPPORTS(cpu_), C const & c, Ts... args)
 {
-  return mask_op(cond, D()(eve::negmaxabs), t, f);
+  return minus[c](maxabs[c](args...));
```
jfalcou, line 30: doesn't it perform two `if_else` ?

include/eve/module/core/regular/binarize.hpp
jtlap: done

include/eve/module/core/regular/impl/negmaxabs.hpp
jtlap: yes when the intrinsic is not used. Is it problematic ?

PR #1386, opened 2022-08-20
 sve: unit arch exe

DenisYaroshevskiy:


include/eve/arch/arm/sve/tags.hpp
```diff
@@ -29,7 +29,7 @@ namespace eve
     static constexpr std::size_t expected_cardinal = bytes / sizeof(Type);
 
     template<typename Type>
-    static constexpr std::size_t fundamental_cardinal = 8 / sizeof(Type);
+    static constexpr std::size_t fundamental_cardinal = expected_cardinal<Type>;
```
jfalcou, line 32: Isn't expected cardinal using this value member ? Sounds a bit circular.

test/unit/arch/top_bits.cpp
```diff
@@ -260,3 +289,14 @@ TTS_CASE_TPL("Check top_bits slicing", eve::test::simd::all_types)
     }
   }
 };
+
+TTS_CASE_TPL("Check top_bits are printable", eve::test::simd::all_types)
+<typename T>(tts::type<T>)
+{
+  using logical = eve::logical<T>;
+  eve::top_bits     x {logical {true}};
+  std::stringstream s;
```
jfalcou, line 298: Thus should be a TTS_EXPECT_COMPILES

include/eve/arch/arm/sve/tags.hpp
DenisYaroshevskiy: No, it doesn't

PR #1392, opened 2022-08-27
Add a local doxygen generation target

jfalcou:
When debugging documentation, it helps if we don't pollute the actual docs/ folder. The new doxygen-local target generates the documentation in the build folder instead.


DenisYaroshevskiy:
Can you maybe add a comment to dev cmake doc?

DenisYaroshevskiy (approved):

DenisYaroshevskiy (approved):

PR #1396, opened 2022-09-03
Missing constexpr leads to constants in the output

DenisYaroshevskiy:


DenisYaroshevskiy:
Can't approve as the original author but approved

jfalcou (approved):

PR #1427, opened 2022-11-02
Prevent tuple related issues with traits and concept

jfalcou:
- Some situations made compilers try to compute traits too eagerly, leading to errors.
 - Some traits forced the requirements that UDT be literal while this is not necessarily enforceable.


include/eve/traits/as_wide.hpp
```diff
@@ -16,7 +16,10 @@
 namespace eve
 {
   template<typename Type, typename Size = expected_cardinal_t<Type>>
-  struct as_wide;
+  struct as_wide
+  {
+    using type = Type;
```
DenisYaroshevskiy, line 21: This does not seem right. Why do you need this

include/eve/traits/max_scalar_size.hpp
```diff
@@ -28,8 +41,8 @@ namespace eve
   //! @}
   //================================================================================================
   template <typename T>
-  constexpr std::size_t max_scalar_size_v = kumi::max_flat(T{}, [](auto m) { return sizeof(m); });
+  constexpr std::size_t max_scalar_size_v = detail::minmax_scalar<T>(true);
 
   template <typename T>
-  constexpr std::size_t min_scalar_size_v = kumi::min_flat(T{}, [](auto m) { return sizeof(m); });
+  constexpr std::size_t min_scalar_size_v = detail::minmax_scalar<T>(false);
```
DenisYaroshevskiy, line 47: min_scalar_size seems unused btw.

include/eve/traits/max_scalar_size.hpp
```diff
@@ -11,6 +11,19 @@
...
+  namespace detail
+  {
+    template<typename T> constexpr auto minmax_scalar(bool mode) noexcept
+    {
+      if constexpr(kumi::product_type<T>)
+      {
+        if(mode) return kumi::max_flat(kumi::as_tuple_t<T>{}, [](auto m) { return sizeof(m); });
+        else     return kumi::min_flat(kumi::as_tuple_t<T>{}, [](auto m) { return sizeof(m); });
+      }
+      else return sizeof(T);
```
DenisYaroshevskiy, line 23: Why do you need this? We only use this in soa_vector, that should be kumi product

DenisYaroshevskiy:
Tests please

include/eve/traits/as_wide.hpp
jfalcou: When making a soa_vector of eve::complex, clang tries incorrectly to instantiate as_wide_as<Z,R> in a requires clause of complex += while trying to resolve some `begin() + i` in soa_vector. This instantiation fails and g++ decides the required clause to be false while clang thinks it warrants a hard error.

include/eve/traits/max_scalar_size.hpp
jfalcou: Yes I noticed it. I can remove it.

include/eve/traits/max_scalar_size.hpp
jfalcou: In fact we also use it here:

```c++
/home/falcou/infra/tindalos-2/_work/eve/eve/include/eve/traits/max_scalar_size.hpp:31:67: error: constraints not satisfied for alias template 'as_tuple_t' [with Tuple = signed char, Meta = std::type_identity]
  constexpr std::size_t max_scalar_size_v = kumi::max_flat( kumi::as_tuple_t<T>{}
                                                                  ^~~~~~~~~~~~~
/home/falcou/infra/tindalos-2/_work/eve/eve/include/eve/detail/compress/simd/x86/compress_using_masks.hpp:167:32: note: in instantiation of variable template specialization 'eve::max_scalar_size_v<signed char>' requested here
    requires ( N() == 16 ) && (max_scalar_size_v<T> <= 2)
```

The old implementation used T{} instead of the as_tuple_t<T>{} that fix the issue of forcing UDT to be constexpr def-constructible. Now we need to explicitley say that we take care of scalars.

If I remove `min_scalar_size` I can however simplify the rest of the code

jfalcou: Unless you tell me this is an issue in `compress_using_masks`

jfalcou: Tests already do what is requried IMHO:

```c++
#include <eve/traits/max_scalar_size.hpp>

TTS_CASE( "Check max_scalar_size" )
{
  TTS_CONSTEXPR_EQUAL(sizeof(int),  eve::max_scalar_size_v<int>);
  TTS_CONSTEXPR_EQUAL(sizeof(std::int8_t), eve::max_scalar_size_v<std::int8_t>);
  TTS_CONSTEXPR_EQUAL(sizeof(std::int8_t), (eve::max_scalar_size_v<kumi::tuple<std::int8_t, std::int8_t>>));
  TTS_CONSTEXPR_EQUAL(sizeof(int),  (eve::max_scalar_size_v<kumi::tuple<
                                       int, kumi::tuple<std::int8_t, std::int8_t>>>));
};
```

jfalcou: An alternative could be that `as_tuple_t<non product-type T>` in kumi returns tuple<T>.

jfalcou: The more it goes, the more the kumi fix makes more sense.

include/eve/traits/as_wide.hpp
jfalcou: One way to make it less awkward would be to specialize on <T,scalar_cardinal> which is the only case that is not handled specifically. This makes the fix feels less artificial.

include/eve/traits/max_scalar_size.hpp
jfalcou: Amended kumi and simplified the fix here

include/eve/traits/as_wide.hpp
DenisYaroshevskiy: I don't remember - what's scalar cardinal? Is it fixed<1>?
Can we add requires false to this struct or smth?

Can you add the complex test you mention?

jfalcou: scalar_cardinal is a special cardinal different from `fixed<1>` that is returned when asked for the cardinal of a scalar.

Let me add a silly compile test, remove this fix so you can see the problem and we decide on the correct process to fix.

include/eve/concept/scalar.hpp
```diff
@@ -34,13 +34,14 @@ namespace detail
...
   template<typename T> constexpr bool scalar_tuple()
   {
-    if constexpr( kumi::product_type<T> )
+    if constexpr( !kumi::product_type<T> ) return false;
+    else
     {
-      return kumi::size_v<T> ? kumi::all_of(kumi::flatten_all(kumi::as_tuple_t<T> {}),
-                                            []<typename M>(M) { return plain_scalar_value<M>; })
-                             : false;
+      if constexpr(kumi::size_v<T> == 0) return false;
```
DenisYaroshevskiy, line 40: kumi all of should do this

DenisYaroshevskiy (approved):

PR #1428, opened 2022-11-02
Use convert for logical operations between logical of different base types

jfalcou:
…type

### DIFF
--- a/include/eve/detail/function/simd/common/friends.hpp
+++ b/include/eve/detail/function/simd/common/friends.hpp
@@ -22,6 +22,8 @@ namespace eve
 {
   EVE_REGISTER_CALLABLE(is_equal_)
   EVE_REGISTER_CALLABLE(is_not_equal_)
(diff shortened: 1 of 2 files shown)


include/eve/detail/function/simd/common/friends.hpp
```diff
@@ -24,6 +24,9 @@ namespace eve
   EVE_REGISTER_CALLABLE(is_not_equal_)
 }
 
+// This file needs tag::if_not_equal_ to be visible
```
DenisYaroshevskiy, line 27: Just do a predeclaration for convert somewhere here, no?

DenisYaroshevskiy (approved):

include/eve/detail/function/simd/common/friends.hpp
jfalcou: Yeah better

DenisYaroshevskiy (approved):

PR #1436, opened 2022-11-08
Add more passing tests to SVE subset

jfalcou:


DenisYaroshevskiy (approved):

include/eve/detail/function/simd/arm/sve/to_logical.hpp
```diff
@@ -0,0 +1,24 @@
...
+#include <eve/forward.hpp>
+#include <eve/traits/as_logical.hpp>
+
+namespace eve::detail
+{
+template<typename T, typename N>
+EVE_FORCEINLINE auto
+to_logical(wide<T, N> const& v) noexcept requires sve_abi<abi_t<T, N>>
+{
+  return v != 0;
```
DenisYaroshevskiy, line 22: Is there maybe an intrinsic?

PR #1438, opened 2022-11-09
Implement combine

jfalcou:
With a side serving of load for logical so that tests can be compiled

### DIFF
--- a/include/eve/detail/function/simd/arm/sve/load.hpp
+++ b/include/eve/detail/function/simd/arm/sve/load.hpp
@@ -38,11 +38,7 @@ load_ ( EVE_SUPPORTS(sve_), ignore_none_ const&, safe_type const&
       ) noexcept
 requires(dereference_as<logical<T>, Pointer>::value && sve_abi<abi_t<T, N>>)
 {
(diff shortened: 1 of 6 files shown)


include/eve/detail/function/simd/arm/sve/load.hpp
```diff
@@ -7,29 +7,42 @@
...
+EVE_FORCEINLINE logical<wide<T, N>>
+load_ ( EVE_SUPPORTS(sve_), ignore_none_ const&, safe_type const&
+      , eve::as<logical<wide<T, N>>> const&, Pointer ptr
+      ) noexcept
+requires(dereference_as<logical<T>, Pointer>::value && sve_abi<abi_t<T, N>>)
+{
+  auto block = [&]() -> wide<T, N>
+  {
+    return load(ignore_none, safe, eve::as<wide<T, N>>{}, ptr_cast<T const>(ptr));
+  }();
```
DenisYaroshevskiy, line 44: Why lambda, confused

DenisYaroshevskiy (approved):

PR #1447, opened 2022-11-14
Enfore convert implementation around logical

jfalcou:
+ no logical<->wide conversion is now properly implemented
 + no wide logical moved to x86

### DIFF
--- a/include/eve/module/core/regular/impl/convert.hpp
+++ b/include/eve/module/core/regular/impl/convert.hpp
@@ -11,14 +11,12 @@
 #include <eve/concept/value.hpp>
 #include <eve/detail/has_abi.hpp>
 #include <eve/detail/implementation.hpp>
-#include <eve/module/core/constant/valmax.hpp>
-#include <eve/module/core/constant/valmin.hpp>
 #include <eve/module/core/constant/zero.hpp>
 #include <eve/module/core/regular/all.hpp>
 #include <eve/module/core/regular/bit_cast.hpp>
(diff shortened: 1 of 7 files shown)


include/eve/module/core/regular/impl/convert.hpp
```diff
@@ -48,9 +55,8 @@ struct convert_lambda
 };
 
 template<product_type T, product_type U>
-requires(kumi::result::flatten_all_t<T>::size()
-         == kumi::result::flatten_all_t<U>::size()) EVE_FORCEINLINE
-    auto convert_(EVE_SUPPORTS(cpu_), T const& v0, eve::as<U>)
+requires(kumi::result::flatten_all_t<T>::size() == kumi::result::flatten_all_t<U>::size())
+EVE_FORCEINLINE auto convert_(EVE_SUPPORTS(cpu_), T const& v0, eve::as<U>)
```
DenisYaroshevskiy, line 59: Can we do `tuple<logical<int>, int>`? Because if we can, this is not a sufficient restriction.

I don't think you should bother very much - we can fix the sfinae friendlinness if we ever need to.

DenisYaroshevskiy (approved):

include/eve/module/core/regular/impl/convert.hpp
jfalcou: yes, we need something like all of members of in convert to elements of U. 
We can fix that later, main focus here is stopping silly wide->logical and moving avx512 like treatment to x86 files

PR #1455, opened 2022-11-23
Activate some basic SVE tests

jfalcou:


include/eve/detail/function/simd/common/to_logical.hpp
```diff
@@ -58,7 +59,10 @@ EVE_FORCEINLINE auto
 to_logical(C c, eve::as<T>) noexcept
 {
   using l_t = typename as_logical<T>::type;
-  using i_t = as_integer_t<typename as_logical_t<T>::mask_type>;
+  using i_t = std::conditional_t< (T::size()>=128 && sizeof(element_type_t<T>) == 1)
+                                , as_integer_t<typename as_logical_t<T>::mask_type,unsigned>
+                                , as_integer_t<typename as_logical_t<T>::mask_type>
+                                >;
```
DenisYaroshevskiy, line 65: This still does not make sense to me

jfalcou: I am gonna add a comment

DenisYaroshevskiy (approved):
Approved but we agreed that the comment will change to reflect that signed compare is better

Issue #1236, opened 2022-02-10
[FEATURE] Support for MSVC


jfalcou:
This is a mutable, ongoing issue to discuss/list steps for proper MSVC support.

# Current Status
Since MSVC 2022 19.30, C++20 support has drastically improved. Building blocks libraries like kumi and raberu now natively compile on MSVC without any hassle. This has triggered the investigation of how much EVE needs to be fixed to make MSVC EVE a thing.

Note that we currently don't care about code generation quality as our first goal here is to get the library to compile.

# Plan of Actions
The first batch of trivial fixes has been PR'ed in #1233 and fixed some glaring issues with fundamental code.
We'll list there next steps to be taken.

# Current Status
- `unit.arch.exe` compiles and pass tests

DenisYaroshevskiy:
CI?

I'll have to install windows somewhere, won't I?

jfalcou:
Was thinking of using appveyor

Tradias:
Is there anything you need help with here?

Apart from appveyor, I have only ever used Github Actions and they were easy to set up.

jfalcou:
I just need to find time to tackle the stupid errors that prevent other test to compile.

 Appveyor is setup already.

jfalcou:
There looks like some compile time trials we have don't fly properly but the error messages are outrageous.

I  an push a more up-to-date list of breaking thing later today.

jfalcou:
So mostly we need help understanding why msvc don't understand our properly standard templatr monstrosity

Tradias:
Sure, sounds like something I can investigate.

jfalcou:
Good. Let me update this issue later withore info

jfalcou:
@Tradias added a small roadmap on how to tackle the thing. Tell me if you need help.

jfalcou:
# unit.api targets
 - Looks like we try to use constepxr value as lambda capture in some of the shuffle tests and MSVC doesn't appreciate that

Tradias:
Looks like, I have managed to get `unit.core.exe` to compile without significant code changes. Creating a pr now.

jfalcou:
Good job wow :DDid you used the in flight PR that fixes API ? or were the issue unrelated ?

Tradias:
I didn't know about https://github.com/jfalcou/eve/pull/1458, one of the issue was in fact related. I think my solution might be faster to compile but potentially incorrect.

jfalcou:
[...]

PR #1460, opened 2022-11-26
Make unit.core.exe compile with MSVC

Tradias:
* I slightly altered the logic of `value_type_t` it now prefers nested `T::value_type` over range/iterator. Not sure whether this will have a negative effect on the other modules.
* I had to add includes for frexp and exp2 to two of the tests, not sure whether it is supposed to be like that.
* `include(CTest)` is not necessary to run tests with ctest, it just clutters the project with unnecessary targets.
* [std::inclusive_scan](https://en.cppreference.com/w/cpp/algorithm/inclusive_scan) is in fact in the header <numeric> and not <algorithm>

There is also an issue in https://github.com/jfalcou/tts/blob/main/include/tts/tts.hpp#L943 which can create a `std::uniform_int_distribution<uint64_t>` from -10 to +10 which become 18446744073709551606 to 10, which triggers an assertion in MSVC:

```
_STL_ASSERT(_Min0 <= _Max0, "invalid min and max arguments for uniform_int");
```

Resolves a part of https://github.com/jfalcou/eve/issues/1236

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -111,12 +111,14 @@ jobs:
           cmake --build . --target unit.arch.exe      --config ${{ matrix.cfg.mode }} --parallel 2
           cmake --build . --target unit.meta.exe      --config ${{ matrix.cfg.mode }} --parallel 2
           cmake --build . --target unit.internals.exe --config ${{ matrix.cfg.mode }} --parallel 2
+          cmake --build . --target unit.core.exe      --config ${{ matrix.cfg.mode }} --parallel 2
       - name: Running Tests
         run: |
           cd build
           ctest -C ${{ matrix.cfg.mode }} --output-on-failure -R ^unit.arch.*.exe
           ctest -C ${{ matrix.cfg.mode }} --output-on-failure -R ^unit.meta.*.exe
           ctest -C ${{ matrix.cfg.mode }} --output-on-failure -R ^unit.internals.*.exe
+          ctest -C ${{ matrix.cfg.mode }} --output-on-failure -R ^unit.core.*.exe
 
   avx512:
     runs-on: [self-hosted, avx512]
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -44,7 +44,6 @@ endif()
 ## =================================================================================================
 if( EVE_BUILD_TEST )
   include(config/dependencies)
-  include(CTest)
   include(target/add_parent_target)
   include(config/compiler)
   include(config/types)
--- a/test/unit/module/core/constants.cpp
+++ b/test/unit/module/core/constants.cpp
@@ -8,6 +8,9 @@
 #include "test.hpp"
 
 #include <eve/module/core.hpp>
+#include <eve/module/math/regular/exp2.hpp>
+
+#include <limits>
 
 TTS_CASE_TPL("Check basic constants behavior", eve::test::simd::ieee_reals)
 <typename T>(tts::type<T>)
@@ -23,7 +26,7 @@ TTS_CASE_TPL("Check basic constants behavior", eve::test::simd::ieee_reals)
 
   if constexpr( eve::floating_value<T> )
   {
-    TTS_IEEE_EQUAL(eve::allbits(as<T>()), T(0.0 / 0.0));
+    TTS_IEEE_EQUAL(eve::allbits(as<T>()), T(std::numeric_limits<double>::quiet_NaN()));
     TTS_EQUAL(eve::mzero(as<T>()), T(-0));
     TTS_EQUAL(eve::half(as<T>()), T(0.5));
     TTS_EQUAL(eve::mhalf(as<T>()), T(-0.5));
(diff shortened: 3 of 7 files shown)


Tradias:
I have used the following compiler:

```
C:\Program Files\Microsoft Visual Studio\2022\Community>cl --version
Microsoft (R) C/C++ Optimizing Compiler Version 19.34.31933 for x64
Copyright (C) Microsoft Corporation.  All rights reserved.
```

and compile options:
```
/wd4267 /wd4244 /wd4146 /bigobj /DWIN32 /D_WINDOWS /GR /EHsc /permissive- /Zc:__cplusplus /Zc:inline /Zc:sizedDealloc /Zc:externConstexpr /Zc:lambda /Zc:throwingNew
```

jfalcou:
Looks good. A few remarks:

> I slightly altered the logic of value_type_t it now prefers nested T::value_type over range/iterator. Not sure whether this will have a negative effect on the other modules.

This probably have to be merged with the fix from #1458 

>I had to add includes for frexp and exp2 to two of the tests, not sure whether it is supposed to be like that.

This is somethign which is an oversight. There is a fix for this in an upcoming PR so this is OK for now.

> include(CTest) is not necessary to run tests with ctest, it just clutters the project with unnecessary targets.

Guess it was some old habits

> [std::inclusive_scan](https://en.cppreference.com/w/cpp/algorithm/inclusive_scan) is in fact in the `<numeric>` header and not `<algorithm>`

Good catch

> There is also an issue in https://github.com/jfalcou/tts/blob/main/include/tts/tts.hpp#L943 which can create a std::uniform_int_distribution<uint64_t> from -10 to +10 which become 18446744073709551606 to 10, which triggers an assertion in MSVC:

Would you mind to file an issue at https://github.com/jfalcou/tts/issues ?

jfalcou:
> ```/wd4267 /wd4244 /wd4146 /bigobj /DWIN32 /D_WINDOWS /GR /EHsc /permissive- /Zc:__cplusplus /Zc:inline /Zc:sizedDealloc /Zc:externConstexpr /Zc:lambda /Zc:throwingNew```

as a rather naive user of MSVC, are those flags somehow classics ? If yes, we may need to update `https://github.com/jfalcou/eve/blob/main/cmake/config/compiler.cmake` ?

jfalcou:
As expected, you hit the "ambiguous `value_type`" eror ron clang/g++
I suggest we merge #1458 and rebase your PR ?

jfalcou:
#1458 is merged, so I guess you can rebase

Tradias:
Out of the compile options only `/bigobj` is actually needed to compile some of the tests, otherwise they fail with:

```
[build] C:\3YOURMIND\eve\test\unit\module\core\add.cpp : fatal error C1128: number of sections exceeded object file format limit: compile with /bigobj
```

The options `/wd4267 /wd4244 /wd4146` disable conversion warnings and a warning about unary minus operator.

It seems that [`/permissive-` is implied by `/std:c++20`](https://learn.microsoft.com/en-us/cpp/build/reference/permissive-standards-conformance?view=msvc-170) so no reason to set it explicitly in this project.

The `/Zc` options just enable additional C++ standard conformance behavior that is not covered by `/permissive-` and in rare cased helped me to write more portable code.

The options `/DWIN32 /D_WINDOWS /GR /EHsc` are usually set by CMake by default, basically: `/GR == -frtti` and `/EHsc == -fexceptions`.

jfalcou:
OK so I guess you need to update the compielrs.cmake file so the appveyor CI can use those options

jfalcou:
OK, first world problem, we hit the 1hr compile time limit on AppVeyor. I'll setup a windows machine next week so we can self-host the tests.

jfalcou:
https://github.com/jfalcou/eve/pull/1461 is tryign to make AppVeyor more happy

DenisYaroshevskiy:
Looked through, looks good so far

jfalcou:
#1461 is merged, woudl you mind rebasing if now appveyor is happy in term of compile time ?

jfalcou:
We still hit the limitations. Trying to see what we can do before I can setup a proper self hosted windows machine

jfalcou:
@Tradias woudl you mind modifying the appveyor.yml file so we have this:

```
- cmake --build . --target unit.arch.exe      --config Debug -j 2
- cmake --build . --target unit.meta.exe      --config Debug -j 2
- cmake --build . --target unit.internals.exe --config Debug -j 2
```

?

jfalcou:
oh wait, I thought we just add -j 2 to the build, not removing the core.exe sorry

Tradias:
In the long run you should consider to link all tests into one executable, at least for CI builds. That is significantly faster, especially on Windows (where we don't have high-speed linkers like [mold](https://github.com/rui314/mold))

jfalcou:
> In the long run you should consider to link all tests into one executable, at least for CI builds. That is significantly faster, especially on Windows (where we don't have high-speed linkers like [mold](https://github.com/rui314/mold))

Yeah that's somethign I was thinking about. 
This may requires a bit of cmakery.

We could move to GA with theire 6h limit but I fear they don't have the proper version of MSVC.
I'll dig for a machine at work.

DenisYaroshevskiy:
I mean, we tried to reduce the number of binaries, right? We even did unity builds which were supposed fto be big time faster. We got no speedups

jfalcou:
I am gonna merge a PR that movee our MSVC tests on Github Actions. THe version of the compiler is OK and we have a 6h limits.

jfalcou:
It may be too soon, got some more error than anticipated. I'll gonna review the code and see when GA windows got updated

jfalcou:
Yeah, I am hacking at it. Looks like same simplifcation that you already did.

jfalcou:
And oh, the core tests have liek few errors so that's encouraging!

PR #1473, opened 2022-12-04
Add support for SVE specific reduction

jfalcou:


include/eve/detail/function/simd/arm/sve/sum.hpp
```diff
@@ -0,0 +1,54 @@
...
+namespace eve::detail
+{
+template<conditional_expr C, typename T, typename N>
+EVE_FORCEINLINE auto
+sum_(EVE_SUPPORTS(sve_), C const& cond, wide<T, N> v) noexcept -> T
+requires sve_abi<abi_t<T, N>>
+{
+  // Fix mask to not touch garbage lanes if it is not an ignore_*
+  auto m = cond.mask(as(v));
+  if constexpr(!relative_conditional_expr<C>) m = remove_garbage(cond.mask(as(v)));
```
DenisYaroshevskiy, line 25: `remove_garbage(m)` - don't call mask twice.

We need a function for this.. But maybe it's not that important though.

jfalcou: remove_garbage works only on wide and logical. If m is a conditional, I need to call remove_garbage on its logical mask.

include/eve/module/core/regular/impl/simd/arm/sve/minimum.hpp
```diff
@@ -0,0 +1,54 @@
...
+
+template<typename T, typename N>
+EVE_FORCEINLINE auto
+minimum_(EVE_SUPPORTS(sve_), wide<T, N> v) noexcept -> T
+requires sve_abi<abi_t<T, N>>
+{
+  if constexpr( N::value == 1 ) return v.get(0);
+  else if constexpr( N::value == expected_cardinal_v<T> ) return minimum[ignore_none](v);
+  else return minimum[keep_first(N::value)](v);
+}
```
DenisYaroshevskiy, line 53: I don't think you need this keep_first. Ignore none should mask properly

DenisYaroshevskiy (approved):

PR #1474, opened 2022-12-06
SVE: Add support for fma-like intrinsics

jfalcou:


include/eve/module/core/regular/impl/simd/arm/sve/fam.hpp
```diff
@@ -0,0 +1,33 @@
...
+}
+
+template<conditional_expr C, arithmetic_scalar_value T, typename N>
+EVE_FORCEINLINE auto
+fam_(EVE_SUPPORTS(sve_), C cond, wide<T, N> v0, wide<T, N> v1, wide<T, N> v2) noexcept -> wide<T, N>
+requires sve_abi<abi_t<T, N>>
+{
+  auto const alt = alternative(cond, v0, as(v0));
+  if constexpr( C::is_complete && !C::is_inverted ) return alt;
+  return svmla_m(cond.mask(as(v0)), alt, v1, v2);
```
DenisYaroshevskiy, line 31: Hmm. So here is the thing. This will zero out top bits at the expense of computing a mask. For ignore none we could pass true here. What do you think?

include/eve/module/core/regular/impl/simd/arm/sve/fsm.hpp
```diff
@@ -0,0 +1,22 @@
...
+#include <eve/detail/implementation.hpp>
+
+namespace eve::detail
+{
+template<arithmetic_scalar_value T, typename N>
+EVE_FORCEINLINE auto
+fsm_(EVE_SUPPORTS(sve_), wide<T, N> v0, wide<T, N> v1, wide<T, N> v2) noexcept -> wide<T, N>
+requires sve_abi<abi_t<T, N>>
+{
+  return -fanm(v0, v1, v2);
```
DenisYaroshevskiy, line 20: Checking, this is not default right?

jfalcou: No, SVE changed their FMA model so the basic ones are not the one we had for ARM

include/eve/module/core/regular/impl/simd/arm/sve/fam.hpp
jfalcou: Well, I used the classical model of using the masked version ot do the non masked. What should we do here ?

DenisYaroshevskiy (approved):

Issue #1478, opened 2022-12-09
[FEATURE] runtime detection on apple clang


DenisYaroshevskiy:
```
/Users/denisarosevskij/space/eve/include/eve/arch/arm/detection.hpp:28:29: error: use of undeclared identifier 'AT_HWCAP'
    auto hwcaps = getauxval(AT_HWCAP);
                            ^
/Users/denisarosevskij/space/eve/include/eve/arch/arm/detection.hpp:38:29: error: use of undeclared identifier 'AT_HWCAP'
    auto hwcaps = getauxval(AT_HWCAP);
                            ^
/Users/denisarosevskij/space/eve/include/eve/arch/arm/detection.hpp:39:22: error: use of undeclared identifier 'HWCAP_ASIMD'
    return (hwcaps & HWCAP_ASIMD) != 0;
```

I found that bitcoin repo solved this:
Discussion: https://github.com/bitcoin/bitcoin/issues/20371#issuecomment-740749188
PR: https://github.com/bitcoin/bitcoin/pull/20603/files

jfalcou:
More digging in the Apple documentation:

```
To verify that your CPU uses the ARM Instruction Set Architecture, use hw.optional.arm64. A value of 1 indicates that the platform uses the ARM ISA.

The following parameters provide information about the availability of advanced single instruction multiple data (SIMD) and floating point capabilties. These features are standard in Apple processors beginning with M1 and A7, and don’t need to be checked.

hw.optional.AdvSIMD
General support for Advanced SIMD instructions.

This parameter replaces hw.optional.neon.

hw.optional.floatingpoint
General support for floating-point instructions.
```

So I guess checking for `hw.optional.arm64` should be enough ?

DenisYaroshevskiy:
I dunno. I think it's always a64 on macos. If you have a code you want me to test, I can

jfalcou:
will drop one yes

jfalcou:
Can you test this quick check ?

```c++
#include <iostream>
#include <sys/types.h>
#include <sys/sysctl.h>

bool has_neon()
{
  int val = 0;
  size_t len = sizeof(val);
  return sysctlbyname("hw.optional.AdvSIMD", &val, &len, nullptr, 0) == 0 && val != 0;
}

int main()
{
  std::cout << std::boolalpha << has_neon() << "\n";
}
```

phprus:
@jfalcou 
Apple M1 Max:
```
[...]

PR #1479, opened 2022-12-09
Fix apple clang build

DenisYaroshevskiy:
It works!

* apple clang does not like size_t and ptrdiff_t much
* had some hard errors in concepts - hence wide_cardinal concept
* aligned alloc needs no less than 8 bytes alignment 

What does not work? 
#1478 

Square is not precise enough. I don't know what to do with this one.
```
[X] - Check behavior of sqr on wide
  >  with [T = eve::wide<double, eve::fixed<4>>]
    [sqr.cpp:45] - ** FAILURE ** : Expected: eve::sqr(z_t{a0,a1}) == z*z but (506.392-47099.3i, -802.259+298.873i, -9748.24-276.767i, -43749.8+1229.74i) == (506.392-47099.3i, -802.259+298.873i, -9748.24-276.767i, -43749.8+1229.74i) within 15.50 ULP when 2.00 ULP was expected.
  >  with [T = eve::wide<double, eve::fixed<4>>]
    [sqr.cpp:47] - ** FAILURE ** : Expected: eve::pedantic(eve::sqr)(z) == z*z but (506.392-47099.3i, -802.259+298.873i, -9748.24-276.767i, -43749.8+1229.74i) == (506.392-47099.3i, -802.259+298.873i, -9748.24-276.767i, -43749.8+1229.74i) within 16.50 ULP when 6.00 ULP was expected.
```

UPD: fixed square by putting 18.0 (according ot @jfalcou ).


include/eve/memory/aligned_allocator.hpp
```diff
@@ -72,11 +74,13 @@ namespace eve
 
     void * allocate_aligned(std::size_t n, std::size_t a)
     {
+      if (a < 8U) a = 8U;
```
jfalcou, line 77: `a = std::min(a,8U);`  ?

test/unit/meta/concepts/scalar.cpp
```diff
@@ -82,8 +79,6 @@ TTS_CASE("Check for arithmetic_scalar_value on plain_scalar_value" )
   TTS_EXPECT( eve::arithmetic_scalar_value<std::uint16_t>  );
   TTS_EXPECT( eve::arithmetic_scalar_value<std::uint32_t>  );
   TTS_EXPECT( eve::arithmetic_scalar_value<std::uint64_t>  );
-  TTS_EXPECT( eve::arithmetic_scalar_value<std::size_t>    );
```
jfalcou, line 85: I need to get more info on what Apple Clang doesn't liek about those. 
size_t and ptrdiff_t are common type people will want to use.

include/eve/memory/aligned_allocator.hpp
DenisYaroshevskiy: max. I thought about it but I figured I can not bring algorithm. 

Also size_t and U are different types on apple clang - I'd have to cast

DenisYaroshevskiy: Tried - because of the cast I don't want to.

include/eve/module/bessel/detail/evaluate_rational.hpp
```diff
@@ -42,7 +42,7 @@ reverse_evaluate_rational(const T num, const U den, V z) noexcept
     z      = rec(z);
     auto r = reverse_horner(z, num) / reverse_horner(z, den);
     if( size(den) == size(num) ) return r;
-    else return r * pow(z, size(num) - size(den));
+    else return r * pow(z, (std::uint64_t) size(num) - size(den));
```
DenisYaroshevskiy, line 45: this pops up because size_t is not passing the concept. And going to pow is not working.

test/unit/meta/concepts/scalar.cpp
DenisYaroshevskiy: So is char. But it doesn't work.

If we want to support all of the weird types that are not actually the same, we have to solve it as a separate problem.

include/eve/algo/views/iota.hpp
```diff
@@ -97,7 +97,7 @@ namespace eve::algo::views
 
     EVE_FORCEINLINE friend T tagged_dispatch(eve::tag::read_, iota_with_step_iterator self)
     {
-      return self.base + eve::convert(self.i * self.step, eve::as<T>{});
+      return self.base + (T)self.i * self.step;
```
DenisYaroshevskiy, line 100: This cast is because you can't eve::convert ptrdifft

DenisYaroshevskiy (commented):
I think that they `size_t` `ptrdifft` `char` and so on are a separate prolbem.
We should solve it as a dedicated thing: https://github.com/jfalcou/eve/issues/1481

jfalcou (approved):

meftunca:
Please Update Conan Package

meftunca:
I am Getting these errors

```bash
ERROR: There are invalid packages (packages that cannot exist for this configuration):
jfalcou-eve/v2022.09.1: Invalid ID: EVE does not support apple Clang due to an incomplete libcpp.
```

jfalcou:
This will be done for next release.

meftunca:
When will the new version be released?

PR #1489, opened 2022-12-16
Some more SVE functions

jfalcou:
+ countl_zero
 + countr_zero
 + popcount
 + Some fixes for dec, inc, minus and add

### DIFF
--- a/include/eve/module/core/regular/impl/simd/arm/sve/countr_zero.hpp
+++ b/include/eve/module/core/regular/impl/simd/arm/sve/countr_zero.hpp
@@ -1,0 +1,38 @@
+//==================================================================================================
+/*
+  EVE - Expressive Vector Engine
+  Copyright : EVE Project Contributors
+  SPDX-License-Identifier: BSL-1.0
+*/
+//==================================================================================================
+#pragma once
+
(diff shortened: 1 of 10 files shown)


include/eve/module/core/regular/impl/simd/arm/sve/countr_zero.hpp
```diff
@@ -0,0 +1,38 @@
...
+countr_zero_(EVE_SUPPORTS(sve_), C const& cond, wide<T, N> v) noexcept -> wide<T, N>
+requires sve_abi<abi_t<T, N>>
+{
+  if constexpr( C::is_complete && C::is_inverted )
+  {
+    return countl_zero(wide<T,N>(svrbit_x(detail::sve_true<T>(),v)));
+  }
+  else
+  {
+    return svclz_m(alternative(cond, v, as(v)), cond.mask(as(v)), svrbit_x(detail::sve_true<T>(),v));
```
DenisYaroshevskiy, line 35: Is the alternative applied correctly here. Or is it backwards? Do we test alternative in counts

DenisYaroshevskiy (approved):

PR #1491, opened 2022-12-17
sve masked load issues

DenisYaroshevskiy:
solves a few simple cases at least


jfalcou:
if you tested locally, maybe we should put some load test back in the specific.json file ?

DenisYaroshevskiy:
> if you tested locally, maybe we should put some load test back in the specific.json file ?

I'm gonna fix the aggregated stuff first

jfalcou (approved):

PR #1494, opened 2022-12-19
different sve load

DenisYaroshevskiy:
This is theoretically a better loading strategy: https://stackoverflow.com/questions/74837019/arm-sve-svld1mask-ptr-vs-svldff1svptrue-ptr/74844397#74844397
Enabled memory tests (without compress ones) as well since they pass now


include/eve/detail/function/simd/arm/sve/load.hpp
```diff
@@ -22,7 +22,7 @@ template<relative_conditional_expr C,
 EVE_FORCEINLINE wide<T, N>
 load_(EVE_SUPPORTS(sve_),
   C const& cond, safe_type const& s, eve::as<wide<T, N>> const& tgt, Ptr p)
-  requires (!has_bundle_abi_v<wide<T, N>>)
+  requires (sve_abi<abi_t<T, N>>)
```
jfalcou, line 25: was n't the old test requried ? Is it what we do in other arch ?

DenisYaroshevskiy: sve_abi includes not bundle

jfalcou (approved):

PR #1500, opened 2022-12-22
Fix #1411 - Rework ARM all and any

### LINKED ISSUE #1411: [FEATURE] Arm, faster all/any
At .Net runtime they did this: https://github.com/dotnet/runtime/pull/75864
We should check what we do and then do the same as them, since they measured

jfalcou:
# Content
Modify all and any to use the trick from https://github.com/dotnet/runtime/pull/75864
Added a better path for ARMv7 following https://stackoverflow.com/a/41007755/737268

# Codegen

REFERENCE FROM https://github.com/dotnet/runtime/pull/75864
```c++
    uminp   v16.4s, v16.4s, v16.4s
    umov    x0, v16.d[0]
    cmn     x0, #1
    cset    x0, eq
```

BEFORE:
```c++
    uminv   s0, v0.4s
    fmov    w0, s0
    cmp     w0, 0
    cset    w0, ne
```

AFTER
```c++
    uminp   v0.4s, v0.4s, v0.4s
    fmov    x0, d0
    cmn     x0, #1
    cset    w0, eq
```

### DIFF
--- a/include/eve/module/core/regular/impl/simd/arm/neon/any.hpp
+++ b/include/eve/module/core/regular/impl/simd/arm/neon/any.hpp
@@ -43,6 +43,7 @@ any_(EVE_SUPPORTS(neon128_),
      logical<wide<T, N>> v0) noexcept requires std::same_as<abi_t<T, N>, arm_128_>
 {
   using u32_4 = typename wide<T, N>::template rebind<std::uint32_t, eve::fixed<4>>;
+  using u64_2 = typename wide<T, N>::template rebind<std::uint64_t, eve::fixed<2>>;
 
   if constexpr( C::is_complete && !C::is_inverted ) return false;
   // we still have to convert down here, so we can do it before ignore.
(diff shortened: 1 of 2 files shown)


include/eve/module/core/regular/impl/simd/arm/neon/any.hpp
```diff
@@ -55,14 +55,22 @@ any_(EVE_SUPPORTS(neon128_),
...
-  else // chars, no asimd
+  else // no asimd
   {
-    auto dwords = eve::bit_cast(v0, eve::as<u32_4>());
-    return eve::any[ignore_none](detail::to_logical(dwords));
+    auto mask = eve::bit_cast(v0, eve::as<u32_4>());
+    if constexpr( sizeof(T) > 2 )
+    {
+      // Adapted from https://stackoverflow.com/a/41007755/737268
+      return vget_lane_u64(vreinterpret_u64_u16(vqmovn_u32(mask)), 0) != 0;
```
DenisYaroshevskiy, line 68: We specifically not did this - because there is no 64 bit numbers. This a no asimd usecase.
Same as the other.

DenisYaroshevskiy (approved):

include/eve/module/core/regular/impl/simd/arm/neon/any.hpp
jfalcou: Rolled back

PR #1507, opened 2022-12-26
Fix #1294 - Implement minmax

### LINKED ISSUE #1294: [FEATURE] We need a minmax function. With predicate support ideally.
I wrote this:

```
return { eve::min(x,y), eve::max(x, y) };
```

But, unfortunately gcc does not optimise it properly.

https://godbolt.org/z/EaPG85KMx

jfalcou:


include/eve/module/core/regular/impl/simd/arm/neon/minmax.hpp
```diff
@@ -0,0 +1,30 @@
...
+namespace eve::detail
+{
+// -----  Regular
+template<arithmetic_scalar_value T, typename N>
+EVE_FORCEINLINE auto
+minmax_(EVE_SUPPORTS(neon128_), wide<T, N> v0, wide<T, N> v1) noexcept requires arm_abi<abi_t<T, N>>
+{
+  if constexpr(sizeof(T) == 8)
+  {
+    // There is no min/max for 64bits type, so we compute the check once
```
DenisYaroshevskiy, line 23: I thought you assume this is optimized. Otherwise you should do it everywhere.

include/eve/module/core/regular/impl/simd/x86/minmax.hpp
```diff
@@ -0,0 +1,43 @@
...
+{
+// -----  Regular
+template<arithmetic_scalar_value T, typename N>
+EVE_FORCEINLINE auto
+minmax_(EVE_SUPPORTS(sse2_), wide<T, N> v0, wide<T, N> v1) noexcept requires x86_abi<abi_t<T, N>>
+{
+  using enum category;
+  constexpr auto c = categorize<wide<T,N>>();
+
+  if constexpr(match(c, integer_)  && current_api == avx && N::value > 1)
```
DenisYaroshevskiy, line 25: This seems bad since the sse api will be targeted by this.

include/eve/module/core/regular/impl/simd/x86/minmax.hpp
```diff
@@ -0,0 +1,43 @@
...
+#include <eve/detail/apply_over.hpp>
+#include <eve/detail/category.hpp>
+#include <eve/detail/implementation.hpp>
+
+namespace eve::detail
+{
+// -----  Regular
+template<arithmetic_scalar_value T, typename N>
+EVE_FORCEINLINE auto
+minmax_(EVE_SUPPORTS(sse2_), wide<T, N> v0, wide<T, N> v1) noexcept requires x86_abi<abi_t<T, N>>
```
DenisYaroshevskiy, line 20: Why not just cpu.

include/eve/module/core/regular/impl/simd/arm/neon/minmax.hpp
jfalcou: I checked all codegen. Neon requires, SVE is fine, PPC is fine.

include/eve/module/core/regular/impl/simd/x86/minmax.hpp
jfalcou: How is SSE taking this as I asked for api to be == avx ?

include/eve/module/core/regular/impl/simd/x86/minmax.hpp
jfalcou: We usually let CPU be the full non-SIMD case.

include/eve/module/core/regular/impl/simd/x86/minmax.hpp
DenisYaroshevskiy: Api is avx. Abi is sse

jfalcou:
Let's have a platform::has_native_minmax

include/eve/module/core/regular/impl/simd/x86/minmax.hpp
jfalcou: Indeed

jfalcou:
I went for a more flexible way to ask if a given functions is implementaly usign intrinsics.
It is now only used for min/max and minmax but we can add it to other functions as time goes.

include/eve/arch/has_native_support.hpp
```diff
@@ -0,0 +1,20 @@
...
+
+#include <eve/arch/spec.hpp>
+#include <type_traits>
+
+namespace eve
+{
+  //! @brief Indicates if a given callable is natively supported for type T on a given API
+  template<typename Callable, typename T, typename API = current_api_type>
+  struct has_native_support : std::true_type
+  {};
```
DenisYaroshevskiy, line 18: Ooofff.. I think this is an overly generic thing that we won't be able to support properly.

DenisYaroshevskiy (commented):
I don't think the "has_" generic solution is good given how ghard it is to maintain for all the callables. Let's not explode our api even more.

Can you have like:

```
detail::minmax_ifelse
detail::minmax_native
detail::maxmin_native
```

and do the if/else chains? Or smth

jfalcou:
Reimplemented in a simpler way

include/eve/module/core/regular/impl/minmax.hpp
```diff
@@ -0,0 +1,91 @@
...
+minmax_(EVE_SUPPORTS(cpu_), C const& c, T0 v0, T1 v1, Ts... vs) noexcept
+    -> decltype(kumi::tuple {eve::min[c](v0, v1, vs...), eve::max[c](v0, v1, vs...)})
+{
+  return kumi::tuple {eve::min[c](v0, v1, vs...), eve::max[c](v0, v1, vs...)};
+}
+
+// -----  Tuple case
+template<kumi::non_empty_product_type Ts>
+EVE_FORCEINLINE auto
+minmax_(EVE_SUPPORTS(cpu_), Ts tup) noexcept -> decltype(kumi::tuple {eve::min(tup), eve::max(tup)})
```
DenisYaroshevskiy, line 67: This seems wrong. tuples never have a native.

include/eve/module/core/regular/impl/minmax.hpp
```diff
@@ -0,0 +1,91 @@
...
+#include <eve/module/core/regular/min.hpp>
+#include <eve/traits.hpp>
+
+namespace eve::detail
+{
+template<typename T>
+constexpr bool has_native_minmax() noexcept
+{
+  using e_t = element_type_t<T>;
+  constexpr bool small_int = !(std::integral<e_t> && sizeof(e_t) == 8);
```
DenisYaroshevskiy, line 23: This is hard to follow but failry obviously wrong since it never takes the number of elements into account

include/eve/module/core/regular/impl/minmax.hpp
```diff
@@ -0,0 +1,91 @@
...
+// -----  Predicate case
+template<typename Callable>
+EVE_FORCEINLINE auto
+minmax_(EVE_SUPPORTS(cpu_), Callable f)
+{
+  if constexpr( std::same_as<Callable, callable_is_less_> ) return eve::minmax;
+  else if constexpr( std::same_as<Callable, callable_is_greater_> )
+  {
+    return [](auto x, auto y) { return kumi::tuple {max(x, y), min(x, y)}; };
+  }
```
DenisYaroshevskiy, line 81: This is not great - call minmax and swap

include/eve/module/core/regular/impl/minmax.hpp
```diff
@@ -0,0 +1,91 @@
...
+  {
+    return kumi::tuple {eve::min(v0, v1, vs...), eve::max(v0, v1, vs...)};
+  }
+  else
+  {
+    // If there is no native min/max, we compute the check once
+    // We use > cause it is more often optimized than <
+    auto check = v0 > v1;
+    return kumi::tuple {if_else(check, v1, v0), if_else(check, v0, v1)};
+  }
```
DenisYaroshevskiy, line 52: Where are ...?

include/eve/module/core/regular/impl/minmax.hpp
jfalcou: there is tuple for min and max support now

include/eve/module/core/regular/impl/minmax.hpp
```diff
@@ -0,0 +1,90 @@
...
+template<typename W>
+constexpr bool has_native_minmax() noexcept
+{
+  if constexpr( scalar_value<W> ) return true;
+  else
+  {
+    using enum category;
+    constexpr bool is_ints64 = match(categorize<W>(), int64x4, uint64x4, int64x2, uint64x2);
+
+    if constexpr( x86_tag<current_api_type> )       return current_api == avx512 || !is_ints64;
```
DenisYaroshevskiy, line 29: avx does not have a native max. Agreed for a rename and comment

DenisYaroshevskiy (approved):

PR #1511, opened 2023-01-03
Use file globbing for easier CMakeLists.txt setup

jfalcou:


DenisYaroshevskiy (commented):
Is it maybe ossible to do the regular change separately?

jfalcou:
> Is it maybe ossible to do the regular change separately?

which one ?

PR #1523, opened 2023-01-22
Provides eve::construct as an extension point for non-trivial UDT ctor

jfalcou:
Some UDT (like the double double @jtlap is working on) has non trivial constructor that don't fit the classic product-type, member per member constructor stragtegy we use. We can't rely on the scalar constructor as calling it N times may be suboptimal in many cases.

This PR introduce construct, a dispatchable function that can take care of non-trivial UDT constructor. The default implementation is usign the same product-type startegy from before and we added a test checking for an arbitrary non trivial ctor on `udt::label_posiiton`.

### DIFF
--- a/include/eve/arch/cpu/wide.hpp
+++ b/include/eve/arch/cpu/wide.hpp
@@ -19,6 +19,7 @@
 #include <eve/detail/abi.hpp>
 #include <eve/detail/function/combine.hpp>
 #include <eve/detail/function/compounds.hpp>
+#include <eve/detail/function/construct.hpp>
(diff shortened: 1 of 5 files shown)


jfalcou:
A discussion can be made to move all wide constructor to construct calls as it would make the wide code less clutterred and woudl make fill() and make() redundant

DenisYaroshevskiy:
I really don't think we need this. We don't have the ability to do construct destroy and such. We just don't support incapsulation that's it.

jfalcou:
s discuss, let's not do that

PR #1527, opened 2023-01-28
Fix #1512 - Reorder concept so that logical scalar value can be used in tuple

### LINKED ISSUE #1512: [BUG] Product type containing logical<T> can't be used as wide type parameter
See: https://godbolt.org/z/nc8zMbMhd

jfalcou:


include/eve/concept/scalar.hpp
```diff
@@ -12,39 +12,98 @@
...
+constexpr bool is_plain() noexcept
 {
-  template<typename T>
-  constexpr bool is_plain() noexcept
-  {
-    return !(std::is_same_v<T, bool> || std::is_same_v<T, long double>)
+  return    !(std::is_same_v<T, bool> || std::is_same_v<T, long double>)
         &&  (std::is_floating_point_v<T> || std::is_integral_v<T>);
-  }
+}
```
DenisYaroshevskiy, line 24: this should be a concept I think. And don't we have it already?

include/eve/concept/scalar.hpp
```diff
@@ -12,39 +12,98 @@
...
+//! The concept `logical_scalar_value<T>` is satisfied if and only if T is an instance of
+//! eve::logical and its contents satisfy eve::plain_scalar_value
+//!
+//! ## Example Types
+//! - `logical<float>`
+//! - `logical<int>`
+//==================================================================================================
+template<typename T>
+concept logical_scalar_value = detail::instance_of<T, logical>
+                               && detail::is_plain<typename T::value_type>();
```
DenisYaroshevskiy, line 44: Yep, this should be plain scalar value

include/eve/concept/scalar.hpp
jfalcou: That's the trait we use to define the concept. It has been implemented and validated when we merged the last concept revision.

include/eve/concept/scalar.hpp
jfalcou: Fixed

DenisYaroshevskiy (approved):

PR #1545, opened 2023-02-09
Fix #1536 - Remove useless traits

### LINKED ISSUE #1536: [FEATURE] clean up expected_carinal (without _t, _v), pointer_cardinal (without _)
I don't think we need them. 
I thought about specialising, but I don't see a single case to specialize them tbh

jfalcou:


include/eve/memory/aligned_ptr.hpp
```diff
@@ -450,7 +450,7 @@ namespace eve
   template <typename T>
   EVE_FORCEINLINE auto previous_aligned_address(T* p) noexcept
   {
-    return previous_aligned_address(p, eve::expected_cardinal<std::remove_cvref_t<T>>{});
+    return previous_aligned_address(p, eve::expected<std::remove_cvref_t<T>>);
```
DenisYaroshevskiy, line 453: I think eve::exprected can also be removed, and eve_expected_cardinal_t used instead

DenisYaroshevskiy (approved):

PR #1563, opened 2023-02-15
Move SVE tests to full spec

jfalcou:


jfalcou:
Failures are tied to all and any not taking garbage value into account. Will fix this first then come back to those

PR #1571, opened 2023-02-23
Fix any behavior in SVE with garbage values

jfalcou:


include/eve/module/core/regular/impl/simd/arm/sve/any.hpp
```diff
@@ -9,14 +9,17 @@
...
 
 namespace eve::detail
 {
 template<scalar_value T, typename N, relative_conditional_expr C>
 EVE_FORCEINLINE bool
-any_(EVE_SUPPORTS(sve_), C const& cond, logical<wide<T,N>> const& v) noexcept
+any_(EVE_SUPPORTS(sve_), C const& cond, logical<wide<T,N>> v) noexcept
 requires sve_abi<abi_t<T, N>>
 {
+  v = remove_garbage(v);
```
DenisYaroshevskiy, line 21: I don't like this very much.
Like for `svptest_any` we can put this remove garbage mask directly there.

jfalcou: Updated

include/eve/module/core/regular/impl/simd/arm/sve/any.hpp
```diff
@@ -9,17 +9,22 @@
...
   {
-    if constexpr( C::is_inverted )  return svptest_any(sve_true<T>(), v);
+    if constexpr( C::is_inverted )
+    {
+      using ec_t  = expected_cardinal_t<T, typename logical<wide<T,N>>::abi_type>;
+      return svptest_any(keep_first(N::value).mask(as<wide<T,ec_t>>{}), v);
+    }
     else                            return false;
   }
   else return svptest_any(cond.mask(as<wide<T,N>>{}),v);
```
DenisYaroshevskiy, line 30: I think this should be

```
  // if not ignore all
      return svptest_any(sve_mask_for(cond, v), v);
```

or smth

jfalcou: turned into sve_true(v) as per discussion

DenisYaroshevskiy (approved):

include/eve/arch/arm/sve/sve_true.hpp
```diff
@@ -24,4 +24,22 @@ sve_true()
...
+// while masking potential garbage value
+template<relative_conditional_expr C, typename T>
+EVE_FORCEINLINE svbool_t
+sve_true(C cond, as<T> tgt)
+{
+  if constexpr(C::is_complete && C::is_inverted)
+  {
+    using v_t   = element_type_t<T>;
+    using ec_t  = expected_cardinal_t<v_t, typename T::abi_type>;
+    return keep_first(T::size()).mask(as<as_wide_t<v_t,ec_t>>{});
```
DenisYaroshevskiy, line 38: Ah - wait! Let's return sve_true() for just native register

include/eve/arch/arm/sve/sve_true.hpp
```diff
@@ -24,4 +24,24 @@ sve_true()
...
+// while masking potential garbage value
+template<relative_conditional_expr C, typename T>
+EVE_FORCEINLINE svbool_t
+sve_true(C cond, as<T> tgt)
+{
+  if constexpr(C::is_complete && C::is_inverted)
+  {
+    using v_t   = element_type_t<T>;
+    using ec_t  = expected_cardinal_t<v_t, typename T::abi_type>;
+
```
DenisYaroshevskiy, line 38: Ehy do you need abi here

DenisYaroshevskiy (approved):

include/eve/arch/arm/sve/sve_true.hpp
```diff
@@ -24,4 +24,24 @@ sve_true()
...
+    if constexpr(ec_t::value == T::size())      return sve_true<v_t>();
```
DenisYaroshevskiy, line 39: Hmm. I think this should actually be: `T::size() < native_cardinal_v<v_t>`

jfalcou: no, if it is smaller, then we got garbage. Remeber than SVE size are one at a time, so haiving SVE512 has no SVE256.

include/eve/arch/arm/sve/sve_true.hpp
jfalcou: Turns out nope

PR #1587, opened 2023-03-13
Fix expected_cardinal_v for non constexpr product_type in nested scenario

jfalcou:
Issue revealed by latest fluxion updates

### DIFF
--- a/include/eve/arch/expected_cardinal.hpp
+++ b/include/eve/arch/expected_cardinal.hpp
@@ -18,9 +18,9 @@ namespace eve::detail
   {
     if constexpr(kumi::product_type<Type>)
     {
-      return kumi::min_flat (  kumi::as_tuple_t<Type>{}
-                            , []<typename M>(M) { return ABI::template expected_cardinal<M>; }
-                            );
+      return kumi::min( kumi::result::flatten_all_t<kumi::as_tuple_t<Type>>{}
+                      , []<typename M>(M) { return ABI::template expected_cardinal<M>; }
+                      );
     }
     else
     {


include/eve/arch/expected_cardinal.hpp
```diff
@@ -18,9 +18,9 @@ namespace eve::detail
   {
     if constexpr(kumi::product_type<Type>)
     {
-      return kumi::min_flat (  kumi::as_tuple_t<Type>{}
-                            , []<typename M>(M) { return ABI::template expected_cardinal<M>; }
-                            );
+      return kumi::min( kumi::result::flatten_all_t<kumi::as_tuple_t<Type>>{}
```
DenisYaroshevskiy, line 21: Why not fix min_flatten if it's broken

jfalcou: The issue is that if T contains types with non constexpr constructor (which may happen), the `kumi::as_tuple_t<Type>{}` fails at compile time before flatten can do anything.

Doing it this way make sure we only constructs basic types and then call min.

jfalcou: we call a function, not a metafunction so we need an object there.

The old version was in fact a silent regression from before where we did this exactly. 
As we didn't updated the fluxion project since, the failure was silent.

jfalcou: Oh bleemy, no we where working on the type indeed.
Want me to make it easier on the eye by calling `kumi::result::min_flat_t<as_tupel_t<T>>` instead ?

PR #1590, opened 2023-03-15
Use GHCR docker registry instead of Dockerhub

jfalcou:
As docker wants my $$$, we move to GHCR.


DenisYaroshevskiy:
There is a doc somewhere too

jfalcou:
> There is a doc somewhere too

for ghcr ? Yeah found it at one point.
I am slowly fixing the script as it works quite differently once you have access to a container

PR #1592, opened 2023-03-23
Bit and byte reverse

jtlap:
three functions:
  bit_swap swap group of bits
  byte_reverse reverse  byte order
  bit reverse reverse bit order
  bit_reverse is useful for fft and can use byte_reverse and bit_swap


jfalcou (commented):
fix the doc and investigate intrinsics

DenisYaroshevskiy:
Are you sure it's not byte_swap? Is it little - big endian thing for bytes?

DenisYaroshevskiy:
Please then in the comment/ doc say byte_swap for searching ( i didn't look - maybe you already do)

test/unit/module/core/reverse.cpp
```diff
@@ -0,0 +1,77 @@
...
+  TTS_EXPR_IS(eve::reverse(T(), std::integral_constant<size_t, 1>()), T);
+//   TTS_EXPR_IS(eve::reverse(v_t()), v_t);
+//   TTS_EXPR_IS(eve::reverse(v_t(), std::integral_constant<size_t, 1>()), v_t);
+
+};
+
+template <typename T>
+void reverse_test(T x)
+{
+//   auto frx =  [&](size_t i, size_t size) { return x.get(size - i - 1); };
```
jfalcou, line 32: reemoove doOoOn't Comememement

jfalcou (changes requested):

test/unit/module/core/reverse.cpp
jtlap: ok

jfalcou:
The doc says: 

```c++
//!   @note eve::byte_reverse is the functional equivalent to std::byte_swap.
//!   However, we decided to name it byte_reverse in order to keep a proper naming
//!   scheme aligned with the eve::bit_swap/eve::bit_reverse functions.
```

jfalcou (approved):

jfalcou (approved):

PR #1603, opened 2023-05-14
FIX-1602: one pass min

DenisYaroshevskiy:
min_element that keeps track of the index. 
Ofc there can never be enough testing and there is some chance that our test infra does not cover something and we don't have fuzzing.

Specifically I am not sure that enough "repeats" covered in test cases, generic testing might not tackle this aggresively enough.

Regardless, I think it's in OK shape - let's review and merge if it's OK.

### DIFF
--- a/include/eve/module/algo/algo/traits.hpp
+++ b/include/eve/module/algo/algo/traits.hpp
@@ -256,7 +256,7 @@ namespace eve::algo
   //! @{
   //!   @var expensive_callable
   //!
-  //!   @brief *NOTE*: equivalent to no_aligning + no_unrolling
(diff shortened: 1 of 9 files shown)


include/eve/module/algo/algo/traits.hpp
```diff
@@ -266,6 +266,28 @@ namespace eve::algo
   //================================================================================================
   inline constexpr auto expensive_callable = ::rbr::flag( expensive_callable_tag{} );
 
+
+  struct one_pass_minmax_tag {};
```
DenisYaroshevskiy, line 270: Joel asked to name it "single_pass"

jfalcou (approved):

PR #1616, opened 2023-06-25
More shuffles

DenisYaroshevskiy:
Some 2 register set up. 

I still do not know how the proper multiregister shuffling will work, I imagine some sort of "shuffle compiler" but I do not know how that looks yet.

So far just working on the native form, added 2 register support so that we can do `_l1` shuffles next.

### DIFF
--- a/include/eve/detail/function/simd/common/canonical_shuffle_adapter.hpp
+++ b/include/eve/detail/function/simd/common/canonical_shuffle_adapter.hpp
@@ -12,32 +12,31 @@
 namespace eve::detail
 {
 
-template<typename Internal, simd_value T, std::ptrdiff_t G, std::ptrdiff_t... I>
-EVE_FORCEINLINE auto
-canonical_shuffle_adapter_impl_(EVE_SUPPORTS(cpu_), Internal, T, fixed<G>, pattern_t<I...>)
-requires(G > T::size())
-{
-  // NOTE: we catch this specifically because:
-  //  * sfinae failure here is most likely an error that the user will just miss.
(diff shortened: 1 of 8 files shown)


include/eve/detail/function/simd/common/canonical_shuffle_adapter.hpp
```diff
@@ -135,96 +135,125 @@ upscale_pattern(pattern_t<I...> p)
...
     auto next_pattern = []<std::size_t... i>(std::index_sequence<i...>) {
       return pattern_t<I..., ((void)i, we_)...> {};
     }(std::make_index_sequence<fundamental_cardinal_v<T> - sizeof...(I)> {});
-    return simplify_plain_shuffle1(cast, g, next_pattern);
+    using T1 = wide<T, fundamental_cardinal_t<T>>;
+    return simplify_plain_shuffle1(
+        next_pattern, g, bit_cast(x, as<T1> {}), bit_cast(xs, as<T1> {})...);
   }
-  else { return simplified_pattern {x, g, p}; }
+  else { return simplified_pattern {kumi::tuple {x, xs...}, g, p}; }
```
DenisYaroshevskiy, line 193: I should add the inversion somewhere. So that `0, 5, 2, 3' and '4, 1, 6, 7' have to be processed once.

DenisYaroshevskiy:
blend_epi16 fails in the weirdest of ways, and doesn't do that in docker. I will probably split that in a separate problem and ask you to have a look.

jfalcou (approved):

PR #1622, opened 2023-07-02
If no masked loads, clang->no undef warning, else -> maskz

### LINKED ISSUE #1619: [BUG] `Wuninitialized`
**Describe the bug**
GCC yields uninitialized warnings when we don't specify an `else` on the conditional, as `r_t that` is then returned by `src`.
https://github.com/jfalcou/eve/blob/cb6cc16813d38b9bea85765b6f38db39db6235ad/include/eve/module/core/regular/impl/simd/x86/load.hpp#L63-L68C21
However, Clang fails to optimize well if we do set an altnerate.

**To Reproduce**
Minimal example of the error:
https://godbolt.org/z/evoKqYTrf

Less minimal example of an error, showing Clang doing a fairly good job optimizing the example away:
https://godbolt.org/z/oo3zYW5sG
Swapping the comment to enable the alternate makes GCC happy, but Clang's codegen is now poor:
https://godbolt.org/z/vecoo554P

I haven't tested to confirm this, but I suspect using the `maskz` variants of the instructions in case of no conditionals would fix this problem. That is, use `_mm512_mask_loadu_pd` if there is an alternate, and `_mm512_maskz_loadu_pd` otherwise.
I could prepare a PR (in which case I could test).

chriselrod:
Fixes #1619

### DIFF
--- a/include/eve/arch/cpu/base.hpp
+++ b/include/eve/arch/cpu/base.hpp
@@ -9,13 +9,9 @@
 
 #include <eve/concept/vectorizable.hpp>
 #include <eve/detail/function/friends.hpp>
-#include <eve/detail/spy.hpp>
+#include <eve/detail/pragmas.hpp>
 
-#if defined(SPY_COMPILER_IS_GCC)
-#pragma GCC diagnostic push
-#pragma GCC diagnostic ignored "-Wuninitialized"
-#pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
-#endif
+EVE_ALLOW_UNINITIALIZED_VARIABLES_PRAGMA
 
 namespace eve::detail
 {
@@ -59,6 +55,4 @@ namespace eve::detail
   };
 }
 
-#if defined(SPY_COMPILER_IS_GCC)
-#pragma GCC diagnostic pop
-#endif
+EVE_RESTORE_ALLOW_UNINITIALIZED_VARIABLES_PRAGMA
--- a/include/eve/detail/pragmas.hpp
+++ b/include/eve/detail/pragmas.hpp
@@ -1,0 +1,28 @@
+#ifndef PRAGMAS_PRAGMA_HPP_INCLUDED
+#define PRAGMAS_PRAGMA_HPP_INCLUDED
+#include <eve/detail/spy.hpp>
+
+#if defined(SPY_COMPILER_IS_GCC)
+
+#  define EVE_ALLOW_UNINITIALIZED_VARIABLES_PRAGMA                                                 \
+    _Pragma("GCC diagnostic push") _Pragma("GCC diagnostic ignored \"-Wuninitialized\"")           \
+        _Pragma("GCC diagnostic ignored \"-Wmaybe-uninitialized\"")
+
+#  define EVE_RESTORE_ALLOW_UNINITIALIZED_VARIABLES_PRAGMA
+_Pragma("GCC diagnostic pop")
+
+#elif defined(SPY_COMPILER_IS_CLANG)
+
+#  define EVE_ALLOW_UNINITIALIZED_VARIABLES_PRAGMA                                                 \
+    _Pragma("clang diagnostic push") _Pragma("clang diagnostic ignored \"-Wuninitialized\"")
+
+#  define EVE_RESTORE_ALLOW_UNINITIALIZED_VARIABLES_PRAGMA _Pragma("clang diagnostic pop")
+
+#else
+
+#  define EVE_ALLOW_UNINITIALIZED_VARIABLES_PRAGMA
+#  define EVE_RESTORE_ALLOW_UNINITIALIZED_VARIABLES_PRAGMA
+
+#endif
+
+#endif
--- a/include/eve/module/core/regular/impl/simd/x86/load.hpp
+++ b/include/eve/module/core/regular/impl/simd/x86/load.hpp
@@ -14,6 +14,7 @@
 #include <eve/detail/category.hpp>
 #include <eve/detail/function/bit_cast.hpp>
 #include <eve/detail/function/to_logical.hpp>
+#include <eve/detail/pragmas.hpp>
 
 namespace eve::detail
 {
@@ -60,14 +61,8 @@ requires simd_compatible_ptr<Ptr, Pack> && (!has_bundle_abi_v<Pack>)
   else if constexpr( !std::is_pointer_v<Ptr> ) return load_(EVE_RETARGET(cpu_), cond, s, tgt, p);
   else if constexpr( C::is_complete ) return load_(EVE_RETARGET(cpu_), cond, s, tgt, p);
   else if constexpr( current_api >= avx512 )
-  {
-#if defined(__clang__)
-#  pragma clang diagnostic push
-#  pragma clang diagnostic ignored "-Wuninitialized"
-#else
-    if constexpr( C::has_alternative )
     {
-#endif
+EVE_ALLOW_UNINITIALIZED_VARIABLES_PRAGMA
     r_t  that;
     auto src = [&](auto const& s)
     {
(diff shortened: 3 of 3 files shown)


DenisYaroshevskiy:
Hi!

Thanks for doing this.

1. Tbh I don't think we should do separate implementation for clang/gcc. Let's keep the old one, unless there is a reason. That solves your optimization issue, right?

2. Turns out we already used these pramgas before: 
https://github.com/jfalcou/eve/blob/1857d74407f9c1dbd8e0678d6a9bfd47558ccbea/include/eve/arch/cpu/base.hpp#L14-L18

3. If you don't mind let's

* create a file `eve/detail/pramgas.hpp`
* That file should only depend on `#include <eve/detail/spy.hpp>`
* define "EVE_ALLOW_UNINITIALIZED_VARIABLES_PRAMGA"
   and "EVE_POP_PRAMGA" in there.

* Use that at the similar way to 
 https://github.com/jfalcou/eve/blob/1857d74407f9c1dbd8e0678d6a9bfd47558ccbea/include/eve/arch/cpu/base.hpp#L14-L18
In both places.

What do you think?

jfalcou:
Looks like I forgot to clean some GA files so the integration failure is normal.
I have pushed a PR to fix this so you will be able to rebase when it is merged.

jfalcou:
Could rebase from main so the Integration test can pass.

chriselrod:
> What do you think?

Sounds good, but --

> Tbh I don't think we should do separate implementation for clang/gcc. 

What about MSVC?
Should I keep a separate implementation for them, or assume that anyone who cares about runtime performance isn't using it?

DenisYaroshevskiy:
> What about MSVC?

I don't think we should have special MSVC code now.

At the moment eve does not compile for MSVC (insufficient C++20 support). Judging by what I've seen their intruction optimization story is poor throughout so having this specific place with a workaround seems not that useful compared to having extra code in there.

chriselrod:
> At the moment eve does not compile for MSVC (insufficient C++20 support).

I don't have any windows computers, so I've never tried MSVC myself, but my impression (from looking at cppreference's feature support tables) was that this was something they were actually good at:
https://en.cppreference.com/w/cpp/20
They're still far behind in c++23, though: https://en.cppreference.com/w/cpp/23
They seem much better than Apple Clang.

DenisYaroshevskiy:
> I don't have any windows computers, so I've never tried MSVC myself, but my impression (from looking at cppreference's feature support tables) was that this was something they were actually good at:

Unfortunately only on paper. Their implementation is very buggy. Once in a while someone attempts to get through the bugs of the latest release but no luck.
Apple clang supports less features but it supports them well enough for us to actually compile.

DenisYaroshevskiy (approved):
I appreciate you putting in the effort. I will merge once the CI passes.

chriselrod:
CI passed.

FWIW, it fails locally with
```
ninja unit/ 
[0/2] Re-checking globbed directories...
[1791/2534] Building CXX object test/unit/CMakeFiles/unit.meta.tag_invoke.exe.dir/meta/tag_invoke.cpp.o
FAILED: test/unit/CMakeFiles/unit.meta.tag_invoke.exe.dir/meta/tag_invoke.cpp.o 
/usr/bin/clang++  -I/home/chriselrod/Documents/progwork/cxx/simd/eve/test -I/home/chriselrod/Documents/progwork/cxx/simd/eve/examples -I/home/chriselrod/Documents/progwork/cxx/simd/eve/include -DEVE_NO_FORCEINLINE -march=skylake-avx512 -O3 -DNDEBUG -std=c++20 -Werror -Wall -Wpedantic -Wextra -Winvalid-pch -Xclang -include-pch -Xclang /home/chriselrod/Documents/progwork/cxx/simd/eve/build_arch/CMakeFiles/test_pch.dir/cmake_pch.hxx.pch -Xclang -include -Xclang /home/chriselrod/Documents/progwork/cxx/simd/eve/build_arch/CMakeFiles/test_pch.dir/cmake_pch.hxx -MD -MT test/unit/CMakeFiles/unit.meta.tag_invoke.exe.dir/meta/tag_invoke.cpp.o -MF test/unit/CMakeFiles/unit.meta.tag_invoke.exe.dir/meta/tag_invoke.cpp.o.d -o test/unit/CMakeFiles/unit.meta.tag_invoke.exe.dir/meta/tag_invoke.cpp.o -c /home/chriselrod/Documents/progwork/cxx/simd/eve/test/unit/meta/tag_invoke.cpp
/home/chriselrod/Documents/progwork/cxx/simd/eve/test/unit/meta/tag_invoke.cpp:109:24: error: redefinition of 'tag_invoke' as different kind of symbol
    friend std::string tag_invoke(eve::tag_of<func1> const&, auto, sort_of x)
                       ^
/home/chriselrod/Documents/progwork/cxx/simd/eve/test/unit/meta/tag_invoke.cpp:183:25: note: in instantiation of template class 'eve::sort_of<int>' requested here
  TTS_EQUAL( eve::func1(eve::sort_of{9})  , "Sort of 9"             );
                        ^
/home/chriselrod/Documents/progwork/cxx/simd/eve/include/eve/traits/invoke/tag_invoke.hpp:49:37: note: previous definition is here
  inline constexpr func_ns::invoker tag_invoke = {};
                                    ^
/home/chriselrod/Documents/progwork/cxx/simd/eve/test/unit/meta/tag_invoke.cpp:183:14: error: call to deleted function call operator in type 'const tags::func1'
  TTS_EQUAL( eve::func1(eve::sort_of{9})  , "Sort of 9"             );
             ^~~~~~~~~~
/home/chriselrod/Documents/progwork/cxx/simd/eve/test/tts/tts.hpp:1087:56: note: expanded from macro 'TTS_EQUAL'
#define TTS_EQUAL(LHS, RHS, ...)          TTS_RELATION(LHS,RHS, eq , "==" , "!=" , __VA_ARGS__)
                                                       ^~~
/home/chriselrod/Documents/progwork/cxx/simd/eve/test/tts/tts.hpp:1078:77: note: expanded from macro 'TTS_RELATION'
#define TTS_RELATION(A, B, OP, T, F, ...)     TTS_RELATION_ ## __VA_ARGS__ (A,B,OP,T,F)
                                                                            ^
/home/chriselrod/Documents/progwork/cxx/simd/eve/test/tts/tts.hpp:1079:65: note: expanded from macro 'TTS_RELATION_'
#define TTS_RELATION_(A, B, OP, T, F)         TTS_RELATION_IMPL(A,B,OP,T,F,TTS_FAIL)
                                                                ^
/home/chriselrod/Documents/progwork/cxx/simd/eve/test/tts/tts.hpp:1085:3: note: expanded from macro 'TTS_RELATION_IMPL'
}(A,B)                                                                                              \
  ^
/home/chriselrod/Documents/progwork/cxx/simd/eve/test/unit/meta/tag_invoke.cpp:35:35: note: candidate function [with T = <eve::sort_of<int>>] has been explicitly deleted
    unsupported_call<func1(T...)> operator()(T... x) const
                                  ^
/home/chriselrod/Documents/progwork/cxx/simd/eve/test/unit/meta/tag_invoke.cpp:29:10: note: candidate template ignored: substitution failure [with T = eve::sort_of<int>]: no matching function for call to object of type 'const func_ns::invoker'
    auto operator()(T const& x) const noexcept -> decltype(eve::tag_invoke(*this, x))
         ^                                                 ~~~
/home/chriselrod/Documents/progwork/cxx/simd/eve/test/unit/meta/tag_invoke.cpp:183:24: error: attempt to use a deleted function
  TTS_EQUAL( eve::func1(eve::sort_of{9})  , "Sort of 9"             );
                       ^
/home/chriselrod/Documents/progwork/cxx/simd/eve/test/unit/meta/tag_invoke.cpp:35:35: note: 'operator()<eve::sort_of<int>>' has been explicitly marked deleted here
    unsupported_call<func1(T...)> operator()(T... x) const
                                  ^
3 errors generated.
[1828/2534] Building CXX object test/unit/CMakeFiles/unit.a...ast_group.exe.dir/api/regular/swizzle/broadcast_group.cpp.o
ninja: build stopped: subcommand failed.
```

jfalcou:
That is a clang 16 issue. I have a fix for it I need to push.

PR #1653, opened 2023-08-23
compress copy, docs

DenisYaroshevskiy:
101 redesign of compress, in progress. Trying to accomodate both sparse and output to a different arrays.
So far - just  one implementation with some barebone tests


include/eve/conditional.hpp
```diff
@@ -160,7 +160,7 @@ namespace eve
 
     template<typename T> EVE_FORCEINLINE constexpr std::ptrdiff_t roffset(eve::as<T> const&) const
     {
-      return cardinal_v<T>;
+      return T::size();
```
jfalcou, line 163: why the change ?

PR #1669, opened 2023-09-05
compress copy simd

DenisYaroshevskiy:
This should, in theory, finish the implementation of compress_copy itself:
both scalar and simd versions are done and working.

Next patch would be to update the usage.


test/unit/api/compress/compress_store.cpp
```diff
@@ -0,0 +1,11 @@
+//==================================================================================================
+/**
+  EVE - Expressive Vector Engine
+  Copyright : EVE Project Contributors
+  SPDX-License-Identifier: BSL-1.0
+**/
+//==================================================================================================
+
+#include "test.hpp"
+
```
jfalcou, line 10: why the empty file then ?

DenisYaroshevskiy: I figured so that when one comes looking, the find this

jfalcou: well it's an empty target that just consume test compilation time no ?

jfalcou (approved):

test/unit/api/compress/compress_store.cpp
DenisYaroshevskiy: Agreed that I'll delete it in the next patch

PR #1674, opened 2023-09-10
copy_if

DenisYaroshevskiy:
eve::algo::copy_if.

Same as with `set_intersection` had to design with output range.

Here it meant an introduciton of a new iteration pattern:

do tails once.

Then duplicate  the main loop:

1. When there is enough space in the output for the whole registter.
2. When it's not.


jfalcou (commented):
LGTM, onl y detail is aren't the stage tests be in internal tests instead of int he public API test set ?
This is non blocking though.

jfalcou (approved):

PR #1678, opened 2023-09-19
Wide of 2 tuple of size 2 doesn't fail silently anymore

jfalcou:


DenisYaroshevskiy:
can we have a test please

jfalcou:
Turns out the issue is not EVE. We're gonna fix the other software

PR #1679, opened 2023-09-21
Better overload

jfalcou:
After some more discussions, this implements a 3rd version of our overload system.


include/eve/traits/overload/default_behaviors.hpp
```diff
@@ -0,0 +1,107 @@
...
+    struct behavior
+    {
+      template<typename T, typename O>
+      static constexpr auto call(auto, options<O> const& opts, eve::as<T> const& tgt)
+      requires( requires{ Tag::value(opts, tgt); } )
+      {
+        // we pass opts to value so each constant can handle their own option support
+        auto v = Tag::value(opts,tgt);
+        if constexpr(option_type_is<option::condition, O, ignore_none_>) return v;
+        else  return v & detail::expand_mask(opts[option::condition], tgt).mask();
```
DenisYaroshevskiy, line 103: This should remove condition from opts I think.

include/eve/traits/overload/protocol.hpp
```diff
@@ -0,0 +1,158 @@
...
+/**/
+
+//======================================================================================================================
+//! @addtogroup extensions
+//! @{
+//!   @def EVE_CALLABLE_NAMESPACE
+//!   @brief Register a namespace as suitable for containing eve::callable overloads.
+//! @}
+//======================================================================================================================
+#define EVE_CALLABLE_NAMESPACE()                                                                                       \
```
DenisYaroshevskiy, line 149: Why function like macro?

include/eve/traits/overload/protocol.hpp
```diff
@@ -0,0 +1,158 @@
...
+struct adl_helper_t {};                                                                                                \
+inline constexpr auto adl_helper = adl_helper_t {}                                                                     \
+/**/
+
+// Flag a function to support delayed calls on given architecture
+#define EVE_EXPECTS(ARCH) adl_helper_t const &, ARCH const &
```
DenisYaroshevskiy, line 155: EXPECTS sounds like contract. I'd say REQUIRES to match standard

include/eve/traits/overload/supports.hpp
```diff
@@ -0,0 +1,234 @@
...
+  //! @}
+  //====================================================================================================================
+  template <rbr::concepts::settings Settings>
+  struct options : Settings
+  {
+    template <rbr::concepts::option... Options>
+    constexpr EVE_FORCEINLINE explicit options(Options && ... opts) : Settings(EVE_FWD(opts) ...) {}
+
+    template <typename... Options>
+    constexpr EVE_FORCEINLINE options(rbr::settings<Options...> const& opts) : Settings(opts) {}
```
DenisYaroshevskiy, line 35: Any specific reason this is not explicit?

include/eve/traits/overload/supports.hpp
```diff
@@ -0,0 +1,234 @@
...
+    }
+  };
+
+  //====================================================================================================================
+  //! @addtogroup extensions
+  //! @{
+  //!   @var condition
+  //!   @brief Keyword for retrieving conditionals decorator
+  //!
+  //!   When a eve::decorated_callable is called with an option provided vi eve::conditional, the value of the
```
DenisYaroshevskiy, line 88: vi?

include/eve/traits/overload/supports.hpp
```diff
@@ -0,0 +1,234 @@
...
+  // Internal option carrying conditional mask or conditional expressions
+  //====================================================================================================================
+  struct condition_ : rbr::as_keyword<condition_>
+  {
+    using rbr::as_keyword<condition_>::operator=;
+    template<typename T> static constexpr bool check()
+    {
+      return conditional_expr<std::remove_cvref_t<T>> || logical_value<std::remove_cvref_t<T>>;
+    }
+  };
```
DenisYaroshevskiy, line 80: I would need relative only version.

include/eve/traits/overload/supports.hpp
```diff
@@ -0,0 +1,234 @@
...
+      else if constexpr(conditional_expr<O> || eve::logical_value<O>)
+      {
+        return (*this)[option::condition = o];
+      }
+      else if constexpr(std::same_as<O,bool>)
+      {
+        using type = std::conditional_t<std::same_as<bool,O>,std::uint8_t,O>;
+        return (*this)[option::condition = if_(logical<type>(o))];
+      }
+    }
```
DenisYaroshevskiy, line 124: If I undersand this correctly, I don't like it. 

tbh we never used `opt =` syntax, we always used tagged types.

I'd expect each "supports" to react with it's own option and update the common set of options.
Conditional should not be special

include/eve/traits/overload/supports.hpp
```diff
@@ -0,0 +1,234 @@
...
+      }
+      else if constexpr(conditional_expr<O> || eve::logical_value<O>)
+      {
+        return (*this)[option::condition = o];
+      }
+      else if constexpr(std::same_as<O,bool>)
+      {
+        using type = std::conditional_t<std::same_as<bool,O>,std::uint8_t,O>;
+        return (*this)[option::condition = if_(logical<type>(o))];
+      }
```
DenisYaroshevskiy, line 202: Same here - conditional should not be special

include/eve/traits/overload/supports.hpp
```diff
@@ -0,0 +1,234 @@
...
+  //!
+  //!   @godbolt{doc/traits/callable_supports.cpp}
+  //! @}
+  //====================================================================================================================
+  struct conditional
+  {
+    auto operator[](bool)                                                   -> void;
+    auto operator[](eve::logical_value auto const&)                         -> void;
+    auto operator[](eve::conditional_expr auto const&)                      -> void;
+    auto operator[](rbr::concepts::exactly<option::condition> auto const&)  -> void;
```
DenisYaroshevskiy, line 225: I think these can actually do things, no?

test/doc/traits/callable_constant.cpp
```diff
@@ -0,0 +1,35 @@
...
+    template<eve::floating_value T> auto call(as<T>) const -> T;
+
+    // A constant must provide a value() static member with two parameters:
+    //  * an eve::options pack containing potential decorators passed to the constant
+    //  * an eve::as instance to specify the output type
+    template<typename T> static constexpr auto value(auto, eve::as<T>) { return T{3.14159216}; }
+
+    EVE_CALLABLE_OBJECT(pi_t, pi_);
+  } inline constexpr Pi;
+};
```
DenisYaroshevskiy, line 19: I like thism except for capital Pi.

How about `pi_example`

test/doc/traits/callable_object.cpp
```diff
@@ -0,0 +1,40 @@
...
+    // Register this callable type to EVE overload system
+    EVE_CALLABLE_OBJECT(func_t, func_);
+  } inline constexpr func;
+};
+
+// As func_t used EVE_CALLABLE_OBJECT, we should write overloads in eve::detail
+namespace eve::detail
+{
+  auto func_(EVE_EXPECTS(cpu_), eve::integral_value auto x) {  return x*x;  }
+  auto func_(EVE_EXPECTS(cpu_), double x)                   {  return 1./x; }
```
DenisYaroshevskiy, line 24: You should show some actual platform. And I think these should accept options even if they are always empty, let's be consistent.

test/doc/traits/callable_specs.cpp
```diff
@@ -0,0 +1,51 @@
...
+}
+
+int main()
+{
+  using namespace eve::option;
+  std::cout << eve::func(1)                         << "\n";
+  std::cout << eve::func[precise](1)                << "\n";
+  std::cout << eve::func[scale = -.1](1)            << "\n";
+  std::cout << eve::func[scale = 10.][precise](1)   << "\n";
+  std::cout << eve::func[precise][scale = -1.5](1)  << "\n";
```
DenisYaroshevskiy, line 50: Is this what we do though? Because we use:

```
eve::func[eve::raw] 
```

and such

DenisYaroshevskiy:
FYI, this how I'd like "supports" to look

```

struct conditional_relative_in_out {
  template <rbr::concepts::settings Settings>
  EVE_FORCEINLINE
  static constexpr auto process_option(options<Settings> opt, eve::relative_conditional_expr auto ignore) 
  {
     if constexpr ( Settings::contains(out_ignore) ) { static_assert("Only supports 'in' and 'out' ignore"); }
     else if constexpr ( Settings::contains(in_ignore) ) { return default_to( opt, options(out_ignore = ignore) ); }
     else return default_to( opt, options(in_ignore = ignore) ); }
  }

  template <rbr::concepts::settings Settings>
  EVE_FORCEINLINE
  static constexpr auto defaults(options<Settings> opt) 
  {
     auto in = default_to(opt, options(in_ignore = ignore_none);
     return default_to(in, options(out_ignore = in[in_ignore]);
  }
};

```

include/eve/traits/overload/default_behaviors.hpp
```diff
@@ -0,0 +1,107 @@
...
+        else                                            return map(Tag{}, xs...);
+      }
+    };
+  };
+
+  //====================================================================================================================
+  //! @addtogroup extensions
+  //! @{
+  //!   @struct constant
+  //!   @brief CRTP base class imbuing a eve::callable with a constant function semantic
```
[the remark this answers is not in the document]
jfalcou: giving ?

include/eve/traits/overload/default_behaviors.hpp
jfalcou: Will do

include/eve/traits/overload/protocol.hpp
```diff
@@ -0,0 +1,158 @@
...
+EVE_FORCEINLINE constexpr auto operator()(Args&&... args) const                                                        \
+-> decltype(std::declval<TYPE>().call(std::declval<Args>()...))                                                        \
+requires( requires { std::declval<TYPE>().call(EVE_FWD(args)...); })                                                   \
+{                                                                                                                      \
+  if constexpr(eve::decorated_callable<TYPE,Args...>)                                                                  \
+    return TYPE::deferred_call(eve::current_api, eve::detail::defaults<TYPE,Args...>(), EVE_FWD(args)...);             \
+  else                                                                                                                 \
+    return TYPE::deferred_call(eve::current_api, EVE_FWD(args)...);                                                    \
+}                                                                                                                      \
```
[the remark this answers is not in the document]
jfalcou: 1/ I wanted to minimize the amount of changes we'll have to do on existing code.
2/ If a given callable supports nothing, why should I force us to handle an empty options ?

include/eve/traits/overload/protocol.hpp
jfalcou: It looks better in IDE and prevent syntax hightlighting fuck up

test/doc/traits/callable_constant.cpp
jfalcou: Will change

test/doc/traits/callable_object.cpp
jfalcou: 1/ will add a platform overload
2/ I am not sold on it, we can discuss that on slack

test/doc/traits/callable_specs.cpp
jfalcou: precise here is a made up raw like equivalent.
There is a bunch of math related options which are today badly named just because the old system was not able to carry values, having the full blown raberu support let us simplify those options but just having then be a key = value pair.

include/eve/traits/overload/supports.hpp
jfalcou: they sort have to. Rabery can not jsut take a random value in [] if it is not associated to a keyword. The condition keyword is also required internally later to be able to say "I want to fetch the condition from the option bundle".

Now, and it is related to the other remark on [], I was rthinking of having the operator[] in condition to do this and generally have people write code to how to process values insiode a option bundle.

include/eve/traits/overload/supports.hpp
jfalcou: See my other remark

include/eve/traits/overload/protocol.hpp
jfalcou: I'm OK with that

include/eve/traits/overload/supports.hpp
jfalcou: I just forgot I guess

include/eve/traits/overload/supports.hpp
jfalcou: via I guess

include/eve/traits/overload/supports.hpp
jfalcou: We used tagged type except for a bunch of place where the option is indeed a single value defined as a `some_kw = the value`. See unroll for example.

I need to wrap the logical/conditional into a rbr keyword, that's the API opf RBR. I don't think addign all the bagage of rbr::keyword to logical and all other thing is a good idea.

jfalcou:
> FYI, this how I'd like "supports" to look
> 
> ```
> 
> struct conditional_relative_in_out {
>   template <rbr::concepts::settings Settings>
>   EVE_FORCEINLINE
>   static constexpr auto process_option(options<Settings> opt, eve::relative_conditional_expr auto ignore) 
>   {
>      if constexpr ( Settings::contains(out_ignore) ) { static_assert("Only supports 'in' and 'out' ignore"); }
>      else if constexpr ( Settings::contains(in_ignore) ) { return default_to( opt, options(out_ignore = ignore) ); }
>      else return default_to( opt, options(in_ignore = ignore) ); }
>   }
> 
>   template <rbr::concepts::settings Settings>
>   EVE_FORCEINLINE
>   static constexpr auto defaults(options<Settings> opt) 
>   {
>      auto in = default_to(opt, options(in_ignore = ignore_none);
>      return default_to(in, options(out_ignore = in[in_ignore]);
>   }
> };
> ```

what's `default_to` here ?

include/eve/traits/overload/supports.hpp
DenisYaroshevskiy: Not really.  My problem in general with this design is that you taking away the logic from modifiers, give it back.

Modifiers should be the thing that determines what `operator[]` is provided.
Modifiers should be in charge of how to process them.

If I need to support a modifier of `int` - so be it.

Take options, return new options. See my `conditional_relative_in_out` example.

test/doc/traits/callable_specs.cpp
DenisYaroshevskiy: I'm not sold on keywords being the thing that is passed in `operator[]`.
I'm also not sold on passing parameters through `scale = 0.5` - I'd do `eve::scale(0.5)`

That way you also don't need `using namespace options`.

jfalcou: Fixed

include/eve/traits/overload/default_behaviors.hpp
```diff
@@ -0,0 +1,107 @@
...
+    static constexpr auto call(Arch const& arch, Args&&... args) requires(supports_default_behavior<T,Arch,Args...>)
+    {
+      return T::behavior::call(arch,EVE_FWD(args)...);
+    }
+
+    template<typename Arch, typename... Args>
+    EVE_FORCEINLINE
+    static constexpr auto call(Arch const&, Args&&...) requires(!supports_default_behavior<T,Arch,Args...>)
+    {
+      static_assert(supports_default_behavior<T, Arch, Args...>, "[EVE] - Incomplete callable object implementation");
```
DenisYaroshevskiy, line 46: I don't understand this error message I think

jfalcou: If you ends up there, it means you have given a callable a defautl behavior but you didnt implemented the correct function in it. So instead of ending up with  a brutal hard error, we notify the user its default behavior is incomplete.

I am open to get a better message

jfalcou:
Applied most of the discussed changes.
Names etc still open if you want to have a 2nd go

include/eve/traits/overload/supports.hpp
DenisYaroshevskiy: Still not explicit

include/eve/traits/overload/supports.hpp
```diff
@@ -0,0 +1,268 @@
...
+//======================================================================================================================
+namespace eve::detail
+{
+  //====================================================================================================================
+  //  Delay the access to defaults() to after the proper operator() has been selected.
+  //  Also returns an empty set of options by default
+  //====================================================================================================================
+  template<typename T, typename...> EVE_FORCEINLINE constexpr auto defaults()
+  {
+    if constexpr(requires { T::defaults(); }) return T::defaults();
```
DenisYaroshevskiy, line 56: The discussion was the defaults is :

```
T::defaults(old_options) -> options
```

You need access to options., for example compress staff needs

include/eve/traits/overload/supports.hpp
```diff
@@ -0,0 +1,268 @@
...
+  template<typename Tag, typename Settings, typename... Specs>
+  struct decorated_fn : Specs...
+  {
+    protected:
+    using Specs::process_option...;
+
+    public:
+    using callable_tag_type = void;
+
+    decorated_fn(Settings s) : opts(std::move(s)) {}
```
DenisYaroshevskiy, line 115: explicit?

include/eve/traits/overload/supports.hpp
```diff
@@ -0,0 +1,268 @@
...
+      auto new_opts = process_option(opts, o);
+      return decorated_fn<Tag,decltype(new_opts),Specs...>{new_opts};
+    }
+
+    // Behave as a callable but pass the current bundle of options to the deferred_call handler
+    template<typename... Args>
+    EVE_FORCEINLINE constexpr auto operator()(Args&&... args) const
+    -> decltype(std::declval<Tag>().call(std::declval<Args>()...))
+    {
+      return Tag::deferred_call(eve::current_api, opts, EVE_FWD(args)...);
```
DenisYaroshevskiy, line 132: Where is defaults?

include/eve/traits/overload/supports.hpp
```diff
@@ -0,0 +1,268 @@
...
+    auto process_option(auto const& base, rbr::concepts::exactly<condition> auto opt) const
+    {
+      return options{rbr::merge(options{opt}, base)};
+    }
+
+    template<std::same_as<bool> O>
+    auto process_option(auto const& base, O opt) const
+    {
+      // Just delay the evaluation of the type by injecting some templates
+      using type = std::conditional_t<std::same_as<bool,O>,std::uint8_t,O>;
```
DenisYaroshevskiy, line 222: Confused as to why

include/eve/traits/overload/supports.hpp
```diff
@@ -0,0 +1,268 @@
...
+  //!
+  //!   eve::conditional is an option specification that can be used when defining a eve::callable type to make it
+  //!   supports decoration via `bool`, eve::logical, any eve::conditional_expr or any eve::relative_conditional_expr.
+  //!
+  //!   @groupheader{Example}
+  //!
+  //!   @godbolt{doc/traits/callable_supports.cpp}
+  //! @}
+  //====================================================================================================================
+  struct conditional
```
DenisYaroshevskiy, line 211: I think the word `support` here would be good somewhere, otherwise confusing.

include/eve/traits/overload/supports.hpp
```diff
@@ -0,0 +1,268 @@
...
+  //! @{
+  //!   @var condition
+  //!   @brief Keyword for retrieving conditionals decorator
+  //!
+  //!   When a eve::decorated_callable is called with an option provided via eve::conditional, the value of the
+  //!   conditional can be retrieved using eve::condition
+  //!   @see eve::supports
+  //! @}
+  //====================================================================================================================
+  inline constexpr detail::condition_ condition = {};
```
DenisYaroshevskiy, line 98: Let's add smth like the word `key` or `option` to it, too many things want to be `condition`

test/doc/traits/callable_specs.cpp
```diff
@@ -0,0 +1,56 @@
...
+#include <type_traits>
+#include <eve/traits/overload.hpp>
+#include <eve/wide.hpp>
+
+namespace eve
+{
+  // Defines two RABERU flags
+  using namespace rbr::literals;
+  inline constexpr auto precise = "precise"_fl;
+  inline constexpr auto scale   = "scale"_fl;
```
DenisYaroshevskiy, line 12: let's use structs, no?

include/eve/traits/overload/default_behaviors.hpp
```diff
@@ -0,0 +1,107 @@
...
+      }
+    };
+  };
+
+  //====================================================================================================================
+  //! @addtogroup extensions
+  //! @{
+  //!   @struct constant
+  //!   @brief CRTP base class imbuing a eve::callable with a constant function semantic
+  //!
```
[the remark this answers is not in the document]
jfalcou: It is somehow geenrated by Doxygen now:

image

include/eve/traits/overload/default_behaviors.hpp
jfalcou: this class is an intermediate steps that gather all possible defautl behavior (like being a constant, or being an elementwise operation) and provide an implementation so we don't have to overload soem familly of functions 1000x times doing the same thing. See the constant example.

include/eve/traits/overload/protocol.hpp
```diff
@@ -0,0 +1,163 @@
...
+//!   @brief Macros, classes and functions designed to extend EVE externally.
+//! @}
+//======================================================================================================================
+
+namespace eve
+{
+  //====================================================================================================================
+  //! @addtogroup extensions
+  //! @{
+  //!   @concept callable
```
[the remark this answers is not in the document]
jfalcou: Same

image

include/eve/traits/overload/protocol.hpp
```diff
@@ -0,0 +1,163 @@
...
+  template<typename T, typename...>
+  concept callable = requires(T const&) { typename T::callable_tag_type; };
+}
+
+//======================================================================================================================
+//! @addtogroup extensions
+//! @{
+//!   @def   EVE_CALLABLE_OBJECT_FROM
+//!   @brief Generate the generic function interface for any EVE-compatible @callable
+//!
```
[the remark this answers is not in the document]
jfalcou: Same it is autogenerated

include/eve/traits/overload/supports.hpp
jfalcou: If I put logical<std::uint8_t> here directly, it instanciates stuff too early.

test/doc/traits/callable_specs.cpp
jfalcou: I wanted to keep the example focused on the support thing, the fact we use the fast API of RABERU is a detail.

include/eve/traits/overload/supports.hpp
jfalcou: Will do

include/eve/traits/overload/supports.hpp
jfalcou: in protocol.hpp

```c++
#define EVE_CALLABLE_OBJECT_FROM(NS,TYPE,NAME)                                                                         \
template<typename... Args>                                                                                             \
EVE_FORCEINLINE constexpr auto operator()(Args&&... args) const                                                        \
-> decltype(std::declval<TYPE>().call(std::declval<Args>()...))                                                        \
requires( requires { std::declval<TYPE>().call(EVE_FWD(args)...); })                                                   \
{                                                                                                                      \
  return TYPE::deferred_call(eve::current_api, eve::detail::defaults<TYPE,Args...>(), EVE_FWD(args)...);               \
}                                                                                                                      \
```

Where do you want it to be ?

include/eve/traits/overload/supports.hpp
jfalcou: I woudl prefer `conditional_options` so it 'll read

```c++
struct foo_t : supports<conditional_options>
```

include/eve/traits/overload/protocol.hpp
```diff
@@ -0,0 +1,163 @@
...
+template<typename... Args>                                                                                             \
+EVE_FORCEINLINE constexpr auto operator()(Args&&... args) const                                                        \
+-> decltype(std::declval<TYPE>().call(std::declval<Args>()...))                                                        \
+requires( requires { std::declval<TYPE>().call(EVE_FWD(args)...); })                                                   \
+{                                                                                                                      \
+  return TYPE::deferred_call(eve::current_api, eve::detail::defaults<TYPE,Args...>(), EVE_FWD(args)...);               \
+}                                                                                                                      \
+static EVE_FORCEINLINE decltype(auto) deferred_call(auto arch, auto&&...args) noexcept                                 \
+requires(requires { NAME(NS::adl_helper, arch, EVE_FWD(args)...); })                                                   \
```
DenisYaroshevskiy, line 80: This does not need sfinae - you sfinae based on signature

include/eve/traits/overload/protocol.hpp
```diff
@@ -0,0 +1,163 @@
...
+{                                                                                                                      \
+  return NAME(NS::adl_helper, arch, EVE_FWD(args)...);                                                                 \
+}                                                                                                                      \
+template<typename... Args>                                                                                             \
+static EVE_FORCEINLINE decltype(auto) deferred_call(auto arch, Args&&...args) noexcept                                 \
```
DenisYaroshevskiy, line 85: Don't overload here - since you don't need sfinae - use if constexpr.

include/eve/traits/overload/protocol.hpp
```diff
@@ -0,0 +1,163 @@
...
+{                                                                                                                      \
+  return NAME(eve::detail::adl_helper, arch, EVE_FWD(args)...);                                                        \
+}                                                                                                                      \
+template<typename... Args>                                                                                             \
+static EVE_FORCEINLINE decltype(auto) deferred_call(auto arch, Args&&...args) noexcept                                 \
+requires(!requires { NAME(eve::detail::adl_helper, arch, EVE_FWD(args)...); })                                         \
+{                                                                                                                      \
+  return eve::detail::default_behavior<TYPE,Args...>::call(arch, EVE_FWD(args)...);                                    \
+}                                                                                                                      \
```
DenisYaroshevskiy, line 143: everything from the above comment

include/eve/traits/overload/default_behaviors.hpp
DenisYaroshevskiy: I don't see why you need this - look at my overload comments

include/eve/traits/overload/protocol.hpp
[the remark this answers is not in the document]
DenisYaroshevskiy: eve/module/core.hpp

include/eve/traits/overload/supports.hpp
DenisYaroshevskiy: Here - you populate options here - fil in the defaults too. I think we don't yet agree on what defaults are

include/eve/traits/overload/supports.hpp
```diff
@@ -0,0 +1,268 @@
...
+  //!
+  //!   eve::conditional is an option specification that can be used when defining a eve::callable type to make it
+  //!   supports decoration via `bool`, eve::logical, any eve::conditional_expr or any eve::relative_conditional_expr.
+  //!
+  //!   @groupheader{Example}
+  //!
+  //!   @godbolt{doc/traits/callable_supports.cpp}
+  //! @}
+  //====================================================================================================================
+  struct conditional_options
```
DenisYaroshevskiy, line 211: the `s` here confuses me. I'd say `conditional_option`

include/eve/traits/overload/supports.hpp
DenisYaroshevskiy: Ah - like logical<std::uint8_t>? ok

include/eve/traits/overload/default_behaviors.hpp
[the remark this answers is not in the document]
jfalcou: OK gotcha

include/eve/traits/overload/protocol.hpp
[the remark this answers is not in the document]
jfalcou: OK

include/eve/traits/overload/supports.hpp
jfalcou: Fixed

include/eve/traits/overload/protocol.hpp
```diff
@@ -0,0 +1,163 @@
...
+requires(requires { NAME(NS::adl_helper, arch, EVE_FWD(args)...); })                                                   \
+{                                                                                                                      \
+  return NAME(NS::adl_helper, arch, EVE_FWD(args)...);                                                                 \
+}                                                                                                                      \
+template<typename... Args>                                                                                             \
+static EVE_FORCEINLINE decltype(auto) deferred_call(auto arch, Args&&...args) noexcept                                 \
+requires(!requires { NAME(NS::adl_helper, arch, EVE_FWD(args)...); })                                                  \
+{                                                                                                                      \
+  return eve::detail::default_behavior<TYPE,Args...>::call(arch, EVE_FWD(args)...);                                    \
```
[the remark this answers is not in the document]
jfalcou: Resolved as per discussion

include/eve/traits/overload/default_behaviors.hpp
```diff
@@ -0,0 +1,90 @@
...
+  template<typename Tag> struct elementwise
+  {
+    struct behavior
+    {
+      template<typename... Ts> static constexpr auto process(auto, Ts const&... xs)
+      requires( simd_value<Ts> || ...)
+      {
+        if constexpr((has_aggregated_abi_v<Ts> || ...)) return aggregate(Tag{}, xs...);
+        else                                            return map(Tag{}, xs...);
+      }
```
DenisYaroshevskiy, line 42: Where is `Tag::deferred_call`

include/eve/traits/overload/default_behaviors.hpp
```diff
@@ -0,0 +1,90 @@
...
+    struct behavior
+    {
+      template<typename T, typename O>
+      static constexpr auto process(auto, options<O> const& opts, eve::as<T> const& tgt)
+      requires( requires{ Tag::value(opts, tgt); } )
+      {
+        // we pass opts to value so each constant can handle their own option support
+        auto v = Tag::value(opts,tgt);
+        if constexpr(option_type_is<condition_key, O, ignore_none_>) return v;
+        else  return v & detail::expand_mask(opts[condition_key], tgt).mask();
```
DenisYaroshevskiy, line 86: this misses else_

include/eve/traits/overload/supports.hpp
```diff
@@ -0,0 +1,298 @@
...
+
+  //====================================================================================================================
+  // Internal option carrying conditional mask or conditional expressions
+  //====================================================================================================================
+  struct condition_key_t : rbr::as_keyword<condition_key_t>
+  {
+    using rbr::as_keyword<condition_key_t>::operator=;
+    template<typename T> static constexpr bool check()
+    {
+      return conditional_expr<std::remove_cvref_t<T>> || logical_value<std::remove_cvref_t<T>>;
```
DenisYaroshevskiy, line 86: this seems unnessesary. The `process_option` has to do the checks

include/eve/traits/overload/supports.hpp
```diff
@@ -0,0 +1,298 @@
...
+  //!   @see eve::supports
+  //! @}
+  //====================================================================================================================
+  inline constexpr detail::condition_key_t condition_key = {};
+}
+
+namespace eve::detail
+{
+  //====================================================================================================================
+  // Intermediate type for handling chains of option like in func[opt1 = y][opt2 = x][ignore](...)
```
DenisYaroshevskiy, line 116: we don't really write options like this in eve, looks confusing I think

include/eve/traits/overload/supports.hpp
```diff
@@ -0,0 +1,298 @@
...
+    }
+
+    auto process_option(auto const& base, eve::conditional_expr auto opt) const
+    {
+      return process_option(base, condition_key = opt);
+    }
+
+    /// Default settings of eve::conditional is eve::ignore_none
+    EVE_FORCEINLINE static constexpr auto defaults() noexcept { return options{condition_key = ignore_none};  }
+  };
```
DenisYaroshevskiy, line 262: hmm, do we want to just inherit relative_condition_option?

test/doc/traits/callable_object.cpp
```diff
@@ -0,0 +1,40 @@
...
+
+    // Register this callable type to EVE overload system
+    EVE_CALLABLE_OBJECT(func_t, func_);
+  } inline constexpr func;
+};
+
+// As func_t used EVE_CALLABLE_OBJECT, we should write overloads in eve::detail
+namespace eve::detail
+{
+  auto func_(EVE_REQUIRES(cpu_), auto, eve::integral_value auto x) {  return x*x;  }
```
DenisYaroshevskiy, line 23: auto is options? Can you specify that?

test/doc/traits/callable_specs.cpp
```diff
@@ -0,0 +1,55 @@
...
+  {
+    auto call(int) -> double;
+    EVE_CALLABLE_OBJECT(func_t, func_);
+  } inline constexpr func;
+};
+
+namespace eve::detail
+{
+  template<typename D>
+  auto func_(EVE_REQUIRES(cpu_), eve::options<D> const& opt, int x)
```
DenisYaroshevskiy, line 40: Seems like we want a concept for options, what do you think? This D looks out of place.
Up to you, I don't care very much

test/doc/traits/callable_specs.cpp
DenisYaroshevskiy: literals are much more confusing then just a struct

include/eve/traits/overload/default_behaviors.hpp
jfalcou: elementiwse is not complete yet

include/eve/traits/overload/default_behaviors.hpp
jfalcou: Good catch

include/eve/traits/overload/supports.hpp
jfalcou: no. That is not related to that.
Raberu asks for option level check to prevent key = bullshti to occurs.
Now, we indeed never let ppl use condition_key, we cna maybe remove it but it is unrelated, I need you to understadn this point.

include/eve/traits/overload/supports.hpp
jfalcou: Oh yes, I forgot to change the doc here

include/eve/traits/overload/supports.hpp
jfalcou: I was thinking about it but wasn't sure.

test/doc/traits/callable_object.cpp
jfalcou: it'll be somethign like `rbr::concepts::settings auto`

test/doc/traits/callable_specs.cpp
jfalcou: the concept is in rbr. I can propagate it to eve so it doesnt look horrible

include/eve/traits/overload/supports.hpp
DenisYaroshevskiy: I understand. We just need a compile time map, that's all we use from raberu. Otherwise our usecase is different. This doesn't do anything

jfalcou: Sure

include/eve/traits/overload/supports.hpp
jfalcou: Fixed

include/eve/traits/overload/default_behaviors.hpp
```diff
@@ -0,0 +1,157 @@
...
+
+#include <eve/concept/value.hpp>
+#include <eve/detail/skeleton.hpp>
+#include <eve/detail/function/conditional.hpp>
+
+namespace eve
+{
+  namespace detail
+  {
+    // Usual object function so we can have EVE_FORCEINLINE there for QoI on ARM/AARCH64
```
DenisYaroshevskiy, line 18: I don't think this comment is meaningful to me. Do you want to maybe remove this comment?

include/eve/traits/overload/default_behaviors.hpp
```diff
@@ -0,0 +1,157 @@
...
+    inline struct { EVE_FORCEINLINE auto operator()(auto, auto x) const { return x; } } eval_2nd = {};
```
DenisYaroshevskiy, line 19: constexpr eval_2nd;

Maybe return_2nd?

include/eve/traits/overload/default_behaviors.hpp
```diff
@@ -0,0 +1,157 @@
...
+
+    template<typename T>
+    auto operator[](T t) const
+    requires( requires(base const& b) { b[t];} )
+    {
+      auto new_traits = base::operator[](t);
+      return  Func<decltype(new_traits)>{new_traits};
+    }
+
+    template<typename T> void operator[](T t) const =delete;
```
DenisYaroshevskiy, line 41: Can you add a comment that this only helps with error messages.
Also - does it? I remeber we tried but I don't remember the results.

include/eve/traits/overload/default_behaviors.hpp
```diff
@@ -0,0 +1,157 @@
...
+  template< template<typename> class Func
+          , typename OptionsValues
+          , typename... Options
+          >
+  struct callable : decorated_with<OptionsValues, Options...>
+  {
+    using base = decorated_with<OptionsValues, Options...>;
+
+    template<typename T>
+    auto operator[](T t) const
```
DenisYaroshevskiy, line 34: constexpr maybe?

include/eve/traits/overload/default_behaviors.hpp
```diff
@@ -0,0 +1,157 @@
...
+  //====================================================================================================================
+
+  //====================================================================================================================
+  template< template<typename> class Func
+          , typename OptionsValues
+          , typename... Options
+          >
+  struct elementwise_callable : callable<Func, OptionsValues, conditional_option, Options...>
+  {
+    Func<OptionsValues> const& current() const { return static_cast<Func<OptionsValues>>(*this); }
```
DenisYaroshevskiy, line 59: I thought it's usually called `self` or `derived_self`.
Aslo - maybe put this into `callable`?
Also - constexpr

include/eve/traits/overload/default_behaviors.hpp
```diff
@@ -0,0 +1,157 @@
...
+
+    template<callable_options O, typename T, typename... Ts>
+    constexpr auto behavior(auto arch, O const& opts, T x0,  Ts const&... xs) const
```
DenisYaroshevskiy, line 62: Should we fwd xs?
Or maybe by value everything?

include/eve/traits/overload/default_behaviors.hpp
```diff
@@ -0,0 +1,157 @@
...
+        auto const cond     = opts[condition_key];
+
+        // Drop the conditional and process the function call with the universal masker
+        auto const rmv_cond = options{rbr::drop(condition_key, opts)};
+        auto const f        = Func<decltype(rmv_cond)>{rmv_cond};
+
+        // If the conditional call is supported, call it
+        if      constexpr( is_supported ) return detail::mask_op(cond, f, x0, xs...);
+        // if not, call the non-masked version then mask piecewise
+        else if constexpr( any_simd )     return detail::mask_op(cond, detail::eval_2nd, x0, f(x0,xs...));
```
DenisYaroshevskiy, line 92: Why not just if_else?

include/eve/traits/overload/default_behaviors.hpp
```diff
@@ -0,0 +1,157 @@
...
+      if constexpr( requires{ func_t::deferred_call(arch, opts, target); } )
+      {
+        return func_t::deferred_call(arch, opts, target);
+      }
+      else
+      {
+        // Compute the raw constant
+        auto const constant_value = [&]()
+        {
+          if      constexpr( requires{ Func<OptionsValues>::value(opts, target); } )
```
DenisYaroshevskiy, line 145: Why is this check? Why not just always one or the other?
Let's do just 2nd and if needed just 1st.

include/eve/traits/overload/default_behaviors.hpp
```diff
@@ -0,0 +1,157 @@
...
+            return Func<OptionsValues>::value(opts, target);
+          else if constexpr( requires{ Func<OptionsValues>::value(target); } )
+            return Func<OptionsValues>::value(target);
+        }();
+
+        // Apply a mask if any and replace missing values with 0 if no alternative is provided
+        if constexpr(option_type_is<condition_key, O, ignore_none_>) return constant_value;
+        else  return detail::mask_op(opts[condition_key], detail::eval_2nd, 0, constant_value);
```
DenisYaroshevskiy, line 153: again - if_else?

include/eve/traits/overload/supports.hpp
```diff
@@ -0,0 +1,267 @@
...
+{
+  //====================================================================================================================
+  //! @addtogroup extensions
+  //! @{
+  //!   @var functor
+  //!   @brief EVE's @callable generator
+  //! @}
+  //====================================================================================================================
+  template<template<typename> class Func>
+  inline constexpr auto functor = Func<eve::options<>>{};
```
DenisYaroshevskiy, line 66: What is this guy?

include/eve/traits/overload/supports.hpp
```diff
@@ -0,0 +1,267 @@
...
+    decltype(auto) options() const
+    {
+      return kumi::fold_left( [&](auto acc, auto const& m) { return m.default_to(acc); }
+                            , kumi::tuple<Options...>{}
+                            , self()
+                            );
+    }
+
+    // Short-cut to current state values
+    decltype(auto) self() const { return static_cast<OptionsValues const&>(*this); }
```
DenisYaroshevskiy, line 135: Seems like not necessary and is a bit weird. Like - will my callable have `self()` method?

include/eve/traits/overload/supports.hpp
```diff
@@ -0,0 +1,267 @@
...
+  //!   type to make it supports decoration via  any eve::relative_conditional_expr.
+  //!
+  //!   @groupheader{Example}
+  //!
+  //!   @godbolt{doc/traits/callable_supports.cpp}
+  //! @}
+  //====================================================================================================================
+  struct relative_conditional_option
+  {
+    auto process(auto const& base, eve::relative_conditional_expr auto opt) const
```
DenisYaroshevskiy, line 196: noexcept if you marked the other noexcept? Or like nothing.
And constexpr.

include/eve/traits/overload/supports.hpp
```diff
@@ -0,0 +1,267 @@
...
+    EVE_FORCEINLINE constexpr auto default_to(auto const& base) const noexcept
+    {
+      auto new_opts = rbr::merge(base, options{condition_key = ignore_none});
+      return options<decltype(new_opts)>{new_opts};
+    }
+  };
+
+  /// Checks if the type associated to a given Keyword in a Option pack is equal to Type
+  template<auto Keyword, typename Opts, typename Type>
+  inline constexpr bool option_type_is =  std::same_as<Type, rbr::result::fetch_t<Keyword,Opts>>;
```
DenisYaroshevskiy, line 266: concept? Why is this needed?

test/doc/traits/callable_constant.cpp
```diff
@@ -0,0 +1,60 @@
...
+    // A constant must provide a value() static member with up to two parameters:
+    //  * an eve::as instance to specify the output type
+    //  * an optional eve::options containing potential decorators passed to the constant
+    template<typename T>
+    static EVE_FORCEINLINE T value(eve::as<T> const&) { return static_cast<T>(3.14159216); }
+
+    // Supported function calls are still required
+    template<eve::floating_value T>
+    EVE_FORCEINLINE T     operator()(as<T>    v) const { return EVE_DISPATCH_CALL(v); }
+    EVE_FORCEINLINE auto  operator()(as<char> v) const { return EVE_DISPATCH_CALL(v); }
```
DenisYaroshevskiy, line 23: We don't do auto! Otherwise the error messages are not as good.

Also - char for some_pi? Comment maybe like "this is purely example or smth"

test/doc/traits/callable_object.cpp
```diff
@@ -0,0 +1,49 @@
...
+namespace eve
+{
+  template<typename Options>
+  struct func_t : callable<func_t, Options>
+  {
+    // operator() are defined here to maximize quality of error message. They all use EVE_DISPATCH_CALL at some point.
+    template<eve::integral_value T>
+    EVE_FORCEINLINE auto  operator()(T v)       const { return EVE_DISPATCH_CALL(v); }
+    EVE_FORCEINLINE auto  operator()(double v)  const { return EVE_DISPATCH_CALL(v); }
+    EVE_FORCEINLINE auto  operator()(float)     const = delete;
```
DenisYaroshevskiy, line 15: No auto for return type - that defeats the purpose.

test/doc/traits/callable_object_from.cpp
```diff
@@ -0,0 +1,53 @@
...
+namespace my_lib
+{
+  template<typename Options>
+  struct func_t : eve::callable<func_t, Options>
+  {
+    // operator() are defined here to maximize quality of error message. They all use EVE_DISPATCH_CALL at some point.
+    template<eve::integral_value T>
+    EVE_FORCEINLINE auto  operator()(T v)       const { return EVE_DISPATCH_CALL(v); }
+    EVE_FORCEINLINE auto  operator()(double v)  const { return EVE_DISPATCH_CALL(v); }
+    EVE_FORCEINLINE auto  operator()(float)     const = delete;
```
DenisYaroshevskiy, line 19: Again - auto for return type.

test/doc/traits/callable_specs.cpp
```diff
@@ -0,0 +1,56 @@
...
+  using namespace rbr::literals;
+  inline constexpr auto precise = "precise"_fl;
+  inline constexpr auto scale   = "scale"_fl;
+
+  // Defines a support specification
+  struct precision
+  {
+    template<eve::callable_option O>
+    auto process(auto const& base, O const& opt) const
+    //requires( eve::options<O>::contains_any(precise, scale))
```
DenisYaroshevskiy, line 18: Should this be deleted?

test/doc/traits/callable_supports.cpp
```diff
@@ -0,0 +1,87 @@
...
+namespace eve
+{
+  // + callable can add rules for decorator to use on a given callable and a default set of options
+  // + eve::conditional_option give the ability to pass a mask as a decoration.
+  template<typename Options>
+  struct func_t : callable<func_t, Options, conditional_option>
+  {
+    // Note that decoration doesn't impact signature declaration
+    template<eve::integral_value T>
+    EVE_FORCEINLINE auto  operator()(T v) const { return EVE_DISPATCH_CALL(v); }
```
DenisYaroshevskiy, line 16: Again - no auto for return type.

test/doc/traits/callable_supports.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+  // Build the callable object from the function object type
+  inline constexpr auto func = functor<func_t>;
+
+  template<typename Options>
+  struct other_func_t : callable<other_func_t, Options, relative_conditional_option>
+  {
+    // Note that decoration doesn't impact signature declaration
+    template<eve::integral_value T>
+    EVE_FORCEINLINE auto  operator()(T v) const { return EVE_DISPATCH_CALL(v); }
```
DenisYaroshevskiy, line 28: ^

include/eve/traits/overload/default_behaviors.hpp
jfalcou: mask_op is written so all the necessary thing are done properly for the masking.
It is what we use everywhere already.

include/eve/traits/overload/supports.hpp
jfalcou: The thing we use to define actual functor with a proper default options.
We verbally agreed on having this last time.

include/eve/traits/overload/supports.hpp
jfalcou: Removed

test/doc/traits/callable_constant.cpp
jfalcou: I am gonna make an arbitraryconsatnt instead, I need that to demonstrate that it can be non template

include/eve/traits/overload/supports.hpp
jfalcou: It is used in some callable for doign thing like : 

```
// Are we dealing with a no-condition call ?
      if constexpr( option_type_is<condition_key, O, ignore_none_> )
      {
       // 
      }
```

I put it there but it def. should move to raberu

include/eve/traits/overload/default_behaviors.hpp
jfalcou: For reference :

```c++
  template<conditional_expr C, typename Op, typename Arg0, typename... Args>
  EVE_FORCEINLINE auto mask_op( C const& c
                              , [[maybe_unused]] Op f
                              , [[maybe_unused]] Arg0 const& a0
                              , [[maybe_unused]] Args const&... as
                              )
  {
    using r_t       = decltype(f(a0,as...));
    auto const cond = c.mask(eve::as<r_t>());

          if constexpr( C::is_complete && !C::is_inverted ) return alternative(c,a0,eve::as<r_t>{});
    else  if constexpr( C::is_complete &&  C::is_inverted ) return f(a0,as...);
    else                                                    return if_else( cond
                                                                          , f(a0,as...)
                                                                          , alternative(c,a0,eve::as<r_t>{})
                                                                          );
  }
  ```

File: include/eve/traits/overload/default_behaviors.hpp (hunk 40 above)
Comment (jfalcou) [reply include/eve/traits/overload/default_behaviors.hpp:41]: # With =delete
## g++
```c++
<source>: In function ‘int main()’:
<source>:42:33: error: use of deleted function ‘constexpr void eve::callable<Func, OptionsValues, Options>::operator[](T) const [with T = const char*; Func = eve::func_t; OptionsValues = eve::options<>; Options = {eve::conditional_option}]’
   42 |   std::cout << eve::func["false"](8) << "\n";
      |                                 ^
In file included from /home/falcou/dev/eve/include/eve/traits/overload.hpp:12,
                 from <source>:3:
/home/falcou/dev/eve/include/eve/traits/overload/default_behaviors.hpp:41:41: note: declared here
   41 |     template<typename T> constexpr void operator[](T t) const =delete;
      |                                         ^~~~~~~~
```

## clang
```c++
<source>:42:25: error: overload resolution selected deleted operator '[]'
  std::cout << eve::func["false"](8) << "\n";
               ~~~~~~~~~^~~~~~~~
/home/falcou/dev/eve/include/eve/traits/overload/default_behaviors.hpp:41:41: note: candidate function [with T = const char *] has been explicitly deleted
    template<typename T> constexpr void operator[](T t) const =delete;
                                        ^
/home/falcou/dev/eve/include/eve/traits/overload/default_behaviors.hpp:34:20: note: candidate template ignored: constraints not satisfied [with T = const char *]
    constexpr auto operator[](T t) const
                   ^
/home/falcou/dev/eve/include/eve/traits/overload/default_behaviors.hpp:35:41: note: because 'b[t]' would be invalid: no viable overloaded operator[] for type 'const eve::callable<eve::func_t, eve::options<>, eve::conditional_option>::base' (aka 'const decorated_with<eve::options<rbr::settings<>>, eve::conditional_option>')
    requires( requires(base const& b) { b[t];} )
                                        ^
```

#  Without `=delete`

## g++
```c++
<source>: In function ‘int main()’:
<source>:42:25: error: no match for ‘operator[]’ (operand types are ‘const eve::func_t<eve::options<> >’ and ‘const char [6]’)
   42 |   std::cout << eve::func["false"](8) << "\n";
      |                         ^
In file included from /home/falcou/dev/eve/include/eve/traits/overload.hpp:12,
                 from <source>:3:
/home/falcou/dev/eve/include/eve/traits/overload/default_behaviors.hpp:34:20: note: candidate: ‘template<class T> constexpr auto eve::callable<Func, OptionsValues, Options>::operator[](T) const requires requires(const eve::base& b) {(b)[eve::callable<Func, OptionsValues, Options ...>::operator[]::t];} [with Func = eve::func_t; OptionsValues = eve::options<>; Options = {eve::conditional_option}]’
   34 |     constexpr auto operator[](T t) const
      |                    ^~~~~~~~
/home/falcou/dev/eve/include/eve/traits/overload/default_behaviors.hpp:34:20: note:   template argument deduction/substitution failed:
/home/falcou/dev/eve/include/eve/traits/overload/default_behaviors.hpp:34:20: note: constraints not satisfied
/home/falcou/dev/eve/include/eve/traits/overload/default_behaviors.hpp: In substitution of ‘template<class T> constexpr auto eve::callable<eve::func_t, eve::options<>, eve::conditional_option>::operator[](T) const requires requires(const eve::base& b) {(b)[eve::callable<Func, OptionsValues, Options ...>::operator[]::t];} [with T = eve::func_t]’:
<source>:42:33:   required from here
/home/falcou/dev/eve/include/eve/traits/overload/default_behaviors.hpp:34:20:   required by the constraints of ‘template<template<class> class Func, class OptionsValues, class ... Options> template<class T> constexpr auto eve::callable<Func, OptionsValues, Options>::operator[](T) const requires requires(const eve::base& b) {(b)[eve::callable<Func, OptionsValues, Options ...>::operator[]::t];}’
/home/falcou/dev/eve/include/eve/traits/overload/default_behaviors.hpp:35:15:   in requirements with ‘const eve::callable<Func, OptionsValues, Options>::base& b’ [with T = const char*; Func = eve::func_t; OptionsValues = eve::options<rbr::settings<> >; Options = {eve::conditional_option}]
/home/falcou/dev/eve/include/eve/traits/overload/default_behaviors.hpp:35:42: note: the required expression ‘b[t]’ is invalid
   35 |     requires( requires(base const& b) { b[t];} )
      |                                         ~^
cc1plus: note: set ‘-fconcepts-diagnostics-depth=’ to at least 2 for more detail
```

## clang
```c++
<source>:42:25: error: no viable overloaded operator[] for type 'const func_t<eve::options<>>'
  std::cout << eve::func["false"](8) << "\n";
               ~~~~~~~~~^~~~~~~~
/home/falcou/dev/eve/include/eve/traits/overload/default_behaviors.hpp:34:20: note: candidate template ignored: constraints not satisfied [with T = const char *]
    constexpr auto operator[](T t) const
                   ^
/home/falcou/dev/eve/include/eve/traits/overload/default_behaviors.hpp:35:41: note: because 'b[t]' would be invalid: no viable overloaded operator[] for type 'const eve::callable<eve::func_t, eve::options<>, eve::conditional_option>::base' (aka 'const decorated_with<eve::options<rbr::settings<>>, eve::conditional_option>')
    requires( requires(base const& b) { b[t];} )
```

# Conclusion
 - For clang, it is about the same
  - For g++, it prevents this long cascade of "or else..."

include/eve/traits/overload/default_behaviors.hpp
jfalcou: derived, and pushed back in callable as a protected member

include/eve/traits/overload/default_behaviors.hpp
DenisYaroshevskiy: Values!

include/eve/traits/overload/supports.hpp
DenisYaroshevskiy: So should be concept anyways

DenisYaroshevskiy (commented):
Will look more

include/eve/traits/invoke/protocol.hpp
```diff
@@ -52,15 +52,6 @@ struct constant
...
-//! @concept callable
-//! @brief **EVE** callable
-//!
-//! A type `T` satisfies eve::callable if and only if it is tagged as such either
-//! manually or by inheriting from a T properties.
-//!
-//! @tparam T  T type for the @callable to check
-//==================================================================================================
 template<typename T>
 concept callable = requires(T) { typename T::callable_tag_type; };
```
DenisYaroshevskiy, line 56: No doc anymore?

include/eve/traits/overload/default_behaviors.hpp
DenisYaroshevskiy: add a comment anyway please

include/eve/traits/overload/default_behaviors.hpp
```diff
@@ -0,0 +1,151 @@
...
+      {
+        // Grab the condition
+        auto const cond     = opts[condition_key];
+
+        // Drop the conditional and process the function call with the universal masker
+        auto const rmv_cond = options{rbr::drop(condition_key, opts)};
+        auto const f        = Func<decltype(rmv_cond)>{rmv_cond};
+
+        // If the conditional call is supported, call it
+        if      constexpr( is_supported ) return detail::mask_op(cond, f, x0, xs...);
```
DenisYaroshevskiy, line 90: I am not sure if I understand how you plan to support conditionals.

You say `is_supported` for a conditional call. What would that spec look like?

Alternatively - it should always be supported and this just keeps being in all overloads.

include/eve/traits/overload/default_behaviors.hpp
```diff
@@ -0,0 +1,151 @@
...
+  {
+    template<callable_options O, typename T, typename... Ts>
+    constexpr auto behavior(auto arch, O const& opts, T x0,  Ts const&... xs) const
+    {
+      using func_t                =  Func<OptionsValues>;
+      constexpr bool is_supported = requires{ func_t::deferred_call(arch, opts, x0, xs...); };
+      constexpr bool any_simd     = (simd_value<T> || ... || simd_value<Ts>);
+
+      // Are we dealing with a no-condition call ?
+      if constexpr( option_type_is<condition_key, O, ignore_none_> )
```
DenisYaroshevskiy, line 69: * Ignore all support?

Do we have a uniform logic on how we handle ignores in element wise?

DenisYaroshevskiy (approved):
I will approve. I don't think we quite understand everything yet and there are some flaws I am seeing,  but they need expirience

include/eve/traits/invoke/protocol.hpp
jfalcou: This file is gonna be removed afterward. I removed the doc cause it was colliding with the new callable we use

include/eve/traits/overload/default_behaviors.hpp
jfalcou: Yup this is fishy, removign elementwise for now as discussed

include/eve/traits/overload/default_behaviors.hpp
jfalcou: I think this is a bit lacking indeed

PR #1693, opened 2023-11-21
Contfrac

jtlap:
this pr ads 2 functions able to evaluate continuous fractions which are needed by some implementation of special function (in particular in kyosu)


include/eve/module/contfrac/impl/lentz.hpp
```diff
@@ -0,0 +1,135 @@
...
+  SPDX-License-Identifier: BSL-1.0
+*/
+//==================================================================================================
+#pragma once
+
+namespace eve::detail
+{
+
+  template <typename T>
+  struct is_pair : public std::false_type{};
```
jfalcou, line 14: you don't need public

include/eve/module/contfrac/impl/lentz.hpp
```diff
@@ -0,0 +1,135 @@
...
+
+  template <typename T, typename U>
+  struct is_pair<kumi::tuple<T,U>> : public std::true_type{};
```
jfalcou, line 17: same here, public is unneeded

jfalcou (changes requested):
Can't this just be in math or does it depends of something else ?

include/eve/module/contfrac/impl/lentz.hpp
jtlap: ok

jfalcou (approved):

PR #1703, opened 2023-12-11
More fix for invoke/overload coexistence (transitional)

jfalcou:


DenisYaroshevskiy (approved):
Assuming the word enity is for a temporary hack

jfalcou:
Once kyosu migrate to new overload, thisnwholenfolder will be removed. This is a temporary measure.

PR #1716, opened 2023-12-28
Init support for RISCV vector extension

ita-sc:
Hello there! I've added initial support for the RISC-V vector extension in the EVE library. All unit tests have successfully passed with this patch. To verify, please follow these steps:

Obtain the riscv-gcc-13 toolchain (for sysroot). Set the path to it as the environment variable RISCV_GCC.
Use clang with the patch currently under review ([https://github.com/llvm/llvm-project/pull/76510]). To build it, navigate to the llvm directory:
```
 mkdir build
 cd build
 cmake -G Ninja -DLLVM_ENABLE_PROJECTS="clang" -DCMAKE_BUILD_TYPE=Release ../llvm
 cmake --build . -t clang
 ```
Add it to your PATH.
After completing the above steps, you should be able to run unit tests by specifying cmake/toolchain/clang.rvv128.cmake as the toolchain file.
Your contributions are highly appreciated! If you encounter any issues or have questions, feel free to reach out. Thanks for your valuable work!

### CODE REVIEW

**Comment by jfalcou** [comment]
Thanks a lot for this huge contribution. I will add the RiscV compiler to our CI image so we can have the tests running. We are gonna review this soonish.

File: include/eve/arch/riscv/tags.hpp [hunk 1]
```diff
@@ -0,0 +1,118 @@
...
+  {
+    auto binary_size   = cardinal * sizeof(Type) * 8;
+    auto min_frac_lmul = get_min_frac_lmul<Type>();
+    auto max_frac_lmul = get_max_frac_lmul<Type>();
+
+    auto frac_lmul_border = Size / max_frac_lmul;
+    if( binary_size <= frac_lmul_border )
+    {
+      auto frac_lmul = Size / binary_size;
+      frac_lmul      = std::bit_floor(frac_lmul);
```
jfalcou, line 73: This requires `<bit>` as per the error in the checks.

include/eve/detail/spy.hpp
```diff
@@ -808,6 +814,11 @@ namespace avx512
 #   endif
 # endif
 #endif
+#if !defined(SPY_SIMD_DETECTED) && defined(__riscv) && defined(__riscv_vector)
+#  define SPY_SIMD_DETECTED ::spy::detail::simd_version::rvv_
+#  define SPY_SIMD_IS_RISCV_FLEXIBLE_SVE
```
jfalcou, line 819: should it be just `SPY_SIMD_IS_RISCV_FLEXIBLE` ?
Also, care to make an issue/PR over at www.github.com/jfalcou/spy ?

include/eve/arch/riscv/as_register.hpp
```diff
@@ -0,0 +1,187 @@
...
+namespace eve
+{
+template<typename Type, typename Size, rvv_abi ABI> struct as_register<Type, Size, ABI>
+{
+  static constexpr auto lmul = riscv_rvv_dyn_::getLMUL<Type>(Size::value);
+
+  static constexpr bool is_fp_v     = std::is_floating_point_v<Type>;
+  static constexpr bool is_signed_v = std::is_signed_v<Type>;
+
+#  ifdef EVE_RISCV_REG_CHOOSE
```
DenisYaroshevskiy, line 29: heavy macro usage seems unjustified on a surface

DenisYaroshevskiy (commented):
Hii.

Thank you for putting in effort!

 96 files will take some time to review, I personally will do in chunks - some comments - then more comments etc.

include/eve/arch/riscv/rvv_common_masks.hpp
```diff
@@ -0,0 +1,56 @@
+//==================================================================================================
+/*
+  EVE - Expressive Vector Engine
+  Copyright : EVE Project Contributors
+  SPDX-License-Identifier: BSL-1.0
+*/
+//==================================================================================================
+#pragma once
+
+#include <eve/detail/function/friends.hpp>
```
DenisYaroshevskiy, line 10: That's not an include you should use

include/eve/arch/riscv/rvv_common_masks.hpp
```diff
@@ -0,0 +1,56 @@
...
+
+namespace eve::detail
+{
+template<arithmetic_scalar_value T, typename N>
+EVE_FORCEINLINE logical<wide<T, N>>
+                rvv_true()
+{
+  static constexpr auto   lmul     = riscv_rvv_dyn_::getLMUL<T>(N::value);
```
DenisYaroshevskiy, line 18: I'm struggling with `riscv_rvv_dyn_`. I can't find where it;s defined.
Is it that different machines use different number of bits per logical element?

Like on one machine logical<char> is represented by 1 bit, on the other logical<char> - 8 bits?

include/eve/arch/riscv/rvv_common_masks.hpp
```diff
@@ -0,0 +1,56 @@
...
+  static constexpr size_t size     = sizeof(T) * 8;
+  static constexpr size_t bit_size = lmul > 0 ? size / lmul : size * (-lmul);
```
DenisYaroshevskiy, line 20: this seems to be copypasted a lot. Should we just have `logical<wide<T, N>>::platform_bit_size` or smth?

Also - really - this is changing depending on machine???

include/eve/arch/riscv/tags.hpp
```diff
@@ -0,0 +1,118 @@
...
+#include <climits>
+
+namespace eve
+{
+//================================================================================================
+// ABI tags for all RISCV bits SIMD registers
+//================================================================================================
+template<std::size_t Size> struct rvv_abi_
+{
+  static_assert(CHAR_BIT == 8, "[eve riscv] - For riscv we expect CHAR_BIT to be 8");
```
DenisYaroshevskiy, line 25: we everywhere expect that - I would remove this.

include/eve/arch/riscv/rvv_common_masks.hpp
```diff
@@ -0,0 +1,56 @@
...
+}
+
+template<arithmetic_scalar_value T, typename N>
+EVE_FORCEINLINE logical<wide<T, N>>
+                rvv_none()
+{
+  static constexpr auto   lmul     = riscv_rvv_dyn_::getLMUL<T>(N::value);
+  static constexpr size_t size     = sizeof(T) * 8;
+  static constexpr size_t bit_size = lmul > 0 ? size / lmul : size * (-lmul);
+  if constexpr( bit_size == 1 ) return __riscv_vmclr_m_b1(N::value);
```
DenisYaroshevskiy, line 37: my lack of knowledge here shows through heavily - what is the difference between different clears? I'd expect all clears to be the same.

include/eve/arch/riscv/rvv_common_masks.hpp
```diff
@@ -0,0 +1,56 @@
...
+  if constexpr( bit_size == 64 ) return __riscv_vmclr_m_b64(N::value);
+}
+
+template<arithmetic_scalar_value T, typename N>
+EVE_FORCEINLINE logical<wide<T, N>>
+                rvv_one(size_t i)
+{
+  std::array<T, N::value> Vals = {};
+  Vals[i]                      = 1;
+  wide<T, N> Vector(Vals.begin(), Vals.end());
```
DenisYaroshevskiy, line 52: More our style

```
std::array<T, N::value> arr = {};
arr[i] = 1;
return wide<T, N>{arr.data()} != 0;
```

include/eve/arch/riscv/tags.hpp
```diff
@@ -0,0 +1,118 @@
...
+  static constexpr std::size_t MaxLmul         = 8;
+  static constexpr std::size_t bits            = Size * MaxLmul;
+  static constexpr std::size_t bytes           = bits / 8;
+  static constexpr bool        is_wide_logical = false;
+
+  template<typename Type>
+  static constexpr bool is_full = ((Type::size() * sizeof(typename Type::value_type)) >= bytes);
+
+  template<typename Type>
+  static constexpr std::size_t fundamental_cardinal = Size / 8 / sizeof(Type);
```
DenisYaroshevskiy, line 35: On a surface this does not strike me as correct.

fundamental cardinal is cardinal for which we have a native register. Like xmm on avx2.

In my mind we differ in terminology with RISK-V here - everything beyond max length requires masking and isn't fundamental cardinal.

include/eve/arch/riscv/tags.hpp
```diff
@@ -0,0 +1,118 @@
...
+
+  template<typename Type>
+  static constexpr std::size_t expected_cardinal = Size / 8 / sizeof(Type) * MaxLmul;
+
+  template<typename Type> static consteval int get_min_frac_lmul()
```
DenisYaroshevskiy, line 40: I'd say constexpr not consteval.

include/eve/arch/riscv/tags.hpp
```diff
@@ -0,0 +1,118 @@
...
+    // Return bigger than size of vector register value.
+    return Size + 1;
+  }
+
+  template<typename Type> static consteval int get_max_frac_lmul()
+  {
+    auto type_size = sizeof(Type);
+    if( type_size == 1 ) return 2;
+    if( type_size == 2 ) return 2;
+    if( type_size == 4 ) return 2;
```
DenisYaroshevskiy, line 56: this looks weird. Maybe comment and return 2, if it's ;ess than 4?

include/eve/arch/riscv/tags.hpp
```diff
@@ -0,0 +1,118 @@
...
+
+  template<arithmetic_scalar_value T> using m1_fixed = fixed<Size / 8 / sizeof(T)>;
+};
+
+#ifdef __riscv_v_fixed_vlen
+struct riscv_rvv_dyn_ : rvv_abi_<__riscv_v_fixed_vlen>
+{};
+#else
+struct riscv_rvv_dyn_ : rvv_abi_<1>
+{};
```
DenisYaroshevskiy, line 90: this doesn't make sense to me. Is it like a fallback? If we can't run v5, risk-v5 should not be used by eve

include/eve/arch/riscv/tags.hpp
```diff
@@ -0,0 +1,118 @@
...
+#endif
+
+//================================================================================================
+// Dispatching tag for ARM SIMD implementation
```
DenisYaroshevskiy, line 94: not arm

include/eve/arch/riscv/tags.hpp
```diff
@@ -0,0 +1,118 @@
...
+{
+  using is_rvv = void;
+};
+struct rvv_api_ : simd_api<rvv_, spy::rvv_>
+{
+  using is_rvv = void;
+};
+
+//================================================================================================
+// SVE extensions tag objects
```
DenisYaroshevskiy, line 106: comments

include/eve/arch/riscv/tags.hpp
```diff
@@ -0,0 +1,118 @@
...
+//================================================================================================
+inline constexpr rvv_     rvv     = {};
+inline constexpr rvv_api_ rvv_api = {};
+
+//================================================================================================
+// RISCV SVE ABI concept
+//================================================================================================
+template<typename T>
+concept rvv_abi = detail::is_one_of<T>(detail::types<riscv_rvv_dyn_> {});
```
DenisYaroshevskiy, line 115: why dyn

DenisYaroshevskiy (commented):
I luck understanding of risk-v simd fundamentals. 
I see this doc https://github.com/riscv/riscv-v-spec/blob/master/v-spec.adoc but it's huge and contains a lot of information that I don't immediately need. 

I also can't find intrinsics descriptions. 

What are you using?

test/unit/module/algo/algorithm/min_element_special_cases.cpp
```diff
@@ -11,11 +11,11 @@
 
 #include <span>
 
-TTS_CASE("Min element one pass, uint8 index")
+TTS_CASE("Min element one pass, uint16 index")
 {
   auto alg = eve::algo::min_element         //
       [eve::algo::single_pass]              //
-      [eve::algo::index_type<std::uint8_t>] //
```
DenisYaroshevskiy, line 18: No - the point was specifically to tests index uint8.

test/unit/memory/load/tuple.cpp
```diff
@@ -45,13 +45,13 @@ TTS_CASE_TPL( "Check load behavior with soa_ptr", eve::test::scalar::all_types)
...
-  auto [data1,idx1] = page<T            , typename w8_t::cardinal_type >();
-  auto [data2,idx2] = page<double       , typename w8_t::cardinal_type >();
+  auto [data0,idx0] = page<std::int8_t  , typename w_t::cardinal_type >();
+  auto [data1,idx1] = page<T            , typename w_t::cardinal_type >();
+  auto [data2,idx2] = page<double       , typename w_t::cardinal_type >();
 
-  auto src = eve::soa_ptr    ( eve::as_aligned(&data0[idx0],typename w8_t::cardinal_type{})
+  auto src = eve::soa_ptr    ( eve::as_aligned(&data0[idx0],typename w_t::cardinal_type{})
                               , &data1[idx1] - 1
-                              , eve::as_aligned(&data2[idx2],typename w8_t::cardinal_type{})
```
DenisYaroshevskiy, line 54: the test was correct. It's testing fixed size wide

test/unit/api/regular/wide.cpp
```diff
@@ -72,7 +72,7 @@ TTS_CASE_TPL( "Check eve::wide splat constructor", eve::test::simd::all_types)
 
   // Test smaller size wide for non-garbage
   using v_t = typename T::value_type;
-  if constexpr( T::size() < eve::fundamental_cardinal_v<v_t> && !eve::has_emulated_abi_v<T> )
```
DenisYaroshevskiy, line 75: this comes from the fundamental wide being one idea I think.

include/eve/arch/riscv/rvv_common_masks.hpp
ita-sc: They return different types.

For example, `__riscv_vmclr_m_b1` returns `vbool1_t`, and `__riscv_vmclr_m_b8` results in `vbool8_t`.

include/eve/arch/riscv/tags.hpp
ita-sc: Removed.

include/eve/arch/riscv/rvv_common_masks.hpp
ita-sc: > Should we just have logical<wide<T, N>>::platform_bit_size or smth?

This is not platform_bit_size of logical, it is the number that should be put to form right type. This number is equal to SEW/LMUL. Renamed it to `ratio`.

> Also - really - this is changing depending on machine?

Well, it depends on `VLEN` of your machine.

For example, you have `VLEN==128`. To operate with 8 int32 with one instruction, you will need to set LMUL( the number of consecutive vector registers that participate as one operand) to M2 (you need 2 registers). And mask for this you will need type `vbool16_t`(32/2=16).

If you have `VLEN==256`, you can operate 8 int32 with LMUL == M1, so you need to use `vbool32_t`(32/1=32).

include/eve/arch/riscv/rvv_common_masks.hpp
ita-sc: It is defined in tags and I use it as common place for different support function, like calculation of LMUL( vector register group count - the number of vector registers, that participate in this operation), as well as standard EVE functionality (e.g. `expected_cardinal`).

> different machines use different number of bits per logical element

Renamed. There I calculate `ratio=SEW/LMUl`.

include/eve/arch/riscv/rvv_common_masks.hpp
ita-sc: Ok, I wanted to use it as I need `self_neq`, but it seems I do not need this. Removed.

include/eve/arch/riscv/as_register.hpp
ita-sc: I am trying to create a group of functions to make the best decision for vector type depending on requested cardinal/type and sew(single element width). For example, let's suppose `vlen == 128`:
1) you need to work with 16 int8 elements - you can use vector type `vint8m1_t`
2) you need to work with 128 uint8 elements - you can use vector type `vuint8m8_t`
3) you need to work with 8 float elements - you can use vector type `vfloat32m4`

include/eve/arch/riscv/tags.hpp
ita-sc: My bad, added.

include/eve/arch/riscv/rvv_common_masks.hpp
ita-sc: Ok, fixed

include/eve/arch/riscv/tags.hpp
ita-sc: Well, I thought about fundamental cardinal as the 
```cpp
the smallest register able to store values of type of type `Type` for a given SIMD `ABI` with no uninitialized lanes
```
(in file include/eve/arch/fundamental_cardinal.hpp)

So in this implementation fundamental cardinal is one vector register (actually, for some cases we can use fractional LMUL) and expected cardinal is the maximum available vector registers, combined together.

include/eve/arch/riscv/tags.hpp
ita-sc: I want to guarantee that it must be called only in compile time

include/eve/arch/riscv/tags.hpp
ita-sc: Yep, fixed

include/eve/arch/riscv/tags.hpp
ita-sc: The idea is following:

You need to specify while running complier `VLEN` (bit size of one vector register) by passing `-mrvv-vector-bits=size`. After this compiler define `__riscv_v_fixed_vlen` that will be equal the number that you passed  to compiler (currently I use 128).

As tags included without information that current platform is RISC-V with vector extension, we should check that `__riscv_v_fixed_vlen` is defined, and if not - define some empty `riscv_rvv_dyn_` just not to break complation.

include/eve/arch/riscv/tags.hpp
ita-sc: My bad, fixed to RISC-V

include/eve/arch/riscv/tags.hpp
ita-sc: Well, I wanted to note that we should use this template for any VLEN. I could rename, if you want.

test/unit/module/algo/algorithm/min_element_special_cases.cpp
ita-sc: Ok, will replace back. There was a problem with rvv with `vlen=128`:

We can process 128 elements by one operation  and with `unroll_count==2` we get overflow for uint8_t.

Will forbid this test in next iteration.

test/unit/api/regular/wide.cpp
ita-sc: For RISC-V I forbid this as we can have different types as storage type depending on cardinal value. To support this I need to add additional constructor in core wide functionality.

DenisYaroshevskiy:
I think I need to understand more about risc v simd/vector code. What docs do you use? I only see a very large GitHub repo, that has a lot of information and no intrinsics.

Can you share smth?

ita-sc:
> I think I need to understand more about risc v simd/vector code. What docs do you use? I only see a very large GitHub repo, that has a lot of information and no intrinsics.
> 
> Can you share smth?

Yes, there is a separate repository: https://github.com/riscv-non-isa/rvv-intrinsic-doc

There you need release v1.0-rc0

include/eve/detail/function/simd/riscv/compounds.hpp
```diff
@@ -0,0 +1,293 @@
...
+    {                                                                                              \
+      using scalar_t            = as_integer_t<P, unsigned>;                                       \
+      using i_t                 = wide_t::template rebind<scalar_t, N>;                            \
+      constexpr auto tgt        = as<i_t>();                                                       \
+      auto           other_cast = std::bit_cast<scalar_t>(other);                                  \
+      auto           self_cast  = bit_cast(self, tgt);                                             \
+      self = bit_cast(i_t(intrinsic(self_cast, other_cast, N::value)), as(self));                  \
+      return self;                                                                                 \
+    }                                                                                              \
+  }
```
DenisYaroshevskiy, line 172: I dislike macros intensely. 
I would prefer callbacks, but it's going to be annoyting because of forceinline.

For exanple we can
```
  enum class bit_op { and_, or_, xor_ }; 

  template<bit_op op, 
        plain_scalar_value T, typename N, plain_scalar_value P>                                 
  EVE_FORCEINLINE auto& rvv_bit_compound_impl(wide<T, N>& self, P const& other) {
    ...
    if constexpr (op == bit_op::and_) self_cast = __riscv_vand(self_cast, other_cast);
    if constexpr (op == bit_op::xor_) self_cast = __riscv_vxor(self_cast, other_cast);
    ...
  }

```

include/eve/detail/function/simd/riscv/load.hpp
```diff
@@ -0,0 +1,92 @@
...
+#include <eve/module/core/regular/unalign.hpp>
+
+namespace eve::detail
+{
+
+template<arithmetic_scalar_value T, typename N, typename PtrTy>
+EVE_FORCEINLINE wide<T, N>
+                perform_load(logical<wide<T, N>> mask, as<wide<T, N>> tgt, PtrTy p)
+{
+  auto           zero_init = make(as<wide<T, N>> {}, static_cast<T>(0));
```
DenisYaroshevskiy, line 26:
```
wide<T, N> zero_init{0};
```

or

```
auto zero_init = eve::zero(tgt);
```

include/eve/arch/riscv/as_register.hpp
```diff
@@ -0,0 +1,187 @@
...
+  static constexpr auto find()
+  {
+    if constexpr( bit_size == 1 ) { EVE_RVV_RET_MASK_TYPE(1); }
+    if constexpr( bit_size == 2 ) { EVE_RVV_RET_MASK_TYPE(2); }
+    if constexpr( bit_size == 4 ) { EVE_RVV_RET_MASK_TYPE(4); }
+    if constexpr( bit_size == 8 ) { EVE_RVV_RET_MASK_TYPE(8); }
+    if constexpr( bit_size == 16 ) { EVE_RVV_RET_MASK_TYPE(16); }
+    if constexpr( bit_size == 32 ) { EVE_RVV_RET_MASK_TYPE(32); }
+    if constexpr( bit_size == 64 ) { EVE_RVV_RET_MASK_TYPE(64); }
+  }
```
DenisYaroshevskiy, line 178: heavy macro usage makes this very difficult to review.

How close can you get to?

```
template <typename rvv_type, std::size_t num, std::size_t denum>
using rvv_with_attr = rvv_type __attribute__((riscv_rvv_vector_bits(__riscv_v_fixed_vlen * num / denum)));

auto find() {
   if constexpr ( std::same_as<T, std::int8_t> ) {
      if constexpr ( N::value == 1 ) return rvv_with_attr<...>{}; 
   }
}
```

DenisYaroshevskiy (commented):
I skimmed a bit of the doc, will look more. In the mean time, I really think that heavy macro usage is not helpful. Let's try to figure out how we can do it less.

include/eve/arch/riscv/as_register.hpp
ita-sc: Removed defines from `as_register.hpp`.

test/unit/module/algo/algorithm/min_element_special_cases.cpp
ita-sc: Done

include/eve/arch/riscv/tags.hpp
ita-sc: Fixed

test/unit/api/regular/shuffle_v2/shuffle_v2_8x1.cpp
```diff
@@ -28,6 +28,8 @@ TTS_CASE("shuffle_v2: 8x1")
       return 3;
     }
 
+    if (eve::current_api == eve::rvv) return 2;
```
DenisYaroshevskiy, line 31: It's fine for now, but I don't believe level 2 is the correct answer here

test/test.hpp
```diff
@@ -251,13 +254,14 @@ namespace tts
...
     using v_t = eve::element_type_t<W>;
+    using c_t  = eve::cardinal_t<W>;
 
     // Add garbage at the end of sub-native registers
     // For emulated type, there is no such gap so we don't do anything
     if constexpr( (W::size() < eve::fundamental_cardinal_v<v_t>) && !eve::has_emulated_abi_v<W> )
     {
       using p_t   = eve::as_arithmetic_t<eve::as_integer_t<v_t, unsigned>>;
-      using ftype = eve::as_wide_t<v_t, eve::fundamental_cardinal_t<v_t>>;
+      using ftype = eve::as_wide_t<v_t, c_t>;
```
DenisYaroshevskiy, line 264: Wait - this can't be right. And why this change, fundamental is 1 and size is never less

include/eve/module/core/regular/impl/simd/riscv/slide_left.hpp
```diff
@@ -0,0 +1,25 @@
...
+  SPDX-License-Identifier: BSL-1.0
+*/
+//==================================================================================================
+#pragma once
+
+namespace eve::detail
+{
+template<scalar_value T, typename N, std::ptrdiff_t Shift>
+EVE_FORCEINLINE auto
+slide_left_(EVE_SUPPORTS(rvv_), logical<wide<T, N>> v, index_t<Shift>) noexcept
```
DenisYaroshevskiy, line 14: again, emulation should do this, why doesn't that work

include/eve/module/core/regular/impl/simd/riscv/logical_xor.hpp
```diff
@@ -0,0 +1,20 @@
...
+namespace eve::detail
+{
+template<arithmetic_scalar_value T, typename N, scalar_value U>
+EVE_FORCEINLINE logical<wide<T, N>>
+                logical_xor_(EVE_SUPPORTS(rvv_),
+                             logical<wide<T, N>> const                &a,
+                             logical<wide<U, N>> const                &b) noexcept
+requires rvv_abi<abi_t<T, N>>
+{
+  return self_neq(a, bit_cast(b, as<logical<wide<T, N>>> {}));
```
DenisYaroshevskiy, line 18: `a != bit_cast(b, as(a))`;

include/eve/module/core/regular/impl/simd/riscv/last_true.hpp
```diff
@@ -0,0 +1,43 @@
...
+requires rvv_abi<abi_t<T, N>>
+{
+  if constexpr( C::is_complete && !C::is_inverted ) return std::nullopt;
+  else
+  {
+    auto                m      = cond.mask(as<wide<T, N>> {});
+    logical<wide<T, N>> masked = __riscv_vmand(v0.storage, m, N::value);
+    return last_true(masked);
+  }
+}
```
DenisYaroshevskiy, line 29: are you sure you need this one.

include/eve/module/core/regular/impl/simd/riscv/last_true.hpp
```diff
@@ -0,0 +1,43 @@
...
+                last_true_(EVE_SUPPORTS(rvv_), top_bits<logical<wide<T, N>>> const                &v0) noexcept
+requires rvv_abi<abi_t<T, N>>
+{
+  // TODO: optimize.
+  for( int i = N::value - 1; i > -1; --i )
+  {
+    if( v0.get(i) ) return i;
+  }
+  return {};
+}
```
DenisYaroshevskiy, line 42: I am not wild about doing this tbh. Is there really no good solution in the isa? should be smth easy-ish. 

You can iota + mask + maximum for example.

include/eve/module/core/regular/impl/simd/riscv/bit_notand.hpp
```diff
@@ -0,0 +1,25 @@
...
+template<scalar_value T, typename N>
+EVE_FORCEINLINE auto
+bit_notand_(EVE_SUPPORTS(rvv_), wide<T, N> const& v0, wide<T, N> const& v1) noexcept -> wide<T, N>
+requires rvv_abi<abi_t<T, N>>
+{
+  auto v0_copy    = v0;
+  auto bitnot_res = self_bitnot(v0_copy);
+  return self_bitand(bitnot_res, v1);
+}
+}
```
DenisYaroshevskiy, line 25: I suspect this file can be deleted

DenisYaroshevskiy (commented):
I will do a deeper dive later - there is still a lot. Thank you for cleaning up macros. I think there are some simplifications that can be made.

FYI: we ususally don't enable all tests for new arch in one go. We do them from  internal.exe, than core.exe and then add more targets. You don't need to change for this pr, just fyi.

include/eve/arch/riscv/top_bits.hpp
```diff
@@ -0,0 +1,146 @@
...
+      to_return <<= 8;
+      to_return |= Values.at(id);
+    }
+    // we need to clear result from agnostic values.
+    std::bitset<64> to_clean;
+    for( int i = 0; i < static_size; ++i ) to_clean.set(i, true);
+
+    to_return &= to_clean.to_ullong();
+    return to_return;
+  }
```
DenisYaroshevskiy, line 108: oof. not amazing. But I guess you can't do anything. Maybe we can just completely not have this method.

include/eve/detail/function/simd/riscv/bit_cast.hpp
```diff
@@ -0,0 +1,586 @@
...
+  if constexpr( match(c, category::uint8) )
+  {
+    if constexpr( out_lmul == -4 ) return __riscv_vlmul_ext_u8mf4(a);
+    if constexpr( out_lmul == -2 ) return __riscv_vlmul_ext_u8mf2(a);
+    if constexpr( out_lmul == 1 ) return __riscv_vlmul_ext_u8m1(a);
+    if constexpr( out_lmul == 2 ) return __riscv_vlmul_ext_u8m2(a);
+    if constexpr( out_lmul == 4 ) return __riscv_vlmul_ext_u8m4(a);
+    if constexpr( out_lmul == 8 ) return __riscv_vlmul_ext_u8m8(a);
+  }
+}
```
DenisYaroshevskiy, line 101: On a surface level, I don't think that extract should be usued in bit_cast

include/eve/detail/function/simd/riscv/bit_cast.hpp
```diff
@@ -0,0 +1,586 @@
...
+template<scalar_value T, typename N, scalar_value U, typename M>
+EVE_FORCEINLINE wide<U, M>
+bit_cast_(EVE_SUPPORTS(rvv_), wide<T, N> const& x, as<wide<U, M>> const& to_as) noexcept
+requires rvv_abi<abi_t<T, N>> && rvv_abi<abi_t<U, M>> && same_wide_size<T, N, U, M>
+         && (sizeof(T) * N::value < sizeof(U) * M::value)
+{
+  if constexpr( is_aggregated_v<abi_t<T, N>> || is_aggregated_v<abi_t<U, M>> ) static_assert(false);
+  if constexpr( is_aggregated_v<abi_t<T, typename N::combined_type>>
+                || is_aggregated_v<abi_t<U, M>> )
+    static_assert(false);
```
DenisYaroshevskiy, line 420: aggregated won't be here, because it is not rvv_abi.

 also if sizeof(T) * N == sizeof(U) * M - should work.

include/eve/detail/function/simd/riscv/bit_cast.hpp
```diff
@@ -0,0 +1,586 @@
...
+  auto extended_in = riscv_lmul_ext(x);
+  for( auto id = N::value; id < N::combined_type::value; ++id )
+    extended_in.set(id, static_cast<T>(0));
+  return bit_cast(extended_in, to_as);
+}
+template<scalar_value T, typename N, scalar_value U, typename M>
+EVE_FORCEINLINE wide<U, M>
+bit_cast_(EVE_SUPPORTS(rvv_), wide<T, N> const& x, as<wide<U, M>> const& to_as) noexcept
+requires rvv_abi<abi_t<T, N>> && rvv_abi<abi_t<U, M>> && same_wide_size<T, N, U, M>
+         && (sizeof(T) * N::value > sizeof(U) * M::value)
```
DenisYaroshevskiy, line 430: You can only bit_cast if sizeof(T) * N::value == sizeof(U) * M::value - this overload is invalid

include/eve/detail/function/simd/riscv/bitmask.hpp
```diff
@@ -0,0 +1,46 @@
...
+
+//================================================================================================
+// Logical to Bits
+//================================================================================================
+template<typename T, typename N>
+EVE_FORCEINLINE auto
+to_mask(rvv_ const&, logical<wide<T, N>> p) noexcept
+{
+  return bit_cast(p.bits(), as<typename logical<wide<T, N>>::mask_type> {});
+}
```
DenisYaroshevskiy, line 44: Why this overload when the previous overload exists? It should be one or the other, unless I am missing something.

include/eve/detail/function/simd/riscv/combine.hpp
```diff
@@ -0,0 +1,65 @@
...
+{
+template<typename T, typename N>
+EVE_FORCEINLINE auto
+combine(rvv_ const&, wide<T, N> const& l, wide<T, N> const& h) noexcept
+requires rvv_abi<abi_t<T, N>>
+{
+  constexpr size_t combined_vl = N::combined_type::value;
+  using that_t                 = wide<T, typename N::combined_type>;
+
+  if constexpr( combined_vl > expected_cardinal_v<T> )
```
DenisYaroshevskiy, line 25: Do you mean aggregated?
I'd prefer if you said 

```
if constexpr ( eve::has_aggregated_abi_v<that_t> ) return that_t{l, h}:
else
{
}
```

include/eve/detail/function/simd/riscv/combine.hpp
```diff
@@ -0,0 +1,65 @@
...
+  }
+  else
+  {
+    auto           wider_l        = bit_cast(l, as<that_t> {});
+    auto           wider_h        = bit_cast(h, as<that_t> {});
+    constexpr auto shift_size     = N::value;
+    that_t         wider_h_placed = __riscv_vslideup(wider_h, wider_h, shift_size, combined_vl);
+    auto           mask_all_ones  = rvv_true<T, N>();
+    auto           wider_mask     = bit_cast(mask_all_ones, as<logical<that_t>> {});
+    return if_else(wider_mask, wider_l, wider_h_placed);
```
DenisYaroshevskiy, line 42: Unless `slide` does something silly, you should be able to just or them instead of if/else, no?

include/eve/detail/function/simd/riscv/combine.hpp
```diff
@@ -0,0 +1,65 @@
...
+  if constexpr( combined_vl > expected_cardinal_v<T> )
+  {
+    that_t that;
+
+    that.storage().segments[0] = l;
+    that.storage().segments[1] = h;
+
+    return that;
+  }
+  else { return to_logical(eve::combine(l.mask(), h.mask())); }
```
DenisYaroshevskiy, line 63: Do logicals not have slide?

include/eve/detail/function/simd/riscv/compounds.hpp
```diff
@@ -0,0 +1,293 @@
...
+  if constexpr( match(c, category::float_) ) return self = __riscv_vfmul(self, other, N::value);
+}
+template<plain_scalar_value T, value U, typename N>
+EVE_FORCEINLINE auto&
+self_mul(wide<T, N>& self, U const& other) noexcept
+requires scalar_value<U> && rvv_abi<abi_t<T, N>>
+{
+  constexpr auto c = categorize<wide<T, N>>();
+  if constexpr( match(c, category::integer_) )
```
DenisYaroshevskiy, line 34: `T y = static_cast<T>(other);` I think will make it cleaner a touch.

include/eve/detail/function/simd/riscv/compounds.hpp
```diff
@@ -0,0 +1,293 @@
...
+  {
+    return self = __riscv_vmul(self, static_cast<T>(other), N::value);
+  }
+  if constexpr( match(c, category::float_) )
```
DenisYaroshevskiy, line 38: you can just else for 2 cases, It's fine.

include/eve/detail/function/simd/riscv/compounds.hpp
```diff
@@ -0,0 +1,293 @@
...
+      auto           other_cast = std::bit_cast<scalar_t>(other);                                  \
+      auto           self_cast  = bit_cast(self, tgt);                                             \
+      self = bit_cast(i_t(intrinsic(self_cast, other_cast, N::value)), as(self));                  \
+      return self;                                                                                 \
+    }                                                                                              \
+  }
+
+RVV_BIT(self_bitand, __riscv_vand)
+RVV_BIT(self_bitxor, __riscv_vxor)
+RVV_BIT(self_bitor, __riscv_vor)
```
DenisYaroshevskiy, line 176: this macro gotta go

include/eve/detail/function/simd/riscv/compounds.hpp
```diff
@@ -0,0 +1,293 @@
...
+  else
+  {
+    using scalar_t             = as_integer_t<T, unsigned>;
+    using i_t                  = wide_t::template rebind<scalar_t, N>;
+    constexpr auto tgt         = as<i_t>();
+    auto           self_cast   = bit_cast(self, tgt);
+    i_t            to_cast_res = __riscv_vnot(self_cast, N::value);
+    self                       = bit_cast(to_cast_res, as(self));
+    return self;
+  }
```
DenisYaroshevskiy, line 204:
```
if constepxr (match(cat, unsigned_)) {
  __riscv_vnot
} else {
  // to unsigned and recurse
}
```

include/eve/detail/function/simd/riscv/compounds.hpp
```diff
@@ -0,0 +1,293 @@
...
+
+template<integral_scalar_value T, typename N, integral_scalar_value U>
+EVE_FORCEINLINE auto&
+self_shr(wide<T, N>& self, wide<U, N> shift) noexcept
+requires(rvv_abi<abi_t<T, N>> && sizeof(T) == sizeof(U))
+{
+  constexpr auto c = categorize<wide<T, N>>();
+  if constexpr( match(c, category::uint_) )
+  {
+    auto shift_casted = convert(shift, as<as_integer_t<U, unsigned>>());
```
DenisYaroshevskiy, line 245: this is the same between branches

include/eve/detail/function/simd/riscv/friends.hpp
```diff
@@ -0,0 +1,120 @@
...
+  {                                                                                                \
+    return fp_func(lhs, static_cast<T>(rhs), N::value);                                            \
+  }
+
+RVV_LOGIC(self_greater, __riscv_vmsgt, __riscv_vmsgtu, __riscv_vmfgt)
+RVV_LOGIC(self_less, __riscv_vmslt, __riscv_vmsltu, __riscv_vmflt)
+RVV_LOGIC(self_geq, __riscv_vmsge, __riscv_vmsgeu, __riscv_vmfge)
+RVV_LOGIC(self_leq, __riscv_vmsle, __riscv_vmsleu, __riscv_vmfle)
+RVV_LOGIC(self_eq, __riscv_vmseq, __riscv_vmseq, __riscv_vmfeq)
+RVV_LOGIC(self_neq, __riscv_vmsne, __riscv_vmsne, __riscv_vmfne)
```
DenisYaroshevskiy, line 53: please no macros like this. We can figure out template to do this if needed.

DenisYaroshevskiy (commented):
Looked a bit more.

I am so far not convinced about `riscv_rvv_dyn_` and using len multipliers in code. Let's first clean up everything else and then come back to this

ita-sc:
Hi again
Sorry for long delay on this review: actually I was waiting when LLVM will support the necessary functionality for this patch to be usable without additional modifications.
Though I tried to address all review comments, I suggest to start this review over.

include/eve/detail/function/simd/riscv/friends.hpp
ita-sc: Removed

include/eve/detail/function/simd/riscv/compounds.hpp
ita-sc: Moved upper

include/eve/detail/function/simd/riscv/compounds.hpp
ita-sc: Done

include/eve/detail/function/simd/riscv/compounds.hpp
ita-sc: Removed

test/unit/module/algo/algorithm/min_element_special_cases.cpp
```diff
@@ -17,68 +17,79 @@ TTS_CASE("Min element one pass, uint8 index")
...
+        for( std::int8_t& x : sub )
+        {
+          std::int8_t *found = std::to_address(alg(sub));
+          TTS_EQUAL(found, &x);
+          x = 0;
+        }
       }
     }
   }
+  else { TTS_PASS("For RISC-V uint8 not enough to store element index."); }
```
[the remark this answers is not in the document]
ita-sc: Done

test/unit/module/algo/algorithm/min_element_special_cases.cpp
```diff
@@ -17,68 +17,79 @@ TTS_CASE("Min element one pass, uint8 index")
       [eve::algo::single_pass]              //
       [eve::algo::index_type<std::uint8_t>] //
       [eve::algo::unroll<2>];
+  if constexpr( eve::current_api != eve::rvv )
```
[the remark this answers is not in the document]
ita-sc: Done

include/eve/module/core/regular/impl/simd/riscv/slide_right.hpp
```diff
@@ -0,0 +1,49 @@
+//==================================================================================================
+/*
+  EVE - Expressive Vector Engine
+  Copyright : EVE Project Contributors
+  SPDX-License-Identifier: BSL-1.0
+*/
+//==================================================================================================
+#pragma once
+
+namespace eve::detail
```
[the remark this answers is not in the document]
ita-sc: Well, for me this is not clear too, I just checked that arm sve do the same thing. Moreover, I actually had tests that fails if I'm not implement slide_left/right, for example unit.api.regular.swizzle.slide_left.

test/test.hpp
ita-sc: Changed to separate case for RISC-V RVV. The problem for rvv is that for type with fundamental cardinal we use one underlying type (with `m1` at the end), and for smaller types we can use other underlying type (to use part of real vector register, if possible).

include/eve/module/core/regular/impl/simd/riscv/logical_xor.hpp
ita-sc: Done

include/eve/arch/riscv/top_bits.hpp
ita-sc: Well, actually, this method is turned off almost for every logical in RISC-V

include/eve/detail/function/simd/riscv/bit_cast.hpp
ita-sc: Yep, removed

include/eve/detail/function/simd/riscv/bit_cast.hpp
ita-sc: Removed

DenisYaroshevskiy:
Welcome back, glad to have you. Will do as the last time I guess, piecemeal. I will do a pass - we work through it, I do another pass. I unfortunately can't do all at once - this is a very big piece of work.

test/unit/memory/load/tuple.cpp
ita-sc: The problem here was as wide type, associated with data1, will have 16 elements, so during load it will load more than needed, changed. Current version increases fixed wide size.

include/eve/module/core/regular/impl/simd/riscv/last_true.hpp
ita-sc: Actually, I suppose we can be faster, but I would prefer to make it in different MR, as this will require some investigation.

For example, I was thinking about casting it to several u64, and find it there. Unfortunately, I could not find in RVV ISA instruction that allows me to find position of  the last set bit.

include/eve/detail/function/simd/riscv/bit_cast.hpp
ita-sc: This is an "extend" operation, I use it when underlying type needs to be widened to support expected type. For example, for "char" type we support using 1/8 part of real vector register (called `mf8`), but when you cast it to "int", you can at least use 1/2 part of register(`mf2`).

This is documented here: `Chapter 7. Type system` of https://github.com/riscv-non-isa/rvv-intrinsic-doc

include/eve/arch/riscv/top_bits.hpp
```diff
@@ -0,0 +1,146 @@
...
+
+  // getters/setter ----------------------
+  static constexpr std::ptrdiff_t size() { return static_size; }
+
+  EVE_FORCEINLINE constexpr void set(std::ptrdiff_t i, bool x) { storage.set(i, x); }
+  EVE_FORCEINLINE constexpr bool get(std::ptrdiff_t i) const { return storage.get(i); }
+
+  EVE_FORCEINLINE constexpr explicit operator bool()
+  {
+    return __riscv_vcpop(storage, static_size) > 0;
```
DenisYaroshevskiy, line 87: !=

include/eve/detail/function/bit_cast.hpp
```diff
@@ -21,6 +21,15 @@ namespace eve
...
     }
+    // In riscv we may have different underlay type sizes
+#if defined(EVE_INCLUDE_RISCV_HEADER)
+    template<typename T, typename Target>
+    requires(sizeof(T) != sizeof(Target))
+    EVE_FORCEINLINE Target operator()(T const& a, as<Target> const& tgt) const noexcept
+    {
+      return EVE_DISPATCH_CALL(a, tgt);
+    }
+#endif
```
DenisYaroshevskiy, line 32: This doesn't make any sense to me. bitcast - means specifically to copy to the same size. That what it does.

What is the problem you are solving.

UPD: I think I see - you want to bitcast between a smaller and larger cardinals. But, I guess, they come out as different size. 

Why do you need to bitcast between them? If the spec says we have all those small registers, why not keep them?

include/eve/detail/function/simd/riscv/bit_cast.hpp
```diff
@@ -0,0 +1,569 @@
...
+  static_assert(N::combined_type::value != N::value);
+  constexpr auto in_lmul  = riscv_rvv_dyn_::getLMUL<T>(N::value);
+  constexpr auto out_lmul = riscv_rvv_dyn_::getLMUL<T>(N::combined_type::value);
+  if constexpr( out_lmul == in_lmul ) return a.storage();
+  if constexpr( match(c, category::float64) )
+  {
+    if constexpr( out_lmul == 2 ) return __riscv_vlmul_ext_f64m2(a);
+    if constexpr( out_lmul == 4 ) return __riscv_vlmul_ext_f64m4(a);
+    if constexpr( out_lmul == 8 ) return __riscv_vlmul_ext_f64m8(a);
+  }
```
DenisYaroshevskiy, line 33: we do else after return in if constexpr chains.

include/eve/detail/function/simd/riscv/bit_cast.hpp
DenisYaroshevskiy: if they are treated as different sizes, bitcasting them wouldn't be correct.

You can't bitcast xmm into ymm in our system, for example.

If every cardinal is fundamental, why are you trying to do this?

include/eve/detail/function/simd/riscv/bitmask.hpp
```diff
@@ -0,0 +1,46 @@
...
+  // As return by arm.
+  using int_type       = as_integer_t<T, unsigned>;
+  using u_wide_t       = wide<int_type, N>;
+  using u_mask_t       = logical<u_wide_t>;
+  auto     u_mask      = bit_cast(p, as<u_mask_t> {});
+  u_wide_t if_else_res = if_else(u_mask,
+                                 /* True value */ static_cast<int_type>(-1),
+                                 /* False Value */ static_cast<int_type>(0));
+  return if_else_res;
+}
```
DenisYaroshevskiy, line 34: What's true on risk-v5? is it not ffff?

include/eve/detail/function/simd/riscv/combine.hpp
DenisYaroshevskiy: +1

include/eve/detail/function/simd/riscv/combine.hpp
DenisYaroshevskiy: +1

include/eve/detail/function/simd/riscv/compounds.hpp
```diff
@@ -0,0 +1,306 @@
...
+EVE_FORCEINLINE auto&
+self_mul(wide<T, N>& self, U const& other) noexcept
+requires scalar_value<U> && rvv_abi<abi_t<T, N>>
+{
+  constexpr auto c = categorize<wide<T, N>>();
+  auto           y = static_cast<T>(other);
+  if constexpr( match(c, category::integer_) ) return self = __riscv_vmul(self, y, N::value);
+  if constexpr( match(c, category::float_) ) return self = __riscv_vfmul(self, y, N::value);
+}
+template<plain_scalar_value T, value U, typename N>
```
DenisYaroshevskiy, line 35: Please new lines between functions.

Also 

```
// self_mult --------------------

// self_add --------------------
```

This file blurs together a bit.

include/eve/arch/riscv/as_register.hpp
```diff
@@ -0,0 +1,497 @@
...
+template<typename Type, typename Size, rvv_abi ABI> struct as_register<Type, Size, ABI>
+{
+  static constexpr auto lmul = riscv_rvv_dyn_::getLMUL<Type>(Size::value);
+
+  static constexpr bool is_fp_v     = std::is_floating_point_v<Type>;
+  static constexpr bool is_signed_v = std::is_signed_v<Type>;
+
+  static consteval auto find_vint_mf8()
+  {
+    constexpr size_t bin_size = sizeof(Type) * 8;
```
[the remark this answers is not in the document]
ita-sc: Fixed

include/eve/arch/riscv/top_bits.hpp
ita-sc: Fixed

include/eve/detail/function/simd/riscv/compounds.hpp
ita-sc: Changed

test/unit/module/algo/algorithm/transform_keep_if_and_remove_generic_test.hpp
```diff
@@ -24,7 +24,7 @@ template <typename T, typename Algo, typename Op>
 void transform_keep_if_generic_test_aligned_ptr(eve::as<T>, Algo alg, Op op)
 {
   using e_t = eve::element_type_t<T>;
-  alignas(64) std::array<e_t, 23> data;
```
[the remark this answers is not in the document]
ita-sc: Because later we have eve::as_aligned(data.begin()) without cardinal, so it uses default cardinal, that is for short type is (128/16)*8=64. (the `*8` appears as at maximum, we use 8 real vector registers), that will result in alignment 64 * 2 = 128 byte.

This could be also fixed if I set  `eve::as_aligned(data.begin(), eve::fixed<8>{})` (or whatever value to be sure that it will not be greater that 64)

include/eve/detail/function/bit_cast.hpp
ita-sc: > Why do you need to bitcast between them?

There are two main reasons. 

First and the main one is when I need to bitcast between, for example, `wide<char, 4>` and `wide<int, 1>`. They have the same `sizeof(T) * N`, however for RISC-V underlying type sizes will be different:
 * `wide<char, 4>` could be represented as a part of real vector register in RISC-V vector intrinsics (for vlen=128 this will be 1/4 of real vector register).
 * For `wide<int, 1>` we have an ability to represent it with vlen=128 as 1/4 of vector register. However,  RISC-V vector intrinsics have some restrictions: they are unable to set LMUL to MF4 (1/4) if elements are of type int. It seems that this restriction comes from the fact that vector intrinsics are expected to work on any expected (by intrinsics) vector unit configuration. For example, if vlen=64 int will not fit into 1/4 part of vector. So, in intrinsics interface we need to use the type that corresponds to at least 1/2 of real vector register.
 
 That is why I need to "increase" LMUL, so underlying sizes will be different.

The second one is that sometimes I need an ability to cast between `wide` with different cardinal sizes, but the same types (for example `test\test.h`). In EVE it seems you usually just get a `.storage()`, and construct from it a new `wide` with increased cardinal. For RISC-V this is unavailable (as underlying types may be different), so for now I've decided to make bit_cast do this thing too. If there is better approach for this, I can try to refactor this.

include/eve/detail/function/simd/riscv/bit_cast.hpp
ita-sc: Yep, I thought about it, but for RISC-V this will made a huge chain of `if constexpr (...) ... else if constexpr (...) ... else if constexpr (...) ...` that will make code less readable and understandable. I try to return each if statement, so while debugging problems it will be easier to find problem

DenisYaroshevskiy: We have those huge chains, this is how we do it

include/eve/detail/function/simd/riscv/combine.hpp
ita-sc: Missed this. What do you mean by slide? In EVE code I found only slide_pattern, but it seems it can not be used here.

include/eve/detail/function/bit_cast.hpp
DenisYaroshevskiy: I see - what you are saying is that bit_casting from wide<char, 4> to wide<int, 1> should work, regardless of the type.

I'd agree with that - but this requires is very wrong. It'd have to say "sizeof(T) * N1::value == sizeof(U) * N2::value".

I'll have a look this weekend to make that change for you in master.

include/eve/detail/function/simd/riscv/bitmask.hpp
DenisYaroshevskiy: this doesn't strike me as a good idea.
sve doesn't do top_bits very well.

You want something (ideally an integer) that is all 1111s for true and all 0s for false.

Can risk-v do that

DenisYaroshevskiy (commented):
I'll be honest, this'll go much better if you split the pull request.
I'm not gonna insist - I really appreciate you putting in work.
But I'm also not going to let through things just because it's a big review, we care for the project.

We usually for a new platform, just make some targets build and then add more.

How would you feel about splitting out unit.arch.exe? unit.internals.exe? unit.memory.exe? Then we do some chunk and do more in the next bit.

include/eve/detail/function/bit_cast.hpp
ita-sc: Thanks.

Also, how should `bit_cast` be used on `logical`s?

I've used mask types (i.e. a pack of bits) as `logical`s for RISC-V.  Is `bit_cast` from `wide<char, 1>` to `logical<wide<int, 8>>` allowed (since sizes are equal)?

I ask since for some functions none of the overloads suited RISC-V (so I have written RISC-V specific ones).
E.G.: https://github.com/jfalcou/eve/blob/c1f74e3b5fe19a539bff8c20d05ba87192224e04/include/eve/detail/function/simd/common/bitmask.hpp#L50
Casts between `eve::logical<eve::wide<T, eve::fixed<N>>>` and `eve::wide<T, eve::fixed<N>>>`, but for RISC-V these types differ in length.

include/eve/detail/function/simd/riscv/bitmask.hpp
ita-sc: This is exactly what is done. I return `wide<T, N>`, where `T` is unsigned with appropriate size, filled with a `1` if corresponding bit of the `logical` was set, otherwise it will have `0`'s.

Representing  `logical` as a `wide<T,N>` with all ones for `true` and all zeroes for `false` will be ineffective for several reasons:
* before a mask operation I will need to cast it to the real mask type, which will be time-consuming
* I will need to use much more vector registers for keeping this logical (at most 8 instead of always 1 for real mask type).

include/eve/detail/function/simd/riscv/bit_cast.hpp
DenisYaroshevskiy: > , for "char" type we support using 1/8 part of real vector register (called mf8), but when you cast it to "int", you can at least use 1/2 part of register(mf2).

I'm not sure I understand what's happening here. 

4 bytes is 4 bytes. Both would be VLEN / 4 bytes. Why not?

DenisYaroshevskiy:
Is there any reason to use vlen > 1 over us just unrolling it manually?

ita-sc:
> I'll be honest, this'll go much better if you split the pull request.

I've split patch into several one. Should I create a separate review for first commit?

DenisYaroshevskiy:
Yeah - please separate review.

include/eve/detail/function/simd/riscv/bit_cast.hpp
ita-sc: As rvv intrinsics for RISC-V usually do not know about vlen, they should fit for any available vlen. Because of this fact, some types that are correspond for fractional usage for real vector register  are unavailable.

For example, you can not fit `int` with usage 1/4 of vector register of vlen=64. So in rvv intrinsics interface there are no types that could be used to represent usage 1/4 of vector register with the element type of `int`.

However, for `char` you can fit at least one element with usage 1/8 of vector register even if vlen is 64. So type that corresponds to usage of 1/8 of vector register with element type of char is available.

In current review I try to use RISC-V vector registers in the most effective way, so I use the smallest possible types for each `wide`, that can result into different sizes for underlying types, even if `sizeof(T1)*N1==sizeof(T2)*N2` because of described above.

include/eve/detail/function/simd/riscv/bit_cast.hpp
ita-sc: Done

test/unit/memory/load/tuple.cpp
DenisYaroshevskiy: I don't understand you. w8_t is 8 elements. This verifies that you can load 8 elements. Not 16. 

What fails and how.

test/unit/arch/top_bits.cpp
```diff
@@ -38,12 +38,15 @@ TTS_CASE_TPL( "Check top bits raw type", eve::test::simd::all_types)
 {
   using v_t         = eve::element_type_t<T>;
   using logical     = eve::logical<T>;
+  using c_t              = eve::cardinal_t<T>;
+  using rvv_logical_type = eve::logical<eve::wide<v_t, c_t>>;
```
DenisYaroshevskiy, line 42: This is just `logical`

test/unit/api/regular/wide.cpp
```diff
@@ -32,7 +32,8 @@ TTS_CASE_WITH( "Check eve::wide enumerating constructor"
...
+  if constexpr( T::size() < eve::fundamental_cardinal_v<v_t> && !eve::has_emulated_abi_v<T>
+                && eve::current_api != eve::rvv )
```
DenisYaroshevskiy, line 36: What doesn't work for rvv here. This seems like it should.

include/eve/detail/shuffle_v2/simd/riscv/shuffle_l2.hpp
```diff
@@ -0,0 +1,53 @@
...
+  else
+  {
+    std::array<int_ty, N::value> ids;
+    for( auto id = 0; id < N::value; ++id )
+    {
+      size_t id_from_pattern = id / G;
+      ids[id]                = P::idxs[id_from_pattern] * G + id % G;
+    }
+    wide<as_integer_t<T, unsigned>, N> offsets(ids);
+    wide<T, N>                         res = __riscv_vrgather_tu(x, x, offsets.storage(), N::value);
```
DenisYaroshevskiy, line 32: Gather is level 3 - it needs a mask. 

Here is how you do it: https://github.com/jfalcou/eve/blob/5c21a5b76ccbf5c00cccaa852421c0746df3990d/include/eve/detail/shuffle_v2/simd/arm/sve/shuffle_l3.hpp#L30

Let me know if that doesn't make sense, and I'll help more.

PR #1717, opened 2023-12-30
pedantic hypot correction

jtlap:
the actual implementation had an overflow bug for calls like pedantic(hypot)(5.0,1.0e-165), the new implementation corrects the problem and is far simpler

### DIFF
--- a/test/doc/math/pedantic/hypot.cpp
+++ b/test/doc/math/pedantic/hypot.cpp
@@ -8,12 +8,15 @@ using wide_ft = eve::wide<float, eve::fixed<4>>;
 
 int main()
 {
-  wide_ft pf = {-1.0f, eve::inf(eve::as(1.0f)), -3.0f, eve::valmax(eve::as<float>())/2};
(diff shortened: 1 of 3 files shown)


test/doc/math/pedantic/hypot.cpp
```diff
@@ -8,12 +8,15 @@ using wide_ft = eve::wide<float, eve::fixed<4>>;
 
 int main()
 {
-  wide_ft pf = {-1.0f, eve::inf(eve::as(1.0f)), -3.0f, eve::valmax(eve::as<float>())/2};
+  wide_ft pf = {-1.0f, eve::inf(eve::as(1.0f)), -3.0f, eve::smallestposval(eve::as<float>())/2};
```
jfalcou, line 11: Why changing the test instead of adding one ?

jtlap: it's a doc test. so I wanted to show the case that failed before. but if you want more outputs I can do that.

jfalcou (approved):

PR #1718, opened 2024-01-01
arm pb definitely due to poor denormals handling

jtlap:
also putting pedantic hypot to 1ulp

### DIFF
--- a/test/unit/module/math/hypot.cpp
+++ b/test/unit/module/math/hypot.cpp
@@ -72,6 +72,8 @@ TTS_CASE_TPL("Check return types of hypot", eve::test::simd::ieee_reals)
   TTS_EXPR_IS(pedantic(hypot)(v_t(), int(), T()), T);
   TTS_EXPR_IS(pedantic(hypot)(v_t(), T(), int()), T);
   TTS_EXPR_IS(pedantic(hypot)(v_t(), v_t(), v_t()), v_t);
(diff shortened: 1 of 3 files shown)


test/unit/module/math/hypot.cpp
```diff
@@ -72,6 +72,8 @@ TTS_CASE_TPL("Check return types of hypot", eve::test::simd::ieee_reals)
   TTS_EXPR_IS(pedantic(hypot)(v_t(), int(), T()), T);
   TTS_EXPR_IS(pedantic(hypot)(v_t(), T(), int()), T);
   TTS_EXPR_IS(pedantic(hypot)(v_t(), v_t(), v_t()), v_t);
+  std::cout << hypot(0.0014, 1.1e+02) << std::endl;
```
jfalcou, line 75: remove

test/unit/module/math/hypot.cpp
```diff
@@ -116,28 +118,31 @@ TTS_CASE_WITH("Check corner-cases behavior of eve::hypot variants on wide",
 <typename T>(T const& a0)
 {
   auto cases = tts::limits(tts::type<T> {});
-
+#ifdef SPY_ARCH_IS_ARM
```
jfalcou, line 121: why do we hide more tests than before ?

jfalcou (approved):

PR #1720, opened 2024-01-04
Feat/refurbished constants

jtlap:
up to now all math constants


include/eve/module/core/constant/eps.hpp
```diff
@@ -7,22 +7,37 @@
...
 {
+template<typename Options>
+struct eps_t : constant_callable<eps_t, Options, downward_option, upward_option>
+{
+  template<typename T>
+  static EVE_FORCEINLINE constexpr T value(eve::as<T> const&, auto const&)
+  {
+    using e_t = element_type_t<T>;
+
+    if      constexpr(std::integral<e_t>        ) return T(1);
```
DenisYaroshevskiy, line 24: eps 1 for integral looks a bit weird? Can we maybe just have `eps` for floating point? Or is this how numeric_limits define ti?

On that note - should we forward to std::numeric_limits?

include/eve/module/core/constant/eps.hpp
```diff
@@ -7,22 +7,37 @@
...
-
-#include <type_traits>
+#include <eve/arch.hpp>
+#include <eve/traits/overload.hpp>
+#include <eve/module/core/decorator/core.hpp>
 
 namespace eve
 {
+template<typename Options>
+struct eps_t : constant_callable<eps_t, Options, downward_option, upward_option>
```
DenisYaroshevskiy, line 17: Why are we supporting these oiptions when they don't do anything?

include/eve/module/math/constant/epso_2.hpp
```diff
@@ -7,10 +7,39 @@
...
+//       if constexpr(Opts::contains(upward2))        return T(0x1.552c98p-2);
+//       else if constexpr(Opts::contains(downward2)) return T(0x1.552c96p-2);
+//       else                                         return T(0x1.552c98p-2);
+//     }
+//     else
+//     {
+//       if constexpr(Opts::contains(upward2))        return T(0x1.552c97fa03696p-2);
+//       else if constexpr(Opts::contains(downward2)) return T(0x1.552c97fa03695p-2);
+//       else                                         return T(0x1.552c97fa03695p-2);
+//     }
```
DenisYaroshevskiy, line 34: should the commented code be deleted?

include/eve/module/math/constant/inv_2eps.hpp
```diff
@@ -7,10 +7,34 @@
...
+  static EVE_FORCEINLINE constexpr T value(eve::as<T> const&, Opts const&)
+  {
+    if constexpr(std::same_as<element_type_t<T>, float>)
+    {
+      return T(0x1p+22);
+    }
+    else
+    {
+      return T(0x1p+51);
+    }
```
DenisYaroshevskiy, line 29: extra `{}` ?

include/eve/module/math/constant/invcbrt_pi.hpp
```diff
@@ -7,10 +7,38 @@
...
+    if constexpr(std::same_as<element_type_t<T>, float>)
+    {
+      if constexpr(Opts::contains(upward2))        return T(0x1.5d95ep-1);
+      else if constexpr(Opts::contains(downward2)) return T(0x1.5d95dep-1);
+      else                                         return T(0x1.5d95ep-1);
+    }
+    else
+    {
+      if constexpr(Opts::contains(upward2))        return T(0x1.5d95df6bd2aeep-1);
+      else if constexpr(Opts::contains(downward2)) return T(0x1.5d95df6bd2aedp-1);
```
DenisYaroshevskiy, line 31: It seems to me like `upward` and `downward` should be one flag `up, down, default` or just `up, down` since they are mutually exclusive.

The interface would still be `[eve::upward]` but then the second one would override the first one.

DenisYaroshevskiy (commented):
Looks cool! Have some comments - lemme know what you think.

include/eve/module/core/constant/eps.hpp
jfalcou: This is the generalized epsilon defined as the smallest positive value so that `1 + eps != 1` and this definition works for integers

include/eve/module/math/constant/invcbrt_pi.hpp
jfalcou: This is true and should be done this way. I will modify so it works this way.
In fact, they are part of the `upward`, `downward` , `to_nearest` and `toward_zero` which are all mutually exclusive.
We will change this

include/eve/module/core/constant/eps.hpp
jfalcou: This is the generalized epsilon defined as the smallest positive value so that `1 + eps != 1` and this definition works for integers

include/eve/module/core/constant/eps.hpp
jtlap: where K is the mathematical value
We use that sometimes for instance for trigonometry
pi is a quite weird case because the nearest double and the nearest float are not on the same size.
It is also a hook for interval arithmetic

include/eve/module/math/constant/inv_2eps.hpp
jtlap: You're right

include/eve/module/math/constant/epso_2.hpp
jtlap: it's a mistake it will be deleted

include/eve/module/core/constant/eps.hpp
jtlap: You're right they are not needed if the three values are the same, but already it is sure that among these three values 2 at least are equal. For me it's orthogonality sugar for the all equal case.

include/eve/module/math/constant/epso_2.hpp
jtlap: done

jfalcou (approved):

PR #1722, opened 2024-01-07
slide: part #0

DenisYaroshevskiy:
only implemented sse2 - sse4.2. 

Probably should merge and finish later, no need to keep it in a branch.

### DIFF
--- a/test/doc/core/named_shuffles/slide_left.cpp
+++ b/test/doc/core/named_shuffles/slide_left.cpp
@@ -1,0 +1,5 @@
+int
+main()
+{
+  // in progress
+}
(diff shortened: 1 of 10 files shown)


test/doc/core/named_shuffles/slide_left.cpp
```diff
@@ -0,0 +1,5 @@
+int
```
jfalcou, line 1: is it normal ?

DenisYaroshevskiy: Yes

jfalcou (approved):

PR #1726, opened 2024-01-12
Add fine grained support for PPC sub-API

jfalcou:
SPY was updated to discriminate POWER sub version but we didnt.
 + Added the requried new ppc API object + fixed some code testing for those specifically.
 +  `is_supported.cpp` also displays more PPC API in the test.


DenisYaroshevskiy:
Do we really want this? Like do you want multiple ppc cis? It will take effort. For what  goal?

jfalcou:
> Do we really want this? Like do you want multiple ppc cis? It will take effort. For what goal?

The main culprit is that currently `eve::current_api` is shown as undefined when printed as vmx and vsx are not printed out by SPY.
It doesn't change any code in a drastic way, it's just there so we have the capability later if needed. The PR is complete as if. I am not going over every PPC function rn to fine grain chnage the implementation.

DenisYaroshevskiy (approved):
Ok. I do not like how we approach ppc much. We don't take it seriously but we don't drop it either. Seems ify but ok

jfalcou:
> Ok. I do not like how we approach ppc much. We don't take it seriously but we don't drop it either. Seems ify but ok

well, let's say that's one step toward taking it more seriously.

DenisYaroshevskiy:
I'm not convinced more seriously is what we want

jfalcou:
This is something we could discuss indeed. But I think fixing the issue that SPY and EVE detection/display of the sub API being wrong is something to do.

PR #1736, opened 2024-02-02
Reinstate MSVC tests on new runners

jfalcou:


DenisYaroshevskiy:
Please add to build and test.

DenisYaroshevskiy (approved):
Even without instructions approved

PR #1745, opened 2024-02-18
slide 5

DenisYaroshevskiy:
avx2 is not done but it's better than it was.

### DIFF
--- a/include/eve/detail/shuffle_v2/simd/x86/shuffle_l4_l5.hpp
+++ b/include/eve/detail/shuffle_v2/simd/x86/shuffle_l4_l5.hpp
@@ -48,11 +48,32 @@ shuffle_l4_l5_x86_put_u64x2_in_position(P, fixed<G>, wide<T, N> x)
   {
     constexpr auto p0 = get<0>(*P::shuffle_16in16);
     constexpr auto p1 = get<1>(*P::shuffle_16in16);
-    auto [r0, l0] = shuffle_v2_core(x, eve::lane<G>, idxm::to_pattern<p0>());
-    auto [r1, l1] = shuffle_v2_core(r0, eve::lane<G>, idxm::to_pattern<p1>());
+    auto [r0, l0]     = shuffle_v2_core(x, eve::lane<G>, idxm::to_pattern<p0>());
+    auto [r1, l1]     = shuffle_v2_core(r0, eve::lane<G>, idxm::to_pattern<p1>());
 
     return kumi::tuple {r1, idxm::add_shuffle_levels(l0, l1)};
   }
+}
+
+template<typename P, typename T, typename N, std::ptrdiff_t G>
+EVE_FORCEINLINE auto
+shuffle_l4_l5_x86_slide_less_than_16(P, fixed<G>, wide<T, N> x)
+{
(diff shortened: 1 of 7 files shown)


include/eve/detail/shuffle_v2/simd/x86/shuffle_l4_l5.hpp
```diff
@@ -48,13 +48,34 @@ shuffle_l4_l5_x86_put_u64x2_in_position(P, fixed<G>, wide<T, N> x)
...
+shuffle_l4_l5_x86_slide_less_than_16(P, fixed<G>, wide<T, N> x)
+{
+  constexpr auto no = kumi::tuple {no_matching_shuffle, eve::index<-1>};
+  // Coudn't figure out how to generalize well
+  // only slide left for now
+  // No masking 0s on avx512
+  if constexpr( current_api < avx2 ) return no;
+  else if constexpr ( constexpr auto slide = idxm::is_slide_left(P::idxs) )
+  {
+    static_assert(G == 1, "verifying assumptions");
```
jfalcou, line 69: if the static_assert is that empty of infos, why not just use the single parameter static_assert and not put a reason ?

test/unit/api/regular/shuffle_v2/idxm.cpp
```diff
@@ -780,6 +799,32 @@ TTS_CASE("is_slide_right")
...
+  };
+
+  test(std::array {0, 1, 2, 3}, 4, 0); // not slide 2, result is unimportant
+  test(std::array {1, 2, 3, 4}, 4, 1);
+  test(std::array {1, 2, 3, 16}, 16, 1);
+  test(std::array {1, we_, 3, 4}, 4, 1);
+  test(std::array {1, we_, 3, 16}, 16, 1);
+  test(std::array {2, 3, 4, we_}, 4, 2);
+  test(std::array {2, 3, 16, we_}, 16, 2);
+  // {2, 3, we_, we_} - is not an important case because it's not slide2 at all.
```
jfalcou, line 818: how so ?

include/eve/detail/shuffle_v2/simd/x86/shuffle_l4_l5.hpp
DenisYaroshevskiy: Well, if it fails for a user, I want them to see that it's internal failure.

Also, now looking at it, this message is bad for that

test/unit/api/regular/shuffle_v2/idxm.cpp
DenisYaroshevskiy: It's a one register shuffle. So it's not slide2

jfalcou: OK

include/eve/detail/shuffle_v2/simd/x86/shuffle_l4_l5.hpp
jfalcou: I'm OK if we go with a better message. I usually try to include "[EVE]" in the message so we know where it comes from

jfalcou (approved):

PR #1746, opened 2024-02-24
FMA has a new style functor

jfalcou:
Fixed options handling, removed old calls


include/eve/module/core/regular/fma.hpp
```diff
@@ -1,18 +1,37 @@
...
 
 namespace eve
 {
-//================================================================================================
+template<typename Options>
+struct fma_t : elementwise_callable<fma_t, Options, pedantic_option, promote_option>
+{
+  template<eve::value T,eve::value U,eve::value V>
+  requires(Options::contains(promote2))
+  constexpr EVE_FORCEINLINE auto operator()(T a, U b, V c) const noexcept { return EVE_DISPATCH_CALL(a,b,c); }
```
DenisYaroshevskiy, line 24: Return type! Always return type in declaration

PR #1754, opened 2024-03-10
New style/math

jtlap:


include/eve/module/math/regular/deginrad.hpp
```diff
@@ -47,8 +58,21 @@ namespace eve
...
+  EVE_FORCEINLINE constexpr T
+  deginrad_(EVE_REQUIRES(cpu_), O const &, T const& a) noexcept
+  {
+    if constexpr( has_native_abi_v<T> )
+    {
+      auto ridh = ieee_constant<0x1.1de0000p-6f, 0x1.1df46a0000000p-6>(eve::as<T>{});
+      auto ridl = ieee_constant<0x1.46a2520p-18f, 0x1.294e9c8ae0ec6p-33>(eve::as<T>{});
+      return fma(a, ridl, a * ridh);
+    }
+    else return apply_over(deginrad, a);
```
DenisYaroshevskiy, line 76: @jfalcou - doesn't element wise do this?
Even if it doesn't. I would expect the `if constexpr` to be unnecessary. Is there a reason why it's necessary?

include/eve/module/math/regular/exp.hpp
```diff
@@ -83,7 +94,77 @@ namespace eve
...
+        auto k = c;
+        x      = fnma(c, Log_2hi, x); // x-c*L
+        if constexpr( std::is_same_v<elt_t, float> )
+        {
+          x      = fnma(c, Log_2lo, x);
+          auto y = eve::reverse_horner(x, T(0x1.000000p-1f), T(0x1.55534ap-3f), T(0x1.5552aep-5f)
+                                      , T(0x1.131b16p-7f), T(0x1.6ef19ep-10f));;
+          c      = inc(fma(y, sqr(x), x));
+        }
+        else if constexpr( std::is_same_v<elt_t, double> )
```
DenisYaroshevskiy, line 148: why will we be here? If the previous if is `native abi` already? What are we trying to achieve?

include/eve/module/math/regular/deginrad.hpp
jfalcou: It is  indeed in flux. We found elementwise_callable needs a slight more massaging to get there.
This is noted and will be done after this PR so we can get rid of apply_over everywhere.

PR #1756, opened 2024-03-12
Make elementwise_callable more efficient

jfalcou:
Aggregation of function returning tuple now is handled generically.
This make apply_over easily replaceable within elementwise callable.

This allow us to remove a bunch of if has_native_abi branches.


jfalcou:
Main changes are : 
 +  include/eve/traits/overload/default_behaviors.hpp
 + include/eve/detail/skeleton.hpp
 
 where the reorganization of detecting aggregation/emulation/regualr call have been done.
 THe change in skeleton is made so that clang doesn't report incorrect call in the middle of a mapping decision instead of doing SFINAE. This is the great classic behavior we already have to avoid elsewhere.
 
 Rest of the PR is functios beign changed to remove the `has_native_abi` checks and random bug fix that this change set uncovered

include/eve/detail/skeleton.hpp
```diff
@@ -188,37 +207,43 @@ namespace eve::detail
...
     }
-    // Recursive case
     else
     {
-      // We end up there if we have a difference of # of replications inside
-      // This happens in conversions context or when we call a function on a
-      // AVX/AVX2 type with no implementation.
-      return wide_t{EVE_FWD(f)(lower(EVE_FWD(ts))...),
-                    EVE_FWD(f)(upper(EVE_FWD(ts))...)};
+      using wide_t = typename half_t::template rescale<typename half_t::cardinal_type::combined_type>;
```
DenisYaroshevskiy, line 218: put it together with half_t. And add `res` in the name somewhere plz or I a very confused

include/eve/detail/skeleton.hpp
```diff
@@ -188,37 +207,43 @@ namespace eve::detail
...
+
+      // Full sliceable case
+      if constexpr( is_fully_sliceable<wide_t,std::decay_t<Ts>...>() )
```
DenisYaroshevskiy, line 221: I think this tries to handle different cardinals.

the number of elements in all wides has to be the same. Otherewise it's not element wise.

UPD: or is it trying to do something else? I don't get it. All of our aggrgation is arrays of 2. What's with indexes?

include/eve/detail/skeleton.hpp
```diff
@@ -188,37 +207,43 @@ namespace eve::detail
...
+      {
+        wide_t that;
+
+        // If everything can be sliced in equal sub-parts, we save compile-time and
+        // simplify debugging by iterating over all smallest sub-parts
+        detail::apply<replication<wide_t>()>
+        ( [&]<typename... I>(I const&...)
```
DenisYaroshevskiy, line 228: no force inline is a problem

test/unit/CMakeLists.txt
```diff
@@ -85,13 +85,12 @@ add_dependencies(unit.exe unit.combinatorial.exe )
...
+
+if(NOT DEFINED EVE_TEST_NO_BESSEL)
 add_custom_target(unit.bessel.exe         )
 add_dependencies(unit.exe unit.bessel.exe )
 glob_unit("unit" ${unit_root} "module/bessel/*.cpp" )
-
-# add_custom_target(unit.complex.exe         )
-# add_dependencies(unit.exe unit.complex.exe )
-# glob_unit("unit" ${unit_root} "module/complex/*.cpp" )
+endif()
```
DenisYaroshevskiy, line 93: If the test doesn't work, let's disable it in code by looking at spy? that macro in targets is not visible most of the time and is annoying to build

test/unit/module/polynomial/laguerre.cpp
```diff
@@ -46,7 +46,7 @@ TTS_CASE_WITH("Check behavior of laguerre on wide",
   for( unsigned int n = 0; n < 5; ++n )
   {
     auto std_laguerre = [&](auto i, auto) { return NAMESPACE::laguerre(n, a0.get(i)); };
-    TTS_ULP_EQUAL(eve__laguerrev(n, a0), T(std_laguerre), 2100);
+    TTS_ULP_EQUAL(eve__laguerrev(n, a0), T(std_laguerre), 6000);
```
DenisYaroshevskiy, line 49: __ is rerved for std

.github/workflows/unit.yml
```diff
@@ -209,19 +209,19 @@ jobs:
       fail-fast: false
       matrix:
         cfg:
-        - { comp: clang_libcpp, arch: x86     , opts: -msse2                                }
+        - { comp: clang_libcpp, arch: x86     , test: "-DEVE_TEST_NO_BESSEL=1", opts: -msse2                                }
```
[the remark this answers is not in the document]
jfalcou: Libcpp doesn't have the bessel special functions, so I disable the bessel teats for them

test/unit/CMakeLists.txt
jfalcou: No. Complex and quaternion has been removed ages ago but didn't cleaned up the cmake

test/unit/module/polynomial/laguerre.cpp
jfalcou: In the middle of names ?

include/eve/detail/skeleton.hpp
```diff
@@ -188,37 +207,43 @@ namespace eve::detail
...
-                                                , that.storage()
-                                                , EVE_FWD(ts)...
-                                                )
-            ),...
-          );
-        }
-      );
-
-      return that;
+      return tuple_combiner( EVE_FWD(f)(lower(EVE_FWD(ts))...), EVE_FWD(f)(upper(EVE_FWD(ts))...));
```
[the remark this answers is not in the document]
jfalcou: Some functions return tuples and we need to take care of them when we aggregate. This code replace the 50+ lines from apply_over.hpp. nothing new.

include/eve/detail/skeleton.hpp
```diff
@@ -131,6 +135,21 @@ namespace eve::detail
...
+  };
+
+  template<typename T>
+  EVE_FORCEINLINE auto tuple_combiner(T const& lo, T const& hi)
+  {
+    using out_t = kumi::apply_traits_t<combiner,T>;
+    return kumi::map_index([](auto i, auto const& l, auto const& h) { return kumi::element_t<i,out_t>{l,h}; }
+                          , lo, hi
+                          );
+  }
```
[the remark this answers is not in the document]
jfalcou: No. We have to aggregate functions that return a tuple of wide.
To do so, we slice the inputs, computes the resulting tuples and combine their subparts.
We can use eve::combine here for dependencies reasons so we just use the combining wide constructor, hence the beed of the subtype.

include/eve/detail/skeleton.hpp
```diff
@@ -54,7 +54,11 @@ namespace eve::detail
 
   // Compute a transformed wide type
   template<typename F, typename... Ts>
-  struct wide_result
+  struct wide_result;
+
+  template<typename F, typename... Ts>
+  requires requires { std::declval<F>()(at(std::declval<Ts>(), 0)...); }
```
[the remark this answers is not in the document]
jfalcou: We need to sfinae the case the scalar type deduction is non sensicak. This requires is just a way to easily prune out the wrong cases.

include/eve/detail/skeleton.hpp
jfalcou: Can certainly do

include/eve/detail/skeleton.hpp
jfalcou: This is an old piece of code that days that if we have aggregate of more than 2 pieces and all inputs are all aggregate, we walk down the pieces array directly instead of going recursive.

This was an old fix we designed to eas aggregation compile time.

.github/workflows/unit.yml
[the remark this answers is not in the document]
DenisYaroshevskiy: oh, ok

include/eve/detail/skeleton.hpp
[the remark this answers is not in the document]
DenisYaroshevskiy: I think i need more explaining here. let's talk on slack

include/eve/detail/skeleton.hpp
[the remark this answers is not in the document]
DenisYaroshevskiy: so, it's aggregating all components in a tuple? can I have a comment

include/eve/detail/skeleton.hpp
DenisYaroshevskiy: that's not a thing anymore though, is it?

test/unit/module/polynomial/laguerre.cpp
DenisYaroshevskiy: I believe so

include/eve/detail/skeleton.hpp
jfalcou: Oh this is indeed an oversight from ages ago, will fix

test/unit/module/polynomial/laguerre.cpp
jfalcou: will fix it then, not a big deal. But I fear we have that in a lot of palce in special too.

include/eve/detail/skeleton.hpp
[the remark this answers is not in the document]
jfalcou: From Slack:

 Map has ot be called on emulated types, think `wide<double>` on ARM32 for ex. The principles of `map(f,x...)` is to compute the result as being a wide of the proper cardinal and type which is computed as the SCALAR return type of f.  So we do compute `decltype( f( <every element_type of all x>() ... )`

However, as the EVE computing model is that function on scalar only exists if all their parameters are the same, you may end up with scenario where calling map would be an error.

Case in point :  let's say you're on ARM32 and you want to call `eve::max( 1, some_wide_double );`
As `wide<double>` is emulated on ARM32, we go through map. And the map code tries to compute `decltype(max(int,double));` which is not a thing.

So this type mistake is captured by the require, it ADL out map and we go through the regular function call. Guess what, the regular call see int  + wide<double> and has code to call itself with max(wide<double>, wide<double>)
This will also trigger a map deduction that this time will work and do the correct thing.

include/eve/detail/skeleton.hpp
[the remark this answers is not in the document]
jfalcou: There is two case of aggregation : aggregation over multiple register and AVX aggregation for integer wide than can be sliced in two because AVX don't have much operation on integers.

We store aggregates as an array of K regular native register, K being 2,4,8 or w/e. So if we end up in the case of two actual aggregate, we slice it up over those K slices as it is faster to compile. In the other case, we do the recursive slice and call

For the tuple return, a bunch of functions like sincos, rempio2, airy etc return a tuple of wide. Now, the real question is should it be a wide of tuple but it is out of scope for this PR.

So the previous version had special apply_over variant like apply_over2, apply_over3 that handled apply_over over function that returned a tuple so that aggregate worked correctly. The tuple_combiner is a streamlined and more compact way of doing it that also happen to work whatever the size of the output tuple.

If you think tuple of wide as return type should go, then maybe we cna devise that afterward but I needed the existing code to work atm.

include/eve/detail/skeleton.hpp
jfalcou: See https://github.com/jfalcou/eve/pull/1756#discussion_r1527145937

include/eve/detail/skeleton.hpp
```diff
@@ -125,7 +129,9 @@ namespace eve::detail
     auto parts = kumi::make_tuple(slicer(ts)...);
 
     // Apply f on both side of the slices and re-combine
-    using wide_t = typename wide_result<Func, Ts...>::type;
+    using half_result_t = decltype(f(get<0>(slicer(ts))...));
+    using wide_t = typename half_result_t::template rescale<typename half_result_t::cardinal_type::combined_type>;
+
     return kumi::apply([&f](auto... m) { return wide_t { f(get<0>(m)...), f(get<1>(m)...)}; }, parts);
```
DenisYaroshevskiy, line 135: kumi apply returns kumi tuple. I thought we  should retiurn wide. Am I wrong?

include/eve/detail/skeleton.hpp
```diff
@@ -79,12 +83,12 @@ namespace eve::detail
     template<typename Func, typename Idx, typename... Ts>
     EVE_FORCEINLINE auto operator()(Func &&fn, Idx const &i, Ts &&... vs) const noexcept
```
DenisYaroshevskiy, line 84: why not i by value.

include/eve/detail/skeleton.hpp
```diff
@@ -79,12 +83,12 @@ namespace eve::detail
...
     {
-      return EVE_FWD(fn)(at(EVE_FWD(vs), i)...);
```
DenisYaroshevskiy, line 82: can we name at as smth longer? like `get_at`? I don't like adl look upabale name that is this short and not part of an api

include/eve/traits/overload/default_behaviors.hpp
```diff
@@ -20,7 +21,29 @@ namespace eve
...
+  //!   This protocol includes options checks and management and proper level of error reporting.
+  //!   eve::callable is the most general base class for defining such a @callable without having to manually handles
+  //!   all these details.
+  //!
+  //!   @tparam Func          Type of current @callable being implemented.
+  //!   @tparam OptionsValues Type of stored options.
+  //!   @tparam Options       List of supported option specifications.
+  //!
+  //!   @groupheader{Example}
+  //!   @godbolt{doc/traits/callable_object.cpp}
```
DenisYaroshevskiy, line 45: i think this file is already there, right?

include/eve/traits/overload/default_behaviors.hpp
```diff
@@ -61,38 +84,38 @@ namespace eve
...
 
     template<callable_options O, typename T, typename... Ts>
-    EVE_FORCEINLINE constexpr auto behavior(auto arch, O const& opts, T x0,  Ts const&... xs) const
-    requires(match_option<condition_key,O,ignore_none_>)
+    constexpr EVE_FORCEINLINE auto adapt_call(auto a, O const& o, T x,  Ts const&... xs) const
     {
-      constexpr bool supports_call = requires{ func_t::deferred_call(arch, opts, x0, xs...); };
-      constexpr bool any_simd     = (simd_value<T> || ... || simd_value<Ts>);
+      constexpr bool has_implementation = requires{ func_t::deferred_call(a, o, x, xs...); };
+      constexpr bool is_emulated        = requires{ map(this->derived(), x, xs...); };
```
DenisYaroshevskiy, line 102: more meaningful name please

include/eve/traits/overload/default_behaviors.hpp
```diff
@@ -61,38 +84,38 @@ namespace eve
...
-        if constexpr((has_aggregated_abi_v<Ts> || ...)) return aggregate(this->derived(), x0, xs...);
-        else                                            return map(this->derived(), x0, xs...);
-      }
-      else
-      {
-        static_assert(any_simd, "[EVE] - Missing scalar implementation for current callable");
-        return ignore{};
-      }
+      if      constexpr(any_aggregated)               return aggregate(this->derived(), x, xs...);
+      else if constexpr(any_emulated && is_emulated ) return map(this->derived(), x, xs...);
```
DenisYaroshevskiy, line 107: Not having 'is_emulated' be true is a bug in eve, right?

so !is_emulated should report a bug here and not fall back

include/eve/traits/overload/default_behaviors.hpp
```diff
@@ -61,38 +84,38 @@ namespace eve
...
+      else if constexpr(has_implementation)           return func_t::deferred_call(a, o, x, xs...);
+      else                                            return ignore{};
```
DenisYaroshevskiy, line 109: this is an indication of bug in eve, right?
I feel ify about checking for internal bugs with a sfinae like this.

include/eve/detail/skeleton.hpp
```diff
@@ -34,7 +34,11 @@ namespace eve::detail
 
   // Compute a transformed wide type
   template<typename F, typename... Ts>
-  struct wide_result
+  struct wide_result;
+
+  template<typename F, typename... Ts>
+  requires requires { std::declval<F>()(eve::detail::at(std::declval<Ts>(), 0)...); }
```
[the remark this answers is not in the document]
jfalcou: I have to investigate but now I fixed aggregate tonot depend on it and we do convertion better, it is maybe unneeded.

include/eve/detail/skeleton.hpp
jfalcou: apply takes a function f and a tupel of Ai and call f(Ai...), it returns w/e f is returning, here a wide

include/eve/detail/skeleton.hpp
jfalcou: because we didn't back in the day, will change.

include/eve/detail/skeleton.hpp
jfalcou: I actually changed it to be fully qualified. But I can rename no pb

include/eve/traits/overload/default_behaviors.hpp
jfalcou: I just added the documentation for it. I don't get your remark.

jfalcou: OK, the file is already there

include/eve/traits/overload/default_behaviors.hpp
jfalcou: As per slack discussion, this stays

include/eve/detail/skeleton.hpp
jfalcou: Renamed to get_at

include/eve/module/bessel/regular/cyl_bessel_kn.hpp
```diff
@@ -44,16 +44,15 @@ namespace eve
   //!   @code
   //!   namespace eve
   //!   {
-  //!      template< eve::ordered_value N, eve::floating_ordered_value T >
-  //!      T cyl_bessel_kn(N n, T x) noexcept;
+  //!     template<eve::value N, eve::floating_value T>
+  //!     constexpr T cyl_bessel_kn(N n, T x) noexcept;
```
DenisYaroshevskiy, line 48: why. is it really constexpr

include/eve/traits/overload/default_behaviors.hpp
DenisYaroshevskiy: It's not `is emulated`

include/eve/detail/function/simd/common/subscript.hpp
```diff
@@ -89,7 +89,7 @@ namespace eve::detail
     {
       // g++ need that
       typename Wide::value_type data[Wide::size()];
-      auto s = p.storage();
+      [[maybe_unused]] auto s = p.storage();
```
DenisYaroshevskiy, line 92: why?

DenisYaroshevskiy (approved):

include/eve/detail/function/simd/common/subscript.hpp
jfalcou: I once compiled by mistake using g++ 13 in release mode and got a warning. I fixed it then realized it was out of our scope. If you want me to remove it I can.

PR #1797, opened 2024-05-12
iterate selected

DenisYaroshevskiy:
from recent discussion - iteration over all elements, selected in a mask.

Will maybe add the algorithm in the same pr later, for now - can be reviewed and I would like to see CI.

### DIFF
--- a/include/eve/concept/invocable.hpp
+++ b/include/eve/concept/invocable.hpp
@@ -7,6 +7,7 @@
 //==================================================================================================
 #pragma once
 
+#include <concepts>
 #include <type_traits>
 
 namespace eve
@@ -82,4 +83,15 @@ concept invocable = requires(F&& f, Args&&...args)
 {
   typename detail::get_invoke_result<F&&, Args&&...>::type;
 };
+
+  //================================================================================================
+  //! @ingroup simd_concepts
+  //! @brief std::predicate but doesn't require regularity
+  //!
+  //! @tparam T
+  //================================================================================================
+  template <typename T, typename ... Args>
+  concept irregular_predicate = requires (T&& f, Args&&... args) {
+    { !f(EVE_FWD(args)...) } -> std::convertible_to<bool>;
+  };
 }
--- a/include/eve/module/core/regular/impl/iterate_selected.hpp
+++ b/include/eve/module/core/regular/impl/iterate_selected.hpp
@@ -1,0 +1,89 @@
+//==================================================================================================
+/*
+  EVE - Expressive Vector Engine
+  Copyright : EVE Project Contributors
+  SPDX-License-Identifier: BSL-1.0
+*/
+//==================================================================================================
+#pragma once
+
+#include <eve/arch/top_bits.hpp>
+#include <eve/conditional.hpp>
+
+#include <bit>
+
+namespace eve::detail
+{
+
+template<typename F> struct plus_offset_lambda
+{
+  F&             f;
+  std::ptrdiff_t offset;
+
+  EVE_FORCEINLINE
+  bool operator()(std::ptrdiff_t i) const { return f(i + offset); }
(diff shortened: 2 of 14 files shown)


include/eve/concept/invocable.hpp
```diff
@@ -82,4 +83,15 @@ concept invocable = requires(F&& f, Args&&...args)
...
 };
+
+  //================================================================================================
+  //! @ingroup simd_concepts
+  //! @brief std::predicate but doesn't require regularity
+  //!
+  //! @tparam T
+  //================================================================================================
+  template <typename T, typename ... Args>
+  concept irregular_predicate = requires (T&& f, Args&&... args) {
```
jfalcou, line 94: why not relaxed_predicate ?

include/eve/module/core/regular/impl/iterate_selected.hpp
```diff
@@ -0,0 +1,92 @@
...
+{
+  F&             f;
+  std::ptrdiff_t offset;
+
+  EVE_FORCEINLINE
+  bool operator()(std::ptrdiff_t i) const { return f(i + offset); }
+};
+
+template<std::ptrdiff_t bits_per_element, std::unsigned_integral Bits>
+bool
```
jfalcou, line 28: no FORCE_INLINE ?

include/eve/module/core/regular/impl/simd/arm/sve/iterate_selected.hpp
```diff
@@ -0,0 +1,41 @@
...
+  {
+    return iterate_selected[ignore_none](l & top_bits<L>{ignore}, f);
+  }
+  else
+  {
+    auto m = l.storage;
+    while( any(m) )
+    {
+      L ignored_before = svbrka_z(sve_true<element_type_t<L>>(), m);
+      m = svbic_z(sve_true<element_type_t<L>>(), m, ignored_before);
```
jfalcou, line 32: why no use bitand_not ?

include/eve/module/core/regular/iterate_selected.hpp
```diff
@@ -0,0 +1,94 @@
...
+  EVE_CALLABLE_OBJECT(iterate_selected_t, iterate_selected_);
+};
+
+//================================================================================================
+//! @addtogroup core_simd
+//! @{
+//!    @var iterate_selected
+//!    @brief a utility to do scalar iteration over all true elements in a `logical`.
+//!
+//!    TODO: @see eve::algo::for_each_selected
```
jfalcou, line 47: Remove the TODO I guess

include/eve/concept/invocable.hpp
DenisYaroshevskiy: Well, std::predicate is not working for me, because it requires `regular`.
so this is `irregular` :) I think it makes sense

include/eve/module/core/regular/impl/simd/arm/sve/iterate_selected.hpp
DenisYaroshevskiy: Good idea. I tried both bit_andnot and logical_andnot - something didn't work. I don't have the capacity to figure out at the moment. Let it go for now?

include/eve/concept/invocable.hpp
jfalcou: We have relaxed_range hence the idea. But I guess this is ok too.

DenisYaroshevskiy: I understand yes. But I don't like it.

Tbh - I think - this should be predicate and the standard one should be regular_predicate. But we are here

jfalcou: OK

jfalcou (approved):

Issue #1749, opened 2024-03-01
[FEATURE] Add requires on multiple argument functions to ensure proper cardinal compatibility


jfalcou:
To day, if we call `pow(x,y)` with x and y being of different cardinals, the error message is atrocious.
I sugegst we : 
- [x] Write a `has_compatible_cardinals<Ts...>` helper that checks that all cardinals are the same for non-scalar types in `Ts...`
- [x] Add a rersuires on this in all 2+ parametrs functions using the new dispatch system

jfalcou:
For reference 

code

```c++
auto x = eve::pow( wide<float,fixed<4>>{}, wide<float,fixed<8>>{});
```

clang output :

```c++
<source>:8:12: error: no matching function for call to object of type 'const detail::callable_object<tag::pow_>'
  auto x = eve::pow( wide<float,fixed<4>>{}, wide<float,fixed<8>>{});
           ^~~~~~~~
/home/falcou/dev/eve/include/eve/module/math/regular/pow.hpp:95:1: note: candidate template ignored: constraints not satisfied [with Args = <eve::wide<float, eve::fixed<4>>, eve::wide<float, eve::fixed<8>>>]
EVE_MAKE_CALLABLE(pow_, pow);
^
/home/falcou/dev/eve/include/eve/detail/overload.hpp:137:3: note: expanded from macro 'EVE_MAKE_CALLABLE'
  EVE_IMPLEMENT_CALLABLE(TAG,NAME)                                                                  \
  ^
/home/falcou/dev/eve/include/eve/detail/overload.hpp:127:3: note: expanded from macro 'EVE_IMPLEMENT_CALLABLE'
  EVE_DECLARE_CALLABLE(TAG,NAME)                                                                    \
  ^
/home/falcou/dev/eve/include/eve/detail/overload.hpp:79:48: note: expanded from macro 'EVE_DECLARE_CALLABLE'
      EVE_FORCEINLINE constexpr decltype(auto) operator()(Args &&... args) const noexcept          \
                                               ^
/home/falcou/dev/eve/include/eve/module/math/regular/pow.hpp:95:1: note: because 'tag_dispatchable<eve::detail::callable_object<eve::tag::pow_, type-parameter-0-0>::tag_type, eve::wide<float, eve::fixed<4> >, eve::wide<float, eve::fixed<8> > >' evaluated to false
[...]

PR #1798, opened 2024-05-13
Partial #1749 - Implements same_lanes and same_lanes_or_scalar

jfalcou:


test/unit/meta/traits/same_lanes_or_scalar.cpp
```diff
@@ -0,0 +1,43 @@
+//==================================================================================================
+/**
+  EVE - Expressive Vector Engine
+  Copyright : EVE Project Contributors
+  SPDX-License-Identifier: BSL-1.0
+**/
+//==================================================================================================
+#include "test.hpp"
```
DenisYaroshevskiy, line 8: If the implementations are in the same, so should be tests I think

jfalcou: True

PR #1802, opened 2024-05-17
Issue1749

jtlap:


jfalcou (commented):
Without sounding lika jerk, the requires would be better above the operator()

jtlap:
what does it change, apart to completely redo the work ?

PR #1811, opened 2024-05-21
logicals_xxx

jtlap:


jfalcou (changes requested):
We added noexcept to some but not all of thos elogical functions, so we need to be throughout. I  will merge once fixed.

include/eve/module/core/regular/logical_and.hpp
```diff
@@ -83,10 +84,9 @@ namespace eve
   namespace detail
   {
     template<typename T, typename U, callable_options O>
+    requires(eve::same_lanes_or_scalar<T, U>)
```
jfalcou, line 87: The requires is on the operator() not the implementation

jfalcou (approved):

jfalcou (approved):

PR #1814, opened 2024-05-22
new popcount

jtlap:


include/eve/module/core/regular/impl/simd/arm/sve/popcount.hpp
```diff
@@ -13,26 +13,26 @@
 
 namespace eve::detail
 {
-template<unsigned_scalar_value T, typename N>
-EVE_FORCEINLINE auto
-popcount_(EVE_SUPPORTS(sve_), wide<T, N> v) noexcept -> wide<T, N>
+template<unsigned_scalar_value T, typename N, callable_options O>
+EVE_FORCEINLINE wide<T, N> popcount_(EVE_SUPPORTS(sve_),
```
jfalcou, line 17: this should be `EVE_REQUIRES`

include/eve/module/core/regular/impl/simd/arm/sve/popcount.hpp
```diff
@@ -13,26 +13,26 @@
...
 {
-  return popcount[ignore_none](v);
+  return  svcnt_x(sve_true<T>(), v);
 }
 
-template<conditional_expr C, unsigned_scalar_value T, typename N>
-EVE_FORCEINLINE auto
-popcount_(EVE_SUPPORTS(sve_), C const& cond, wide<T, N> v) noexcept -> wide<T, N>
+template<conditional_expr C, unsigned_scalar_value T, typename N, callable_options O>
+EVE_FORCEINLINE wide<T, N> popcount_(EVE_SUPPORTS(sve_),
```
jfalcou, line 26: this should be `EVE_REQUIRES`

include/eve/module/core/regular/popcount.hpp
```diff
@@ -7,10 +7,24 @@
...
+#include <eve/traits/overload.hpp>
+#include <eve/module/core/decorator/core.hpp>
+#include <eve/detail/assert_utils.hpp>
 
 namespace eve
 {
+  template<typename Options>
+  struct popcount_t : elementwise_callable<popcount_t, Options>
+  {
+    template<integral_value T>
```
jfalcou, line 19: You didnt changed the concept on the implementation so callign with a signed integer will lead to mapping.

jfalcou: Also the implementation in SVE and NEON are non consistent.
Either you fix them or you go back to unsigned_value here.

jfalcou (changes requested):

include/eve/module/core/regular/popcount.hpp
jtlap: I fixed sve and x86, but there is nothing in neon, so did you mean x86?

jfalcou: Yes x86. My issue is that it is not fixed and it only works because we do test on unsigned only.
In SVE, svcnt_x requires unsigned integer types yet the concept is integer not unsigned_integer.
If you want to keep it this way, hwich different from what STD does, you need a bit_cast in the implementation. I thnik we should go back to popcoutn only working on unsigned values and remove the concept in the implemenation.

jfalcou (approved):

PR #1820, opened 2024-05-27
sort

jtlap:


include/eve/module/core/regular/sort.hpp
```diff
@@ -7,10 +7,26 @@
...
+#include <eve/traits/overload.hpp>
+#include <eve/module/core/decorator/core.hpp>
 
 namespace eve
 {
+  template<typename Options>
+  struct sort_t : callable<sort_t, Options>
+  {
+    template<eve::value T, typename Less>
+    constexpr EVE_FORCEINLINE T operator()(T v, Less l) const noexcept
```
DenisYaroshevskiy, line 20: There should be a concept for less. Like maybe `eve::strict_weak_ordering<T>`
I'm not sure where that one should be declared.

Also probably not value but simd_value, I don't think scalars work

jtlap: ok

DenisYaroshevskiy: I guess? Also should be the same as std for scalar probably?

jfalcou: yean it's a bit complicated. What fi we fix that in another PR ?

DenisYaroshevskiy (approved):

PR #1822, opened 2024-05-27
[DRAFT] CI: skip jobs if no changes to source/tests/cmake are detected

SadiinsoSnowfall:
Skip the CI checks if nothing that touches the source, the tests or the way the project is compiled has changed.


jfalcou:
Let's keep that in the fridge

PR #1824, opened 2024-05-28
Add CPP version guard to eve

SadiinsoSnowfall:
Added to `eve/module/core.hpp` instead of `eve/eve.hpp` because not everyone use the later as the main eve header.


jfalcou (approved):

jfalcou:
Think it is worth a shot. @DenisYaroshevskiy your opinion ?

DenisYaroshevskiy (approved):

PR #1829, opened 2024-05-30
Support unsigned masked min/max intrinsics on x86

SadiinsoSnowfall:


include/eve/module/core/regular/impl/simd/x86/max.hpp
jtlap, line 1: not a problem, but why adding blank before parameters parenthesis?

jtlap (approved):

PR #1833, opened 2024-06-01
Update external dependies management

jfalcou:


DenisYaroshevskiy:
Is this just moving files?

jfalcou:
Nah, if you look at the end of the detail file, you'll find the `__has_include` thing we discussed

```c++
#define KUMI_NO_STD_ADAPTORS
#if __has_include(<kumi/tuple.hpp>)
# include <kumi/tuple.hpp>
#else
# if defined(EVE_USE_EXTERNAL_DEPENDENCIES)
#   error "[EVE] External KUMI dependencies requested but none is available"
#   include <kumi/tuple.hpp>
# else
#   include <eve/deps/kumi/tuple.hpp>
# endif
```

DenisYaroshevskiy:
Figure out some ci for external deps and I'll stump this. 
Noting here too that the necessity to use NO_STD_ADAPTORS will cause odr violations.

jfalcou:
> Noting here too that the necessity to use NO_STD_ADAPTORS will cause odr violations.
I propose we remove it later by fixing the actual EVE code that is now ambiguous in a later PR

DenisYaroshevskiy (approved):

PR #1838, opened 2024-06-05
Transparent types support in eve::wide

SadiinsoSnowfall:
TODO:

- [ ] what to do with eq/neq operators
- [ ] what to do with the other operators

### DIFF
--- a/include/eve/arch/arm/neon/as_register.hpp
+++ b/include/eve/arch/arm/neon/as_register.hpp
@@ -96,7 +96,7 @@ namespace eve
 
   // ---------------------------------------------------------------------------------------------
   // logical cases
-  template<typename T, typename Size, arm_abi ABI>
+  template<arithmetic_scalar_value T, typename Size, arm_abi ABI>
   struct  as_logical_register<T, Size, ABI>
         : as_register<as_integer_t<T, unsigned>, Size, ABI>
   {};
--- a/include/eve/arch/arm/sve/as_register.hpp
+++ b/include/eve/arch/arm/sve/as_register.hpp
@@ -89,7 +89,7 @@ namespace eve
 
   // ---------------------------------------------------------------------------------------------
   // logical cases
-  template<typename Type, typename Size, sve_abi ABI>
+  template<arithmetic_scalar_value Type, typename Size, sve_abi ABI>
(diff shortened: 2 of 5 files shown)


DenisYaroshevskiy:
I think this needs a discussion before doing anything. 
enums are not special, char, uintptr_t or smth like that is similar.
+ needs to handle so that the amount of work required on the changes is minial - so probably callable should do smart things.

jfalcou:
> I think this needs a discussion before doing anything. enums are not special, char, uintptr_t or smth like that is similar.
> 
> * needs to handle so that the amount of work required on the changes is minial - so probably callable should do smart things.

This preliminary work is mostly so that `wide<ENUM>` works in term of basic functionality. @SadiinsoSnowfall put soemthign in place that minimizes the amoutn of touch up in non common files and arch specific ones. The question we have is that we can't have any function works on any enum cause MAYBE for the user it make no sense.

So the plan of this draft is to explore what we should do, startign with basic features. We can chat about what we should be doing further in a call.

DenisYaroshevskiy:
Well, for example, existance of logical<wide<enum>> is not obvious to me. 
There is no logicals for bundles, for example.

Also - I am pretty sure f16 won't work if you try to use the same mechanism as enum.
I would like some design doc or at lesat ideas outline - otherwise I believe we will rush into something that we don't fully understand.

jfalcou:
> Well, for example, existance of logical<wide> is not obvious to me. There is no logicals for bundles, for example.
> 
> Also - I am pretty sure f16 won't work if you try to use the same mechanism as enum. I would like some design doc or at lesat ideas outline - otherwise I believe we will rush into something that we don't fully understand.
> 
> UPD: if you just want to prototype to see what's up - no problem. But please - discussion before you publish this pr as pr.


Yup sure, this draft is there so @SadiinsoSnowfall  can see the CI and how it reacts.
We need a call.

DenisYaroshevskiy:
A post or a call - just some ideas/goals.

DenisYaroshevskiy:
Hi.

I think this pull request is growing a lot. It is preferable to not put one big patch all at once. Do you think you can split?

DenisYaroshevskiy:
There is 36 files changed and you are still working. It doesn't have to work all or nothing. You can do a bit

Issue #1817, opened 2024-05-23
[FEATURE] transform_copy_if and friends


DenisYaroshevskiy:
We need a moral equivalent of this: https://godbolt.org/z/3oYnsf8fe
for implace and implace versions.

We also need to migrate the existing `copy_if` `remove_if` `remove` to it.
And we need an implace version of `copy_if`

jfalcou:
Can this be closed ?

PR #1842, opened 2024-06-07
add algo::transform_copy_if

Temulgeh:
Implementation of #1817 (partially, no in-place version yet) 

Ended up not migrating `copy_if` and similar functions since `transform_copy_if` uses `compress_store` instead of `compress_copy`.

Wrote some documentation and moved the tests for `copy_if` to `transform_copy_if` as per Denis' advice.

I wonder if the name is alright, since it sort of implies that the function acts like `transform` followed by `copy_if` but it's actually more like the opposite, `copy_if` followed by `transform`.

Tried *roughly* measuring the performance and it seems to be faster than `eve::algo::copy_if` and `eve::algo::transform_to` separately but slower than `std::copy_if` followed by `std::transform`. I don't really know how to properly benchmark things though. Also, maybe the in-place version will be faster?

### DIFF
--- a/include/eve/module/algo/algo/copy_if.hpp
+++ b/include/eve/module/algo/algo/copy_if.hpp
@@ -96,6 +96,9 @@ template<typename TraitsSupport> struct copy_if_ : TraitsSupport
 //!   #include <eve/module/algo.hpp>
 //!   @endcode
 //!
+//!   @note
+//!   If you need to apply a transformation, you can use `eve::algo::views::map` or `eve::algo::transform_copy_if`.
+//!
 //!   The main difference from std::copy_if is that it accepts output range and not
 //!   an output iterator. If the output doesn't have enough space, algorithm will fill
 //!   all of the avaliable output and then stop.
--- a/include/eve/module/algo/algo/copy_if.hpp
+++ b/include/eve/module/algo/algo/copy_if.hpp
@@ -135,6 +135,9 @@ template<typename TraitsSupport> struct copy_if_ : TraitsSupport
 //!   ** Return value **
 //!
 //!    relaxed_iterator past the last written element.
+//!
+//!   @see `transform_copy_if`
+//!
 //! @}
 //================================================================================================
 inline constexpr auto copy_if = function_with_traits<copy_if_>;
--- a/include/eve/module/algo/algo/transform_copy_if.hpp
+++ b/include/eve/module/algo/algo/transform_copy_if.hpp
@@ -104,15 +104,15 @@ namespace eve::algo
   //!
   //!   @brief Similar to applying `eve::transform_to` and `eve::copy_if` at the same time.
   //!
+  //!   @note
+  //!   If the scalar operation is cheap enough, `::copy_if` + `views::map` might be slightly faster.
+  //!
   //!   Conditionally copies values from an input range to an output range,
   //!   transforming them in the process.
   //!
   //!   If the output range is too small, fills all the available space and then stops.
   //!   
   //!   If the output range's element type is different from the type of the values returned by the transforming function, performs the appropriate conversions.
-  //!
-  //!   @note
-  //!   If the scalar operation is cheap enough, `::copy_if` + `views::map` might be slightly faster.
   //!
   //!   @groupheader{Callable Signatures}
   //!
--- a/include/eve/module/algo/algo/transform_copy_if.hpp
+++ b/include/eve/module/algo/algo/transform_copy_if.hpp
@@ -13,9 +13,6 @@
 #include <eve/module/algo/algo/views/convert.hpp>
 #include <eve/module/core.hpp>
 
-#include <utility>
-#include <type_traits>
-
 namespace eve::algo
 {
 
@@ -23,32 +20,29 @@ namespace eve::algo
   {
     template<typename TraitsSupport> struct transform_copy_if_ : TraitsSupport
     {
-      template<typename OutIt, typename Func> struct delegate
+      template<typename O, typename F> struct delegate
       {
-        OutIt out_first;
-        OutIt out_last;
-        Func func;
-
-        static constexpr auto compress_copy()
-        {
-          auto density = density_for_compress_copy<typename TraitsSupport::traits_type>();
-          return eve::compress_copy[eve::unsafe][density];
-        }
+        O of;
+        O ol;
+        F func;
 
         EVE_FORCEINLINE
         bool tail(auto f, auto ignore)
         {
-          auto out_ignore   = eve::keep_first(std::min(out_last - out_first, eve::iterator_cardinal_v<OutIt>));
-          auto loaded       = eve::load(f);
+          auto out_ignore   = eve::keep_first(std::min(ol - of, eve::iterator_cardinal_v<O>));
+          auto loaded       = eve::load[ignore](f);
           auto [vals, mask] = func(loaded);
-          out_first = eve::compress_store[eve::unsafe][ignore][out_ignore](vals, mask, out_first);
-          return out_first == out_last;
+          using out_t = eve::element_type_t<decltype(vals)>;
+          auto cvt_and_store_it = views::convert(of, eve::as<out_t>{});
+          auto write_end = eve::compress_store[eve::unsafe][ignore][out_ignore](vals, mask, cvt_and_store_it);
+          of += write_end - cvt_and_store_it;
+          return of == ol;
         }
 
         EVE_FORCEINLINE
         std::ptrdiff_t left_for_stage1() const
         {
-          return out_last - out_first;
+          return ol - of;
         }
 
         EVE_FORCEINLINE
@@ -56,7 +50,10 @@ namespace eve::algo
         {
           auto loaded       = eve::load(f);
           auto [vals, mask] = func(loaded);
-          out_first = eve::compress_store[eve::unsafe](vals, mask, out_first);
+          using out_t = eve::element_type_t<decltype(vals)>;
+          auto cvt_and_store_it = views::convert(of, eve::as<out_t>{});
+          auto write_end = eve::compress_store[eve::unsafe](vals, mask, cvt_and_store_it);
+          of += write_end - cvt_and_store_it;
           return false;
         }
 
@@ -64,11 +61,14 @@ namespace eve::algo
         bool step_2(auto f)
         {
           // ol - of < cardinal
-          auto out_ignore   = eve::keep_first(out_last - out_first);
+          auto out_ignore   = eve::keep_first(ol - of);
           auto loaded       = eve::load(f);
           auto [vals, mask] = func(loaded);
-          out_first = eve::compress_store[eve::unsafe][eve::ignore_none][out_ignore](vals, mask, out_first);
-          return out_first == out_last;
+          using out_t = eve::element_type_t<decltype(vals)>;
+          auto cvt_and_store_it = views::convert(of, eve::as<out_t>{});
+          auto write_end = eve::compress_store[eve::unsafe][eve::ignore_none][out_ignore](vals, mask, cvt_and_store_it);
+          of += write_end - cvt_and_store_it;
+          return of == ol;
         }
       };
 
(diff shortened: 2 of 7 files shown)


DenisYaroshevskiy:
Awesome - will have a look in a bit. I keep benchmarks separately and I have some copy_if becnhmarks. 
For this algorithm, I wouldn't worry much about benchmarks tbh (we don't benchmark everything - too expensive).
But if you want to - I can help you set that up

include/eve/module/algo/algo/transform_copy_if.hpp
```diff
@@ -0,0 +1,169 @@
...
+        {
+          auto density = density_for_compress_copy<typename TraitsSupport::traits_type>();
+          return eve::compress_copy[eve::unsafe][density];
+        }
+
+        EVE_FORCEINLINE
+        bool tail(auto f, auto ignore)
+        {
+          auto out_ignore   = eve::keep_first(std::min(out_last - out_first, eve::iterator_cardinal_v<OutIt>));
+          auto loaded       = eve::load(f);
```
DenisYaroshevskiy, line 42: not ok to just load from f - you missed ignore. This should trigger a test

include/eve/module/algo/algo/transform_copy_if.hpp
```diff
@@ -0,0 +1,169 @@
...
+      };
+
+      template<relaxed_range In, relaxed_range Out, typename Func>
+      EVE_FORCEINLINE
+      auto operator()(In&& in, Out&& out, Func func) const -> unaligned_iterator_t<Out>
+      {
+        if( in.begin() == in.end() || out.begin() == out.end() )
+          return out.begin();
+
+        using out_type = eve::element_type_t<std::remove_reference_t<decltype(get<0>(func(std::declval<wide<value_type_t<In>>>())))>>;
```
DenisYaroshevskiy, line 82: No, this is not right. This code means you miss a test for precise wide size being accepted.

You cannot call:

```
func(wide<int>{})
```

If the user passed in:

```
func(wide<int, eve::fixed<2>>)
```

That's just how we do this.

Solution:

You preprocess without convert and then convert later.
See transform
https://github.com/jfalcou/eve/blob/8969c8242499d6a91274931bfa16797254507464/include/eve/module/algo/algo/transform.hpp#L38

include/eve/module/algo/algo/transform_copy_if.hpp
```diff
@@ -0,0 +1,169 @@
...
+namespace eve::algo
+{
+
+  namespace detail
+  {
+    template<typename TraitsSupport> struct transform_copy_if_ : TraitsSupport
+    {
+      template<typename OutIt, typename Func> struct delegate
+      {
+        OutIt out_first;
```
DenisYaroshevskiy, line 28: I would prefer if you follow the existing naming style.

test/doc/algo/transform_copy_if.cpp
```diff
@@ -0,0 +1,31 @@
...
+#include <vector>
+
+int main()
+{
+  std::vector<int> input(16);
+  eve::algo::iota(input, 0);
+  std::cout << "Input vector:\n  " << tts::as_string(input) << "\n\n";
+
+  auto func = [](eve::wide<int> e) {
+    return kumi::make_tuple(-e, e % 2 == 0);
```
DenisYaroshevskiy, line 15: e reads like a math constant, plz call this x.

test/doc/algo/transform_copy_if.cpp
```diff
@@ -0,0 +1,31 @@
...
+  };
+
+  std::vector<int> output_big(42);
+  auto end_big = eve::algo::transform_copy_if(input, output_big, func);
```
DenisYaroshevskiy, line 19: please show the full erase idiom

```
output_big.erase(
  eve::algo::transform_copy_if(input, output, [](eve::like<int> auto x) {
    return kumi::tuple{ -x, eve::is_even(x) );
  });
```

Temulgeh:
the documentation tests that aren't passing seem to be from a clang bug? [as seen here](https://stackoverflow.com/questions/64300832/why-does-clang-think-gccs-subrange-does-not-satisfy-gccs-ranges-begin-functi)

tried it on a new version and it seems to compile just fine

Temulgeh:
i could rewrite the example to not use std::ranges::subrange

test/doc/algo/transform_copy_if.cpp
```diff
@@ -0,0 +1,31 @@
...
+#include <ranges>
+#include <vector>
+
+int main()
+{
+  std::vector<int> input(16);
+  eve::algo::iota(input, 0);
+  std::cout << "Input vector:\n  " << tts::as_string(input) << "\n\n";
+
+  auto func = [](eve::wide<int> e) {
```
DenisYaroshevskiy, line 14: it's eve::nofs_wide unfortunately. That's a separate can of worms - just use eve::like<int> auto.

include/eve/module/algo/algo/transform_copy_if.hpp
```diff
@@ -0,0 +1,168 @@
...
+        if( in.begin() == in.end() || out.begin() == out.end() )
+          return out.begin();
+
+        auto [processed_in, processed_out] = temporary_preprocess_ranges_hack(
+          TraitsSupport::get_traits(), in, out);
+
+        auto iteration = two_stage_iteration(processed_in.traits(), processed_in.begin(), processed_in.end());
+
+        auto of = unalign(processed_out.begin());
+        auto ol = unalign(processed_out.end());
```
DenisYaroshevskiy, line 88: I don't like that we do conversions in all callbacks. 

Let's figure out conversion here once or - hide it in some function.

include/eve/module/algo/algo/transform_copy_if.hpp
```diff
@@ -0,0 +1,168 @@
...
+  //!   @brief Similar to applying `eve::transform_to` and `eve::copy_if` at the same time.
+  //!
+  //!   Conditionally copies values from an input range to an output range,
+  //!   transforming them in the process.
+  //!
+  //!   If the output range is too small, fills all the available space and then stops.
+  //!   
+  //!   If the output range's element type is different from the type of the values returned by the transforming function, performs the appropriate conversions.
+  //!
+  //!   There are two overloads: one that takes in two separate operation and predicate functions,
```
DenisYaroshevskiy, line 126: I don't think we want this one. Is it really beneficial?

Temulgeh: the version with two functions still seemed to be faster (?) and it was convenient when testing things out, but if you think it's not a good interface choice we can remove it

Temulgeh:
i was thinking about names for in-place `copy_if` and i think `keep_if` makes sense

so the in-place version of `transform_copy_if` could be something like `transform_keep_if`

however that might be inconsistent with the already existing `transform_to` and `transform_inplace`. maybe something like `transform_to_if` and `transform_inplace_if` could also work?

DenisYaroshevskiy:
As much as I'm not excited about `keep_if` - I can't come up with anything better, since filter has to be a view.

I think `keep_if`, `transform_keep_if`, `transform_copy_if` is the way we go. Yuck.

include/eve/module/algo/algo/transform_copy_if.hpp
DenisYaroshevskiy: Try this one maybe please? I think it'd be cleaner.

include/eve/module/algo/algo/transform_copy_if.hpp
DenisYaroshevskiy: > still seemed to be faster

It's a bug somewhere if it is, Shouldn't be. (Probably measurement issue).
I don't think we want it.

include/eve/module/algo/algo/transform_copy_if.hpp
```diff
@@ -0,0 +1,164 @@
...
+  //!   * The first form gives the same input to both functions.
+  //!     You can think of it as applying `eve::copy_if` and then `eve::transform_to`.
+  //!   * The second form is especially useful when the predicate and the transforming operation
+  //!     have some overlap or depend on each other.
+  //!
+  //!   @groupheader{Callable Signatures}
+  //!
+  //!   @code
+  //!   namespace eve::algo
+  //!   {
```
DenisYaroshevskiy, line 133: Misses name of the function

Temulgeh: when you say "try this one" do you mean passing a converting view to preprocess_ranges or to two_stage_iteration? i can try this later tonight

DenisYaroshevskiy: I meant to figure out what to convert to once. I think it'll look nicer

DenisYaroshevskiy: You can't pass converting virw to preprocessing range, you don't know the type

test/unit/module/algo/algorithm/copy_if_generic.cpp
```diff
@@ -6,10 +6,21 @@
...
 
-#include "copy_if_generic_test.hpp"
+#include "transform_copy_if_generic_test.hpp"
+
+template <typename TraitsSupport>
+struct copy_if_ignore_op_ : TraitsSupport
+{
+  auto operator()(auto&& in, auto&& out, auto, auto p)
+  {
+    return eve::algo::copy_if(in, out, p);
```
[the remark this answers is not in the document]
DenisYaroshevskiy: still  a problem, you lose any traits/options passed to the wrapping algorithm.

Here is how you do it: https://github.com/jfalcou/eve/blob/0ef778e9d567e89a1feaa5ced90c115ac4330a33/test/unit/module/algo/algorithm/find_if_not_generic.cpp#L21

include/eve/module/algo/algo/transform_copy_if.hpp
```diff
@@ -0,0 +1,144 @@
...
+        return eve::unalign(out.begin()) + (d.of - convert_out.begin());
+      }
+
+    };
+  }
+
+  //==============================================================================================
+  //! @addtogroup algos
+  //! @{
+  //!   @var transform_copy_if
```
DenisYaroshevskiy, line 97: I believe the code is OK.

Now - you need to cross reference in the docs with `copy_if` and you also need to mention that `copy_if` + `views::map` might be slightly faster if the scalar operation is cheap enough

include/eve/module/algo/algo/copy_if.hpp
```diff
@@ -135,6 +135,9 @@ template<typename TraitsSupport> struct copy_if_ : TraitsSupport
 //!   ** Return value **
 //!
 //!    relaxed_iterator past the last written element.
+//!
+//!   @see `transform_copy_if`
+//!
```
DenisYaroshevskiy, line 140: I don't think this is a good place.

After defined in the header

Note: if you need to apply a transformation, you can use eve::views::map view or eve:: algo::transform_copy_uf

Temulgeh:
i don't understand why this one test didn't pass, i will push a dummy commit to restart the tests

include/eve/module/algo/algo/transform_copy_if.hpp
DenisYaroshevskiy: please mention these

DenisYaroshevskiy (commented):
Let me know if you need help figuring out the failing test

Temulgeh:
seems to have passed now, must have been a bug in the CI?

include/eve/module/algo/algo/transform_copy_if.hpp
DenisYaroshevskiy: This one still needs fixing.

jfalcou:
> seems to have passed now, must have been a bug in the CI?

it's a somehow random concept bug we have sometimes. Nothing from your side.

include/eve/module/algo/algo/transform_copy_if.hpp
Temulgeh: do you mean you want me to move it higher in the documentation?

DenisYaroshevskiy (approved):
I see what you did. LGTM.

jfalcou (approved):

jfalcou:
So is this a merge ?

DenisYaroshevskiy:
@Temulgeh merging?

Temulgeh:
i don't see anything i'd want to change so fine with it

PR #1850, opened 2024-06-11
interleave

jtlap:
interleave to new_style

### DIFF
--- a/include/eve/detail/function/interleave.hpp
+++ b/include/eve/detail/function/interleave.hpp
@@ -24,7 +24,6 @@ namespace eve
 
     template<simd_value T, std::same_as<T>... Ts>
     EVE_FORCEINLINE kumi::tuple<logical<T>, logical<Ts>...> operator()(logical<T> v, logical<Ts>... vs) const noexcept
-    requires( T::abi_type::is_wide_logical )
     { return EVE_DISPATCH_CALL(v, vs...); }
 
     EVE_CALLABLE_OBJECT(interleave_t, interleave_);
--- a/include/eve/detail/function/interleave.hpp
+++ b/include/eve/detail/function/interleave.hpp
@@ -7,12 +7,30 @@
 //==================================================================================================
 #pragma once
 
+
 #include <eve/arch.hpp>
-#include <eve/detail/overload.hpp>
+#include <eve/traits/overload.hpp>
 
 namespace eve
 {
-  EVE_MAKE_CALLABLE(interleave_, interleave);
+  template<typename Options>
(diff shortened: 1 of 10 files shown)


include/eve/detail/function/interleave.hpp
```diff
@@ -7,12 +7,30 @@
...
+  {
+
+    template<eve::simd_value T, std::same_as<T>... Ts>
+    EVE_FORCEINLINE kumi::tuple<T, Ts...> operator()(T v, Ts... vs) const noexcept
+    { return EVE_DISPATCH_CALL(v, vs...); }
+
+
+    template<simd_value T, std::same_as<T>... Ts>
+    EVE_FORCEINLINE kumi::tuple<logical<T>, logical<Ts>...> operator()(logical<T> v, logical<Ts>... vs) const noexcept
+    requires( T::abi_type::is_wide_logical )
```
jfalcou, line 27: this should never happen

jfalcou: the requries is an implemantiond etail, the () must work for all kind of logicals

jfalcou (changes requested):

include/eve/detail/function/interleave.hpp
jtlap: line suppressed

include/eve/detail/function/simd/common/interleave.hpp
```diff
@@ -16,16 +16,16 @@ namespace eve::detail
   // N-ary interleave
   // This is the main one as it is easier to specify optimisation on this interface than on tuple
   //================================================================================================
-  template<simd_value T, std::same_as<T>... Ts>
-  EVE_FORCEINLINE auto interleave_(EVE_SUPPORTS(cpu_),T v0, Ts... vs) noexcept
+  template<callable_options OO, simd_value T, std::same_as<T>... Ts>
+  EVE_FORCEINLINE auto interleave_(EVE_REQUIRES(cpu_), OO const&, T v0, Ts... vs) noexcept
```
DenisYaroshevskiy, line 20: Are we passing callable options by const& Is there any reason to? 

I'd expect by value but maybe it's important?

jtlap: I think that this is optimized out anyway here, as only the type is used

jfalcou: I think it is irrelevant as we carry mostly type inforamtions. I guess I was sloppy and jtlap mimicked me.
We can later have a cleaning session to be homogeneous

jfalcou (approved):

PR #1859, opened 2024-06-18
transparent types - traits & concepts

SadiinsoSnowfall:
First PR for supporting transparent types.

@DenisYaroshevskiy I have decided to split the main PR into multiple smaller ones. This is the first which introduced the traits and concepts required to tag types as transparent.

I think the second PR will be adding support for the operators, what do you think ?


include/eve/concept/scalar.hpp
```diff
@@ -110,10 +111,19 @@ concept product_scalar_value = detail::scalar_tuple<T>();
...
 concept arithmetic_scalar_value = plain_scalar_value<T> || product_scalar_value<T>;
 
+//================================================================================================
+//! @concept element_value
+//! @brief Represents any type that can be loaded into a eve::wide
+//! The concept `element_value<T>` is satisfied if and only if T satisfies either
+//! transparent_value or arithmetic_scalar_value.
+//================================================================================================
+ template <typename T>
+ concept element_value = transparent_value<T> || arithmetic_scalar_value<T>;
```
DenisYaroshevskiy, line 121: 1.
I'm not so sure about the name. I don't think it tells me anything.
You will also encounter `arithmetic_value` and then what? 

Let's think about it.

One option is to say that all enums and such are actually arithmentic values.
This is a bit ify, because `std::is_arithmetic` is false for enums.

I would suggest `extended_arithmetic_scalar_value`

include/eve/concept/transparent.hpp
```diff
@@ -0,0 +1,18 @@
...
+#pragma once
+
+#include <eve/traits/transparent.hpp>
+
+#include <concepts>
+
+namespace eve
+{
+  template <typename T>
+  concept transparent_value = !std::same_as<transparent_inner_t<T>, T>;
```
DenisYaroshevskiy, line 17: This seems like it will return `true` for `unique_ptr`.

Let's do a type trait, with explicit opt ins.

include/eve/traits/transparent.hpp
```diff
@@ -0,0 +1,28 @@
...
+#pragma once
+
+#include <type_traits>
+
+namespace eve
+{
+	// Default case: normal types do not have specific storage type
+	template <typename T>
+	struct transparent_trait {
+			using type = T;
```
DenisYaroshevskiy, line 17: what's with formatting.

We don't call them trait in c++.

You call them 

```
struct transparent;

transparent_t
```

I don't like the word transparent very much even if it's used in rust.
Here - let's do vote best of 3 - if you and @jfalcou both think that transparent is the word - fine.

I would go from `equivalent_of`, `identical_to` or smth.

test/unit/api/transparent/base.cpp
```diff
@@ -0,0 +1,36 @@
...
+{
+  enum class E: T { };
+
+  static_assert(!eve::transparent_value<T>, "T is not transparent");
+  static_assert(eve::transparent_value<E>, "E is transparent");
+  static_assert(std::same_as<eve::transparent_inner_t<E>, T>, "E inner type is T");
+};
+
+template <typename E>
+struct BS { E value; };
```
DenisYaroshevskiy, line 23: Let's not use BS as a name unless we have to.

include/eve/concept/transparent.hpp
jfalcou: How so, there is no ties between our traits ans unique ptr

include/eve/traits/transparent.hpp
jfalcou: Equivalent_tp sounds good to me

DenisYaroshevskiy:
> Not for operators, they go through specific code inside wide for obvious circular dependencies issues.

Absolutely not obvious.

test/unit/api/transparent/base.cpp
DenisYaroshevskiy: Use S for struct, that's fine.

jfalcou:
> > Not for operators, they go through specific code inside wide for obvious circular dependencies issues.
> 
> Absolutely not obvious.

Well.
Operators are friends of wide and call the self_* familly of functions that don't use callable system for the fact that the previous system depended on wide.

I am not disputing that maybe now we can do better but it is clearly not the scope of this PR nor the following ones.

include/eve/traits/transparent.hpp
jfalcou: > If we keep `equivalent_to` for the trait, maybe `has_equivalent` works better for the concept ? 
I agree

> And then `as_equivalent_t` and `as_equivalent` (to better differentiate with `equivalent_to`) for the inner type alias, and the cast function ?

I wish we keep name starting by `as` for the type wrapper familly of type. We can probably juse have `equivalent_to_t`
and some `equivalent_value` function.

jfalcou:
Naming scheme for all the names we don't like. I propose:

```c++
namespace eve
{
  // the concept
  template<typename T>
  concept has_equivalent = ...;

  // the extension point
  template<typename T>
  struct equivalent_to { using type= ...; };

  // extract the value (implies bit_cast)
  template<has_equivalent T>
  EVE_FORCEINLINE constexpr auto as_equivalent(T x) noexcept;

  // traits extracting the equivalent type
  template<has_equivalent T>
  using as_equivalent_t = typename equivalent_to<T>::type;
}
```

include/eve/concept/scalar.hpp
jfalcou: What about we say, that's a type can be put into a wide, so it is vectorizable.
What about `concetp vectorizable` ?.??

include/eve/detail/function/simd/common/make.hpp
```diff
@@ -36,20 +36,34 @@ namespace eve::detail
...
   {
     return make_emulated<logical<wide<T, N>>>(vs...);
   }
 
+  template<has_plain_translation T, typename N, typename... Vs>
+  EVE_FORCEINLINE auto make(eve::as<logical<wide<T, N>>> const &, Vs... vs) noexcept
+    requires std::same_as<abi_t<T, N>, emulated_>
+  {
+    return bit_cast(make(eve::as<logical<wide<translate_t<T>, N>>>{}, static_cast<translate_t<Vs>>(vs)...), as<logical<wide<T, N>>>{});
+  }
```
DenisYaroshevskiy, line 65: @jfalcou  - how much is this in the same boat as operators.

I think it might be beneficial if it was, so that we don't compile this for non-native types.

include/eve/traits/as_translation.hpp
```diff
@@ -0,0 +1,34 @@
...
+
+  template <typename T, typename N>
+  constexpr auto translate(wide<T, N> val) {
+    if constexpr (has_plain_translation<T>) {
+      return bit_cast(val, as<wide<translate_t<T>, N>>{});
+    } else {
+      return val;
+    }
+  }
+}
```
DenisYaroshevskiy, line 34: \n

include/eve/traits/as_translation.hpp
```diff
@@ -0,0 +1,34 @@
...
+#include <eve/detail/wide_forward.hpp>
+
+#include <bit>
+
+namespace eve {
+  template <typename V>
+  constexpr auto translate(V val) {
+    if constexpr (has_plain_translation<V>) {
+      static_assert((sizeof(V) == sizeof(translate_t<V>)) && (alignof(V) == alignof(translate_t<V>)));
+      return std::bit_cast<translate_t<V>>(val);
```
DenisYaroshevskiy, line 20: this doesn't support the non recursion definition.
It probably should.

include/eve/traits/translation.hpp
```diff
@@ -0,0 +1,28 @@
...
+
+	template <typename T>
+	using translate_t = typename translation_of<T>::type;
+
+	// Covers every enum
+	template <typename T>
+	requires (std::is_enum_v<T>)
+	struct translation_of<T>: std::underlying_type<T>
+	{ };
+}
```
DenisYaroshevskiy, line 28: formatting. Just use clang format.

DenisYaroshevskiy: Yes you can, per file

jfalcou:
good to merge ?

include/eve/concept/scalar.hpp
```diff
@@ -108,12 +109,12 @@ concept product_scalar_value = detail::scalar_tuple<T>();
 //! - `float`
 //==================================================================================================
 template<typename T>
-concept arithmetic_scalar_value = plain_scalar_value<T> || product_scalar_value<T>;
+concept arithmetic_scalar_value = plain_scalar_value<T> || product_scalar_value<T> || has_plain_translation<T>;
```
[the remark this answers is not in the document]
DenisYaroshevskiy: I think I meant documentation

include/eve/detail/function/simd/common/make.hpp
jfalcou: As stated earlier, we are going to migrate everything to callable as much as it makes sense.

jfalcou:
LGTM, @DenisYaroshevskiy ?

DenisYaroshevskiy (approved):
Hey. So this is missing translation to product types. but doing it as a follow up is ok

PR #1860, opened 2024-06-19
transparent types - wide manipulation

SadiinsoSnowfall:
Base load / slice / combine / get / set / make / etc...


include/eve/detail/function/simd/x86/make.hpp
```diff
@@ -32,48 +34,55 @@ namespace eve::detail
...
-
-          if constexpr( c == category::float64x2) return _mm_setr_pd(static_cast<T>(vs)...);
-    else  if constexpr( c == category::float32x4)
+    if constexpr(has_plain_translation<T>)
     {
-      return [&]<std::size_t... N>(std::index_sequence<N...> const&)
-      {
-        return _mm_setr_ps(vs..., (N ? 0:0)...);
-      }(std::make_index_sequence<4 - sizeof...(vs)>());
+      return make(as<wide<translate_t<T>, S>>{}, translate(vs)...);
```
DenisYaroshevskiy, line 39: I don't like that we have to do so much work in custom ways. I'd seriosly consider migrating make first.

include/eve/detail/function/simd/x86/slice.hpp
```diff
@@ -91,15 +95,25 @@ namespace eve::detail
...
   {
-    std::array<wide<T, typename N::split_type>, 2> that{slice(a, lower_), slice(a, upper_)};
-    return that;
+    if constexpr (has_plain_translation<T>)
+    {
+      using sub_t = as<as_wide_t<T, typename N::split_type>>;
+
+      auto [l, h] = bit_cast(a, as<wide<translate_t<T>, N>>{}).slice();
+      return std::array{bit_cast(l, sub_t{}), bit_cast(h, sub_t{})};
+    }
```
DenisYaroshevskiy, line 108: No - in x86 you definetly shouldn't do this. In generic code maybe. You probably can be helped with some better concepts.

Did you consider using `plain_scalar_value`

PR #1866, opened 2024-06-26
Basic operations for translation-aware types

jfalcou:


include/eve/arch/cpu/base.hpp
```diff
@@ -24,31 +24,32 @@ namespace eve::detail
...
     EVE_FORCEINLINE storage_type &      storage() &       noexcept { return data_; }
 
-    //! @brief Retrieve the architecture-specific storage help by wide
+    //! Retrieve the architecture-specific storage help by wide
     EVE_FORCEINLINE storage_type        storage() &&      noexcept { return data_; }
 
-    //! @brief Implicit conversion to the architecture-specific storage help by wide
-    EVE_FORCEINLINE operator storage_type const& () const &  noexcept { return data_; }
+    //! Implicit conversion to the architecture-specific storage help by wide
+    EVE_FORCEINLINE explicit(PreventConversion) operator storage_type const& () const &  noexcept { return data_; }
```
DenisYaroshevskiy, line 46: Why are we prevernting conversion here? what's the usecase?

jfalcou:
The strategy for that has changed. We're keeping the branch but it'll be dissected as other PR on the subjects advance.

PR #1867, opened 2024-06-26
Add fall-through checks to if-constexpr chains

SadiinsoSnowfall:
A lof ot code in EVE is composed of `if-constexpr` chains which do not have a final `else` branch. Most of the time this is not a problem as the implicit `else` branch is never reached. However, when debugging EVE or adding new functionalities, such branches may be reached and in some cases introduce UB in the program without any compiler diagnostic.

I propose to add a new macro, `EVE_UNREACHEABLE` which raises an error using `static_assert` when reached, and use this macro in every as-of-now implicit `else` `constexpr` branch to ensure that no fall-through ever occur without the user noticing.

As an example, calling an eve function with a type no covered in its `if-constexpr` chain leads to the following code gen under GCC:
```x86asm
test():
     brk
```
Under clang, no assembly is generated as the compiler consider the caller function UB.
With the macro, the following error is raised:
```c++
error: static assertion failed: [EVE] Fallthrough rejected
   43 |         else    EVE_UNREACHABLE();
```

This implies that most of the source files should be modified to accommodate this change, this PR only applies to the neon backend as a demonstration.

### DIFF
--- a/include/eve/detail/meta.hpp
+++ b/include/eve/detail/meta.hpp
@@ -15,6 +15,8 @@
 #include <cstdint>
 #include <cstddef>
 
+#define EVE_UNREACHABLE() []<bool c=false>(){static_assert(c, "[EVE] Fallthrough rejected");}()
+
 namespace eve::detail
 {
   // Values list helper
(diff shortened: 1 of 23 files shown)


DenisYaroshevskiy:
I have serious issues with this.

#1. Naming. Unreacheable has a vell established meaning - and it's code not reacheable at runtime. 
See https://en.cppreference.com/w/cpp/utility/unreachable
If you put static_assert(false) in there - that's not good.

#2. I'm not convinced by motivation.

In eve almost all of our functions (at least public ones) return an explicit type.
If you forget a branch, you'll get a warning: https://godbolt.org/z/4cfqnqqec

Exception might be if your header is in PCH - where CMAKE disables warnings.
There are hacks we can try to do to enable warnings in PCH, this is not the only problem.
Or you can work locally with no pch sometimes.

jfalcou:
Let self this one for now.

PR #1881, opened 2024-07-06
Make make callable

SadiinsoSnowfall:
Turn internal make into a callable


DenisYaroshevskiy:
It should still be in detail. + You might have a recursion problem, let me know if you need help

jfalcou (approved):

PR #1882, opened 2024-07-07
shuffling emulated doubles on neon is level 0

DenisYaroshevskiy:
Fixing the issue for slide_left for doubles on neon.

Conflict of 2 default shuffle levels + a special case: emulation is always 0, while just 1 constant is level 1.
emulation always 0 should win.


jfalcou (approved):

jfalcou:
Is thisbcomplete, if yes lgtm to merge

PR #1911, opened 2024-08-05
Adding sigmoid and chi

jtlap:


jtlap:
abel also was added in polynomial

include/eve/module/core/regular/chi.hpp
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    EVE_CALLABLE_OBJECT(chi_t, chi_);
+  };
+
+//================================================================================================
+//! @addtogroup core_arithmetic
+//! @{
+//!   @var chi
+//!   @brief `elementwise_callable` indicatrix of the interval \f$[lo, hi[\f$ or of the set for which the invocable returns true.
```
jfalcou, line 44: it is not elementwise

jfalcou (changes requested):

include/eve/module/core/regular/chi.hpp
jtlap: done

jfalcou (approved):

PR #1913, opened 2024-08-09
Make `callable` return lvalue reference if necessary

jfalcou:
This is a requirement for some external libraries that have callable that do stuff like

```c++
kyosu::real(z) = 9;
```

Where `real` is a function object based on `eve::callable`.


DenisYaroshevskiy:
Wait - no - this will mess up the enums

jfalcou:
Yes indeed, I'll fix the external libraries internally.

PR #1921, opened 2024-08-20
Make Slice callable

SadiinsoSnowfall:


jfalcou (approved):

jfalcou:
> Two SVE-512 tests timed out, for reference they both took more than 1 minute to compile on my M3 macbook. Is this something to be concerned about ?

care to make an issue about that ?

jfalcou (approved):

PR #1925, opened 2024-08-21
Callable convert

SadiinsoSnowfall:
just seeing if it passes CI for now.


include/eve/detail/function/simd/arm/sve/friends.hpp
```diff
@@ -59,8 +60,7 @@ requires(sve_abi<abi_t<T, N>> || sve_abi<abi_t<U, N>>)
 {
   if constexpr(!is_aggregated_v<abi_t<T, N>>)
   {
-    callable_object<tag::convert_> const cvt{};
-    return svmov_z(v, cvt(w, as<logical<T>>{}));
+    return svmov_z(v, convert(w, as<logical<T>>{}));
```
DenisYaroshevskiy, line 63: If the dependcy cycles cause you problems, concider have a forward declaration of some helper function.

See simd_cast pull request for an example

jfalcou: The include cycle was due to unrequired incldue elsewhere. This is actually removing useless anti-cycle thing.

PR #1927, opened 2024-08-22
Adapt constants for translated_type - Refactoring & adaptation

jfalcou:
Made `constant_callable` deals with translated type by makeing the specific value() interface to only deals with scalar basic value.
Cleaned up constants and related concepts/traits.

### DIFF
--- a/include/eve/module/core/constant/constant.hpp
+++ b/include/eve/module/core/constant/constant.hpp
@@ -9,11 +9,10 @@
 
 #include <eve/as.hpp>
 #include <eve/concept/value.hpp>
-#include <eve/detail/function/bit_cast.hpp>
-#include <eve/detail/implementation.hpp>
-#include <eve/detail/meta.hpp>
 
-#include <type_traits>
+#include <eve/detail/implementation.hpp>
+#include <eve/detail/function/bit_cast.hpp>
+#include <eve/traits/underlying_type.hpp>
 
 namespace eve
 {
@@ -53,13 +52,12 @@ namespace eve
 //! @}
 //================================================================================================
 
-template<typename T, auto BitsPattern>
-requires(plain_scalar_value<element_type_t<T>>)
+template<value T, auto BitsPattern>
 EVE_FORCEINLINE auto Constant(eve::as<T> const& = {})
 {
-  using t_t = element_type_t<T>;
+  using t_t = translate_element_type_t<T>;
 
-  if constexpr( std::is_integral_v<t_t> ) { return static_cast<T>(BitsPattern); }
+  if constexpr( std::integral<t_t> ) { return static_cast<T>(BitsPattern); }
   else
   {
     if constexpr( sizeof(t_t) != sizeof(BitsPattern) )
@@ -68,7 +66,7 @@ EVE_FORCEINLINE auto Constant(eve::as<T> const& = {})
                     "[eve::constant] floating_point case - BitsPattern has not the correct size");
       return T {};
     }
-    else { return static_cast<T>(bit_cast(BitsPattern, as<t_t> {})); }
+    else return static_cast<T>(bit_cast(BitsPattern, as<t_t> {}));
   }
 }
 }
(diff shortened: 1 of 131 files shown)


include/eve/concept/behave_as.hpp
```diff
@@ -0,0 +1,26 @@
...
+
+  template<typename T>
+  concept behave_as_plain = plain_scalar_value<translate_underlying_type_t<T>>;
+
+  template<typename T>
+  concept behave_as_integral = integral_scalar_value<translate_underlying_type_t<T>>;
+
+  template<typename T>
+  concept behave_as_floating = floating_scalar_value<translate_underlying_type_t<T>>;
+}
```
DenisYaroshevskiy, line 26: two questions:

1. Why doesn't `constant` callable framework do the translaction
2. I don't like the `behave` thingy. will write in the chat.

jfalcou: I have simplified the impelmetnationa fetr our discussion as I fixed the external type in the external library.

include/eve/module/core/constant/constant.hpp
```diff
@@ -53,13 +52,12 @@ namespace eve
 //! @}
 //================================================================================================
 
-template<typename T, auto BitsPattern>
-requires(plain_scalar_value<element_type_t<T>>)
+template<value T, auto BitsPattern>
 EVE_FORCEINLINE auto Constant(eve::as<T> const& = {})
```
DenisYaroshevskiy, line 56: Why capital letter Comstant

include/eve/module/core/constant/constant.hpp
```diff
@@ -53,13 +52,12 @@ namespace eve
...
 {
-  using t_t = element_type_t<T>;
+  using t_t = translate_element_type_t<T>;
 
-  if constexpr( std::is_integral_v<t_t> ) { return static_cast<T>(BitsPattern); }
+  if constexpr( std::integral<t_t> ) { return static_cast<T>(BitsPattern); }
```
DenisYaroshevskiy, line 60: I'm ok with it but I'm not sure that's better.
Concepts sometimes are slower to compile std::same_as should be slower than is_same_as_v.

Probably unimportant though

include/eve/traits/as_floating_point.hpp
```diff
@@ -9,11 +9,16 @@
 
 #include <eve/detail/meta.hpp>
 #include <eve/detail/wide_forward.hpp>
+#include <eve/traits/translation.hpp>
 
 namespace eve
 {
   template<typename T>
-  struct as_floating_point : detail::make_floating_point<(sizeof(T) <= 4) ? 4: sizeof(T)>
+  inline constexpr auto under_size = sizeof(translate_element_type_t<T>);
```
DenisYaroshevskiy, line 17: under size sounds like an action. size of trans;ated is always the same, you don't need this

include/eve/module/core/constant/constant.hpp
jfalcou: it's a old name from back there was a constant something. We never changed it. 
Do you want me to make it non capitalized and change evrywhere we may use it ?

include/eve/module/core/constant/constant.hpp
jfalcou: I changed mostly for consistency

include/eve/traits/as_floating_point.hpp
jfalcou: That's a remnant of before I forgot to remove. Will cleanup

include/eve/module/core/constant/constant.hpp
jfalcou: OK, they are only used in horner so, very few places

include/eve/traits/as_floating_point.hpp
jfalcou: Fixed

DenisYaroshevskiy (approved):

PR #1920, opened 2024-08-17
[riscv] Add support for RISC-V vector extension in arch

ita-sc:
First part of initial support for RISC-V vector extension.

Full pack of patches for support could be found in https://github.com/jfalcou/eve/pull/1716

With this patch `unit.arch` tests are fully passed.

PR #1929, opened 2024-08-24
simd_cast

DenisYaroshevskiy:
casting between simd_values that's necessary for rvv simd support.


DenisYaroshevskiy:
@ita-sc

Does this look like it'll work for you?
What about when casting up to a bigger register size - is zeroing up the top part ok?

ita-sc:
> Does this look like it'll work for you?

As an extension for `bit_cast`, this new cast looks great, for RISC-V I suppose I can replace by this cast every place in RISC-V parts, where I need to have bigger/smaller wide output.

> What about when casting up to a bigger register size - is zeroing up the top part ok?

For using in usual code zeroing up the top part looks good, but for RISC-V in some places (actually, when I combine `wide`) I would prefer to be able to say that it really doesn't matter what will be there (as later I will shift value to fill the leading part of `wide`). Can this be achieved by special option maybe? As if I must have zeros there, I will need to add additional instructions here.

Also I have a question about https://github.com/jfalcou/eve/pull/1920 in context of this new cast:
* suppose in EVE common code we use `bit_cast` with `wide<int, 1>` to `wide<char, 4>`. In terms of RISC-V `storage_type` for these types should be different. Am I right to think that if I find these places, I should replace there `bit_cast` to `simd_cast`?

DenisYaroshevskiy:
>  I would prefer to be able to say that it really doesn't matter what will be there

That's mainly what I am asking. If it's cheaper to have garbage there, this interface should return garbage. I will change to that 
Interesting idea about an option, I will think about it.

> I should replace there bit_cast to simd_cast

Yeah, that's the intention. I will do some replacement in this pr too.

DenisYaroshevskiy:
@ita-sc 

>  I will change to that
Done. I didn't go with the option. Let's see if we need that at some point.

> I will do some replacement in this pr too.
Don't have the energy. Please use either `simd_cast` or `call_simd_cast` - when necessary, either one is fine, but call_simd_cast doesn't need new includes.

include/eve/module/core/regular/impl/simd/x86/simd_cast.hpp
```diff
@@ -0,0 +1,25 @@
...
+namespace eve::detail
+{
+template<callable_options O, logical_simd_value T, logical_simd_value U>
+EVE_FORCEINLINE constexpr U
+simd_cast_(EVE_REQUIRES(avx512_), O const&, T x, eve::as<U>)
+requires(!has_aggregated_abi_v<T> && !has_aggregated_abi_v<U>)
+{
+  // The minimum simd logical for chars is mask16.
+  // However mask8 exists.
+  // This causes problems in the generic logic.
```
jfalcou, line 19: care to elaborate on why it's an issue ?

include/eve/module/core/regular/impl/simd_cast.hpp
```diff
@@ -0,0 +1,38 @@
...
+namespace eve::detail
+{
+template<callable_options O, typename T, typename Target>
+EVE_FORCEINLINE constexpr Target
+simd_cast_(EVE_REQUIRES(cpu_), O const&, T x, as<Target> tgt) noexcept
+{
+  if constexpr( product_simd_value<T> )
+  {
+    auto make_as       = []<typename M>(M) { return as<M> {}; };
+    auto lifted_target = kumi::result::map_t<decltype(make_as), Target> {};
```
jfalcou, line 22: sidenote: should kumi provides a lift() function for this kind of scenario ?

include/eve/module/core/regular/simd_cast.hpp
```diff
@@ -0,0 +1,133 @@
...
+//!    * if !is_wide_logical: any logical_simd_value can cast to any other logical_simd_value,
+//!      but semantics changes, based on the representation of the logical.
+//!    * on rvv platform allows for an efficient cast between logical and plain_simd_value.
+//!
+//!   @groupheader{Callable Signatures}
+//!   @code
+//!   namespace eve
+//!   {
+//!      template <eve::simd_value T, eve::simd_value Target>
+//!        requires T can be cast to Target
```
jfalcou, line 49: traditional std parlance in those cases is `required(/* see documentation*/)`

include/eve/module/core/regular/impl/simd_cast.hpp
DenisYaroshevskiy: I don't like that I had to create a lambda.  But I dunno - this is some mpl style shenanigans.

jfalcou: OK, will make a KUMI issue thoguh, it looks like a valid use case

jfalcou (approved):

PR #1932, opened 2024-08-27
Move self_add into the add callable

SadiinsoSnowfall:
Special note for `common_value`: it no longer relies on `operator+` as it previously did, see commit message.

### DIFF
--- a/include/eve/traits/common_value.hpp
+++ b/include/eve/traits/common_value.hpp
@@ -8,40 +8,32 @@
 #pragma once
 
 #include <eve/concept/value.hpp>
+#include <eve/concept/simd.hpp>
 #include <eve/traits/as_wide.hpp>
 #include <type_traits>
 
 namespace eve::detail
 {
   template<typename T>
-  struct fth { using type = T; };
+  struct find_type_impl {
+    using type = T;
 
-  template<scalar_value S0, scalar_value S1>
-  consteval fth<decltype((std::declval<S0>() + std::declval<S1>()))> operator%(fth<S0>, fth<S1>) noexcept
-  {
-    return {};
-  }
-
-  template<typename T, typename N, scalar_value S>
-  consteval auto operator%(fth<wide<T, N>>, fth<S>) noexcept
-  {
-    return fth<wide<T, N>>{};
-  }
-
-  template<typename T, typename N, scalar_value S>
-  consteval auto operator%(fth<S>, fth<wide<T, N>>) noexcept
-  {
-    return fth<wide<T, N>>{};
-  }
-
-  template<typename T, typename N>
-  consteval auto operator%(fth<wide<T, N>>, fth<wide<T, N>>) noexcept
-  {
-    return fth<wide<T, N>>{};
-  }
+    template<value U>
+    friend consteval find_type_impl<std::conditional_t<
+        scalar_value<T> && scalar_value<U>,
+          decltype(std::declval<T>() + std::declval<U>()),
+          std::conditional_t<
+              (scalar_value<T> && arithmetic_simd_value<U>) ||
+              (arithmetic_simd_value<T> && scalar_value<U>) ||
+              (arithmetic_simd_value<T> && std::same_as<T, U>),
+                std::conditional_t<arithmetic_simd_value<T>, T, U>,
+                void
+          >
+    >> operator%(find_type_impl, find_type_impl<U>) noexcept { return {}; }
+  };
 
   template<typename... Ts>
-  using find_type = typename decltype((fth<Ts>{} % ...))::type;
+  using find_type = typename decltype((find_type_impl<Ts>{} % ...))::type;
 
   template<typename... Ts>
   requires(!scalar_value<Ts> || ... )
--- a/include/eve/traits/common_value.hpp
+++ b/include/eve/traits/common_value.hpp
@@ -14,27 +14,29 @@
 namespace eve::detail
 {
   template<typename T>
-  struct fth {
-    using type = T;
-  };
+  struct fth { using type = T; };
 
   template<scalar_value S0, scalar_value S1>
-  consteval auto operator%(fth<S0>, fth<S1>) noexcept {
-    return fth<decltype((std::declval<S0>() + std::declval<S1>()))>{};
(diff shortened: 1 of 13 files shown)


DenisYaroshevskiy:
You are going to make the risk pull request to do a lot of work. It's probably OK, if this is the most important ones you can be doing. If you can not touch core for a bit, without sacrificing productivity, I'd suggest to do that.

include/eve/traits/common_value.hpp
```diff
@@ -13,15 +13,42 @@
...
+
+  template<typename T, typename N>
+  consteval auto operator%(fth<wide<T, N>>, fth<wide<T, N>>) noexcept
+  {
+    return fth<wide<T, N>>{};
+  }
+
   template<typename... Ts>
-  using find_type = decltype(( std::declval<Ts>() + ... ));
+  using find_type = typename decltype((fth<Ts>{} % ...))::type;
```
DenisYaroshevskiy, line 44: 1. The name `fth` doesn't mean anyhthing. And it get's exposed everywhere.
2. My find_type has the same problem, we should name it better.  
3. don't pollute the namespace, operators should be done through a `hidden friends idiom`
4. Don't do the overloading, just do if constexpr chain.

```
template <typename T> 
struct find_common_value_combinator {
  using type = T;
  
 template <typename U> 
  friend auto operator%(
    find_common_value_combinator x,
    find_common_value_combinator<U> y) {
    if constexpr (...) {
   }
  }
};
```

DenisYaroshevskiy: I don't like it. Because we don't have local names in headers - it's better to have the word `common` in there somewhere.

DenisYaroshevskiy: find_common_type? common_value_find? I'm not sure why we called a type value, but that's what it is already.

include/eve/traits/common_value.hpp
```diff
@@ -8,20 +8,39 @@
...
+    friend consteval find_type_impl<std::conditional_t<
+        scalar_value<T> && scalar_value<U>,
+          decltype(std::declval<T>() + std::declval<U>()),
+          std::conditional_t<
+              (scalar_value<T> && arithmetic_simd_value<U>) ||
+              (arithmetic_simd_value<T> && scalar_value<U>) ||
+              (arithmetic_simd_value<T> && std::same_as<T, U>),
+                std::conditional_t<arithmetic_simd_value<T>, T, U>,
+                void
+          >
```
DenisYaroshevskiy, line 31: Why would you do this.

Use auto and if constexpr, this is impossible to read.

jfalcou: We need sfinae there or we ends up with hard error in some places with ugly messages. That is why the old.code was decltype based.

jfalcou: And why thebprevious iteration was overload based.

include/eve/traits/common_value.hpp
```diff
@@ -8,20 +8,39 @@
...
-  requires(!plain_scalar_value<Ts> || ... )
+  requires(!scalar_value<Ts> || ... )
   auto find_common_value() -> find_type<Ts... >;
 
   template<typename T0, typename... Ts>
-  requires(   (plain_scalar_value<T0> && ... && plain_scalar_value<Ts>)
+  requires(   (scalar_value<T0> && ... && scalar_value<Ts>)
           &&  (std::same_as<T0,Ts> && ...)
           )
   T0 find_common_value();
```
DenisYaroshevskiy, line 46: This looks very sus btw. Why would this be the case? 
I remember the point of this was - that standard common_type did promotions. We didn't want promotions in simd. Why is first type preferred.

include/eve/traits/common_value.hpp
DenisYaroshevskiy: This passes unittests.
I really don't like that it's not symetrical.

```
  template<typename T>
  struct find_common_value_wide_reducer;

  template<>
  struct find_common_value_wide_reducer<void> {};

  template<typename T>
  struct find_common_value_wide_reducer {
    using type = T;

    template <typename U>
    friend auto operator%(
      find_common_value_wide_reducer x,
      find_common_value_wide_reducer<U> y
    ) {
      if constexpr (scalar_value<T> && scalar_value<U>) {
        // Or use decltype if you want
        return find_common_value_wide_reducer<decltype(T{} + U{})>{};
      } else if constexpr ( scalar_value<T> && arithmetic_simd_value<U> ) {
        return y;
      } else if constexpr ( arithmetic_simd_value<T> && (scalar_value<U> || arithmetic_simd_value<U>) ) {
        return x;
      } else {
        return find_common_value_wide_reducer<void>{};
      }
    }
  };

  template<typename... Ts>
  using find_common_value_wide = typename decltype((find_common_value_wide_reducer<Ts>{} % ...))::type;

  template<typename... Ts>
  requires(!scalar_value<Ts> || ... )
  auto find_common_value() -> find_common_value_wide<Ts... >;

```

include/eve/traits/common_value.hpp
jfalcou: We check T0 and all Ts are the same anyway. So we return the first one cause it is the same one than all the other.

include/eve/traits/common_value.hpp
jfalcou: This looks ok for me

include/eve/traits/common_value.hpp
DenisYaroshevskiy: Ah. Why is this a thing though? What does it do?

include/eve/traits/common_value.hpp
DenisYaroshevskiy: Since you answered about scalar, answer here - why is left prefred to right? Why between `wide<int8_t>` and `wide<int16_t>` we chose the first one?

DenisYaroshevskiy (approved):

include/eve/traits/common_value.hpp
DenisYaroshevskiy: I see - now it makes sense, sorry - was confused.

DenisYaroshevskiy (approved):

PR #1936, opened 2024-08-28
Move self_div to the div callable

SadiinsoSnowfall:
TODO: check if it is possible to optimize the saturation/rounding in the backends

### DIFF
--- a/include/eve/detail/function/compounds.hpp
+++ b/include/eve/detail/function/compounds.hpp
@@ -9,25 +9,20 @@
 
 #include <eve/arch.hpp>
 #include <eve/detail/function/simd/common/bit_compounds.hpp>
-#include <eve/detail/function/simd/common/arithmetic_compounds.hpp>
 #include <eve/detail/function/operators.hpp>
 
 #if defined(EVE_INCLUDE_X86_HEADER)
 #  include <eve/detail/function/simd/x86/bit_compounds.hpp>
-#  include <eve/detail/function/simd/x86/arithmetic_compounds.hpp>
 #endif
 
 #if defined(EVE_INCLUDE_POWERPC_HEADER)
 #  include <eve/detail/function/simd/ppc/bit_compounds.hpp>
-#  include <eve/detail/function/simd/ppc/arithmetic_compounds.hpp>
 #endif
 
 #if defined(EVE_INCLUDE_ARM_HEADER)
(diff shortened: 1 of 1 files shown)


include/eve/module/core/regular/impl/simd/arm/sve/div.hpp
```diff
@@ -12,40 +12,66 @@
...
+    }
+    else
+    {
+      if constexpr (sizeof(T) >= 4)
+      {
+        return svdiv_x(sve_true<T>(), a, b);
+      }
+      else 
+      {
+        apply<N::value>([&](auto... I) { (a.set(I, a.get(I) / b.get(I)), ...); });
```
DenisYaroshevskiy, line 36: this should be just in generic. Let's forward cpu.

DenisYaroshevskiy (commented):
If you are really interested in saturated arithmetics - here is some article I stamped upon: https://locklessinc.com/articles/sat_arithmetic/

include/eve/module/core/regular/impl/simd/arm/sve/div.hpp
DenisYaroshevskiy: Then there is a bug that needs to be fixed.  Copy paste so that the bug doesn't trigger, is bad. Why is there an infinite.

include/eve/arch/cpu/wide.hpp
```diff
@@ -647,35 +648,35 @@ namespace eve
...
     {
-      auto that = v;
-      return that /= w;
+      return div(a, b);
     }
 
     //! @brief Performs the division between a scalar and all lanes of a eve::wide
     //! See also: eve::div
-    friend EVE_FORCEINLINE auto operator/(plain_scalar_value auto s, wide const& v) noexcept
+    friend EVE_FORCEINLINE wide operator/(plain_scalar_value auto s, wide const& w) noexcept
```
DenisYaroshevskiy, line 668: fyi: wides can always pass by value, it's probably preferred.

include/eve/module/core/regular/impl/simd/arm/sve/div.hpp
DenisYaroshevskiy: still an issue

DenisYaroshevskiy (approved):

DenisYaroshevskiy (approved):

PR #1939, opened 2024-08-31
Remove the bit_compatible family of concepts

SadiinsoSnowfall:
In preparation of moving all `bit_compound` function to proper callables, we found out that the current callable versions of these functions accepted a more relaxed set of parameters, by using `bit_value_t` as a guard instead of `bit_compatible_value`.

The goal of this PR is to see if it would be possible to completly scrap the `bit_compatible_*` family of concepts, as they might not be needed when moving all `bit_compound` functions into their callables counterparts.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -43,6 +43,7 @@ endif()
 ## Test and Benchmarks target
 ## =================================================================================================
 if( EVE_BUILD_TEST )
+  set(CTEST_COVERAGE_EXTRA_FLAGS "--long-file-names --hash-filenames")
   include(CTest)
   include(config/dependencies)
   include(config/compiler)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -43,7 +43,6 @@ endif()
 ## Test and Benchmarks target
 ## =================================================================================================
 if( EVE_BUILD_TEST )
-  set(CTEST_COVERAGE_EXTRA_FLAGS "--long-file-names --hash-filenames")
   include(CTest)
   include(config/dependencies)
   include(config/compiler)
--- a/include/eve/arch/cpu/wide.hpp
+++ b/include/eve/arch/cpu/wide.hpp
@@ -401,9 +401,9 @@ namespace eve
     //! @brief Performs a compound bitwise and on all the wide lanes and assign the result to the current
     //! one
     template<value V>
-    friend EVE_FORCEINLINE bit_value_t<wide, V>& operator&=(wide& w, V o) noexcept
+    friend EVE_FORCEINLINE wide& operator&=(wide& w, V o) noexcept
 #if !defined(EVE_DOXYGEN_INVOKED)
-        requires (!kumi::product_type<Type>)
+        requires (!kumi::product_type<Type> && supports_bitwise_call<wide, V>)
 #endif
     {
       return detail::self_bitand(w, o);
@@ -412,9 +412,9 @@ namespace eve
     //! @brief Performs a bitwise and between all lanes of two wide instances.
     //! Do not participate to overload resolution if both wide doesnot have the same `sizeof`
     template<scalar_value U, typename M>
-    friend EVE_FORCEINLINE bit_value_t<wide, wide<U, M>> operator&(wide const& v, wide<U, M> const& w) noexcept
+    friend EVE_FORCEINLINE wide operator&(wide const& v, wide<U, M> const& w) noexcept
 #if !defined(EVE_DOXYGEN_INVOKED)
-        requires (!kumi::product_type<Type>)
+        requires (!kumi::product_type<Type> && supports_bitwise_call<wide, wide<U, M>>)
 #endif
     {
       auto that = v;
@@ -424,9 +424,9 @@ namespace eve
     //! @brief Performs a bitwise and between all lanes of a eve::wide and a scalar
     //! Do not participate to overload resolution if `sizeof(Type) != sizeof(S)`
     template<scalar_value S>
-    friend EVE_FORCEINLINE bit_value_t<wide, S> operator&(wide const& v, S w) noexcept
+    friend EVE_FORCEINLINE wide operator&(wide const& v, S w) noexcept
 #if !defined(EVE_DOXYGEN_INVOKED)
-        requires(!kumi::product_type<Type>)
+        requires(!kumi::product_type<Type> && supports_bitwise_call<wide, S>)
 #endif
     {
       auto that = v;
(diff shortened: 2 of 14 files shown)


include/eve/arch/cpu/wide.hpp
```diff
@@ -401,9 +401,9 @@ namespace eve
     //! @brief Performs a compound bitwise and on all the wide lanes and assign the result to the current
     //! one
     template<value V>
-    friend EVE_FORCEINLINE auto operator&=(wide& w, V o) noexcept
+    friend EVE_FORCEINLINE bit_value_t<wide, V> operator&=(wide& w, V o) noexcept
```
DenisYaroshevskiy, line 404: This feels weird.

1. &= should return reference. to *this. so wide&.
2. I'm not sure what's happening - are we allowing to bit and anything with anything?

what does `eve::wide<char> & double` do?

include/eve/module/core/regular/impl/simd/x86/popcount.hpp
```diff
@@ -22,123 +22,68 @@
...
+    const i8_t pattern_4bit(0x33);
+    const i8_t pattern_16bit(0x0f);
+    auto xx = bit_cast(in, as<i8_t>()); // put
+
+    xx -= bit_shr(xx, 1) & pattern_2bit; // put
+    xx  = (xx & pattern_4bit) + (bit_shr(xx, 2) & pattern_4bit);
+    xx  = (xx + bit_shr(xx, 4)) & pattern_16bit; // put count of each 8 bits into those 8 bits
+
+    return bit_cast(xx,as(in));
+  };
```
DenisYaroshevskiy, line 40: this looks like cpu implementation, no? should this maybe be in cpu.

test/unit/api/regular/arithmetic.cpp
```diff
@@ -126,4 +139,16 @@ TTS_CASE_WITH( "Check behavior of arithmetic operators on wide and scalar"
...
+  TTS_EQUAL(cpy, T([&](auto i, auto) { return a0.get(i) + s; }));
+
+  cpy = a0; (cpy -= s);
+  TTS_EQUAL(cpy, T([&](auto i, auto) { return a0.get(i) - s; }));
+
+  cpy = a0; (cpy *= s);
+  TTS_EQUAL(cpy, T([&](auto i, auto) { return a0.get(i) * s; }));
+
+  cpy = a0; (cpy /= s);
+  TTS_ULP_EQUAL(cpy, T([&](auto i, auto) { return a0.get(i) / s; }), 1);
```
DenisYaroshevskiy, line 153: run clang format on the file. I don't mind your two steatements on one line but it will - it's not worth it.

include/eve/arch/cpu/wide.hpp
```diff
@@ -401,9 +401,9 @@ namespace eve
     //! @brief Performs a compound bitwise and on all the wide lanes and assign the result to the current
     //! one
     template<value V>
-    friend EVE_FORCEINLINE auto operator&=(wide& w, V o) noexcept
+    friend EVE_FORCEINLINE bit_value_t<wide, V>& operator&=(wide& w, V o) noexcept
```
DenisYaroshevskiy, line 404: I have a problem with this style here a bit.

You see - it was confusing for me, what is `bit_value_t`, so what can you expect from a casual reader of this file (or the error message).

This is `wide`, right? You can only do bit operations on the same type `wide` and scalar of the same type?

I believe the result type then should be a wide and requires clause explicitly specify when the operator is enabled.

DenisYaroshevskiy: I think it doesn't mean anything. You want a concept with a meaning.

CMakeLists.txt
```diff
@@ -43,6 +43,7 @@ endif()
 ## Test and Benchmarks target
 ## =================================================================================================
 if( EVE_BUILD_TEST )
+  set(CTEST_COVERAGE_EXTRA_FLAGS "--long-file-names --hash-filenames")
```
jfalcou, line 46: noise ?

jfalcou (changes requested):

include/eve/arch/cpu/wide.hpp
```diff
@@ -401,9 +401,9 @@ namespace eve
     //! @brief Performs a compound bitwise and on all the wide lanes and assign the result to the current
     //! one
     template<value V>
-    friend EVE_FORCEINLINE auto operator&=(wide& w, V o) noexcept
+    friend EVE_FORCEINLINE wide& operator&=(wide& w, V o) noexcept
 #if !defined(EVE_DOXYGEN_INVOKED)
-        requires(!kumi::product_type<Type> && bit_compatible_values<wide, V>)
+        requires (!kumi::product_type<Type> && supports_bitwise_call<wide, V>)
```
DenisYaroshevskiy, line 406: the word "supports bitwise call" is not really sayinmg much too. It would be fine if that had a doc but it doesn't.

What are the actual requirements? same size?
Usually we cast scalars to Type - that doesn't work here?

Can we bit and doubles? C++ standar can't. https://godbolt.org/z/hhfvYM7Kh

I appreciate that you didn;t do this originally, but let's fix it up proper, since you touched it

DenisYaroshevskiy: This is kinda not awesome.

1. I'm ify about bit operations on floats, since the standard doesn't do it.
2. two scalars - the same size in bits - doesn't work, this operator will never be called for scalars.
3. "element of the wide and the scalar have the same size" - this is inconistent with how other operators work, I believe we always cast. 

@jfalcou 

How about

```
wide& operator&=(Type x); // sclaras are implicitly cast, only integrals

wide& operator&=(wide<U, N>) // both are integrals. same element size. 
```

DenisYaroshevskiy: yeah - the problem is there.

ok then - how about `same_element_size_value` concept? and not product?

jfalcou: Operator& on wide won't let us do `float & float` anyway.

The semantic of bit_and(x,y) is to treat y as a bitmask and return x & bits(y) in the same type of x. 
Has been that since ages and it works.

The semantic of bit operations taking whatever as long as the # of bits fits is quite the source of a lot of optimisation in EVE. This semantic is documented properly for bit_* functions and changing all the place where we use it means 100s of bit_cast noisy changes. So I am not too keen on that.

jtlap: I deeply agree with joel point of view. I am fiercely (but kindly) against changes to that semantic.

jfalcou (approved):

PR #1942, opened 2024-09-01
Fpclass

jtlap:
replace the previous pr which presented weird fails


test/doc/core/is_bit_equal.hpp
```diff
@@ -0,0 +1,32 @@
+// revision 0
```
jfalcou, line 1: useless file, remove

test/doc/math/pow.cpp
```diff
@@ -1,18 +1,19 @@
...
 #include <iostream>
+#include <cmath>
 
 int main()
 {
-  eve::wide pf = {2.0f, 3.0f, -4.0f, 2.0f, 2.0f,
+  eve::wide qf = {0.0f, 3.0f, -4.0f, 2.0f, 2.0f,
                   eve::inf(eve::as<float>()), eve::minf(eve::as<float>()), eve::nan(eve::as<float>())};
-  eve::wide qf = {4.0f, 1.0f, -1.0f,  0.5f,  0.0f, 2.0f, -3.0f,  2.5f};
+  eve::wide pf = {4.0f, 1.0f, -1.0f,  0.5f,  0.0f, 0.0f, -0.0f,  2.5f};
```
jfalcou, line 10: why changing those values ?

jfalcou (changes requested):

test/doc/core/is_bit_equal.hpp
jtlap: done

test/doc/math/pow.cpp
jtlap: I had an error due to is_minf and this made it visible

jfalcou (approved):

PR #1944, opened 2024-09-03
Partial CI for RISC-V

jfalcou:
Setup a basic RISC-V CI instance.


DenisYaroshevskiy (approved):

test/unit/meta/traits/has_abi.cpp
```diff
@@ -33,13 +36,16 @@ TTS_CASE( "Check for detection of aggregated ABI")
...
   constexpr auto native = eve::wide<float>::size();
 
-#if defined(EVE_NO_SIMD)
-  TTS_EXPECT_NOT(( eve::has_aggregated_abi_v<eve::wide<float, eve::fixed<2*native>>>              ));
-  TTS_EXPECT_NOT(( eve::has_aggregated_abi_v<eve::logical<eve::wide<float, eve::fixed<2*native>>>>));
-#else
-  TTS_EXPECT(( eve::has_aggregated_abi_v<eve::wide<float, eve::fixed<2*native>>>              ));
-  TTS_EXPECT(( eve::has_aggregated_abi_v<eve::logical<eve::wide<float, eve::fixed<2*native>>>>));
-#endif
+  if constexpr(eve::current_api == eve::undefined_simd)
```
DenisYaroshevskiy, line 39: What changed?

jfalcou: the test over EVE_NO_SIMD was used to discriminate when we were running this test on actual arch or on the NO-SIMD case. But there is one case which is running test on an unsupproted arch without NO_SIMD enabled. 

Checking that the current_api is unsupprtoed_simd covers both case and limit the failure when someone will, later, run EVE on a yet unsupported arch.

PR #1945, opened 2024-09-03
fix SVE2 support

SadiinsoSnowfall:
Currently, both GCC and Clang assume that ARMv9 means SVE2 (and SVE).
This is not necessarily true (for example, Apple M4), but this also means that everything targeting armv9 will assume SVE2 (for now).

This patch makes sure that SVE2 is detected properly by EVE, and falls back to SVE in the SIMD API hierarchy.

### DIFF
--- a/include/eve/arch/arm/predef.hpp
+++ b/include/eve/arch/arm/predef.hpp
@@ -8,10 +8,7 @@
 #pragma once
 
 #include <eve/detail/spy.hpp>
-
-#if defined(SPY_SIMD_IS_ARM_FIXED_SVE) || defined(SPY_SIMD_IS_ARM_FIXED_SVE2)
-#  define EVE_SIMD_IS_ANY_FIXED_SVE
-#endif
+#include <eve/arch/cpu/predef.hpp>
 
 // We successfully detected some native SIMD
 #if defined(EVE_SIMD_IS_ANY_FIXED_SVE) && !defined(EVE_NO_SIMD)
--- a/include/eve/arch/cpu/predef.hpp
+++ b/include/eve/arch/cpu/predef.hpp
@@ -1,0 +1,14 @@
+//==================================================================================================
+/*
+  EVE - Expressive Vector Engine
+  Copyright : EVE Project Contributors
+  SPDX-License-Identifier: BSL-1.0
+*/
+//==================================================================================================
+#pragma once
+
+#include <eve/detail/spy.hpp>
+
+#if defined(SPY_SIMD_IS_ARM_FIXED_SVE) || defined(SPY_SIMD_IS_ARM_FIXED_SVE2)
+#  define EVE_SIMD_IS_ANY_FIXED_SVE
+#endif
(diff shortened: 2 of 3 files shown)


DenisYaroshevskiy:
we don't ci sve2, right? only sve1?
sve2 is shipping on like one phone.

test/unit/meta/concepts/native_simd_for_abi.cpp
```diff
@@ -66,7 +66,7 @@ TTS_CASE_TPL( "Check that wide<T,N,ppc*> does not satisfy native_simd_for_abi wi
   TTS_EXPECT_NOT((eve::native_simd_for_abi<logical<wide<T>> , eve::arm_64_ , eve::arm_128_  > ));
   TTS_EXPECT_NOT((eve::native_simd_for_abi<logical<wide<T>> , eve::x86_128_ , eve::x86_256_, eve::x86_512_ > ));
 };
-#elif defined(SPY_SIMD_IS_ARM_FIXED_SVE)
+#elif defined(SPY_SIMD_IS_ARM_FIXED_SVE) || defined(SPY_SIMD_IS_ARM_FIXED_SVE2)
```
DenisYaroshevskiy, line 69: should we have one macro? @jfalcou

jfalcou: Some global one ? EVE could add one if you think it improves ergonomics.
Some `EVE_SIMD_IS_ANY_FIXED_SVE` ?

DenisYaroshevskiy: Maybe

DenisYaroshevskiy (commented):
lgtm. @jfalcou - it's more your department, plz take a look

include/eve/arch/arm/sve/tags.hpp
```diff
@@ -45,14 +45,24 @@ namespace eve
...
-  inline constexpr sve128_  sve128    = {};
-  inline constexpr sve256_  sve256    = {};
-  inline constexpr sve512_  sve512    = {};
+  inline constexpr sve_     sve        = {};
+  inline constexpr sve_vls_ sve_vls    = {};
+  inline constexpr sve128_  sve128     = {};
+  inline constexpr sve256_  sve256     = {};
+  inline constexpr sve512_  sve512     = {};
+
+  inline constexpr sve2_      sve2     = {};
```
jfalcou, line 62: There is one thing to consider here and I am not sure it is a big deal.
In this PR, we make sve2_xxx to depends on sve_xxx, completely by passing sve2_vls.

Is it a real issue or not ? Or should we go to sve2_vls ?
Or we don't care for now ?

DenisYaroshevskiy: Which ordering breaks in this case?

jfalcou: > Which ordering breaks in this case?

well, now that I think of it, none interesting.
So this is maybe moot.

jfalcou (approved):

PR #1947, opened 2024-09-05
Move self_shl & self_shr to their respective callable version

SadiinsoSnowfall:
+ added tests for assigning shifts


include/eve/detail/assert_utils.hpp
```diff
@@ -14,7 +14,7 @@
 namespace eve::detail
 {
   template<typename A0, eve::value A1>
-  constexpr EVE_FORCEINLINE bool assert_good_shift(A1 const &t) noexcept
+  constexpr EVE_FORCEINLINE bool assert_good_shift(A1 t) noexcept
```
DenisYaroshevskiy, line 17: I find it weird that we allow a shift by negative number for signed value type.

gcc complains: https://godbolt.org/z/hEdT5zfed

Not for this pr though.

DenisYaroshevskiy (commented):
lgtm

include/eve/detail/assert_utils.hpp
DenisYaroshevskiy: You either go all signed shifts or none signed shifts. The "signed only for signed numbers" makes no sense.

DenisYaroshevskiy:
Iota test is correct. It shows a real problem in the system. Don't put asserts

jfalcou:
The decision we took for now is to remove assert from shifts. We can advance on that, finish the callable transition and suppport for type translation and think about how EVE shoudl sue assert in the mean time.

@DenisYaroshevskiy are you OK with that ?

PR #1950, opened 2024-09-06
[riscv] Add support for RISC-V vector extension in arch v2

ita-sc:


ita-sc:
@DenisYaroshevskiy , @jfalcou created a new review as requested. Could you please take a look?

Please do note: removed changes related to `bit_cast`(for now). Going forward, I think that enabling bit_cast may faster, however this may be an improvement for the future.

include/eve/arch/riscv/as_register.hpp
```diff
@@ -0,0 +1,299 @@
...
+#include <type_traits>
+#if defined(EVE_INCLUDE_RISCV_HEADER)
+
+namespace eve
+{
+template<typename T> struct logical;
+}
+
+#  if defined(EVE_RETURN_M)
+#    error EVE_RETURN_M already defined
```
jfalcou, line 23: In which scenario can this happens ?

include/eve/arch/riscv/as_register.hpp
```diff
@@ -0,0 +1,299 @@
...
+
+#  define EVE_RETURN_MF(REG, LMUL)                                                                 \
+    do {                                                                                           \
+      using type = REG __attribute__((riscv_rvv_vector_bits(__riscv_v_fixed_vlen / LMUL)));        \
+      return type {};                                                                              \
+    }                                                                                              \
+    while( 0 )
+
+#  define EVE_RETURN_M(REG, LMUL)                                                                  \
+    do {                                                                                           \
```
jfalcou, line 38: the macro is used in a if consetxpr context, why not putting the { } in the if constexpr and just have the macro be the using + return. The do while(0) here is a bit overkill

DenisYaroshevskiy: It was my suggetion - I liked the semicolon. If Joel doesn't like it - do the brackets

ita-sc: Done

include/eve/arch/riscv/as_register.hpp
ita-sc: This is just to be safe. I've already had an experience of debugging override of macros, so I prefer to have a guarantee that this situation is not happening here. If you wish, I can delete this, as now we do not have any possibility to have this macro defined.

ita-sc:
> Do you think we can close the older prs for now? at leats the previous version of arch one?

Closed previous version of arch, full version for now moved to draft.

ita-sc:
Hi folks - do you have any opinion on how close we are to wrap up this review? To me it looks like we addressed most concerns - I just wanted to double-check if there are major issues that were not discussed yet

jfalcou:
> Hi folks - do you have any opinion on how close we are to wrap up this review? To me it looks like we addressed most concerns - I just wanted to double-check if there are major issues that were not discussed yet

I think this is good. @DenisYaroshevskiy ?

DenisYaroshevskiy:
I think this is close but every time I work through it again, I see small but important things to be fixed. 

I appreciate that it's frustrating but I wouldn't stop smth if I didn't think it was important.

@jfalcou - can you confirm that this tests aggregates?

DenisYaroshevskiy:
One iption is: we merge more or less as is and then I fix up small thingies afterwards myself.

Does that sound reasonable?

jfalcou:
The tests pass on the RISCV CI. So I am all for merging then fixinf the details

jfalcou (approved):

DenisYaroshevskiy:
@ita-sc - basically - up to you. I can continue giving you feedback. Or I can just fix it up after we merge.

If you feel like continue contributing your work further, please split it up futrher things in small chunks (function or two)- we will do those quickly and well

ita-sc:
> @ita-sc - basically - up to you. I can continue giving you feedback. Or I can just fix it up after we merge.

As main idea for this patch is clean and discussed, I suggest to merge this now, and make a separate review for small fix-ups later.

I will start work to create one more review that will clean some of TODOs here (like, extend `simd_cast` to remove functions like `riscv_lmul_trunc`) and next MR with commits to support next part of unit tests. 

Unfortunately, splitting commit one more time for basic tests seems quite hard -  for tests `top_bits` and `new_arch_first_tests` I almost sure that I will need full commit.

PR #1961, opened 2024-09-13
RISCV bit_cast

ita-sc:
Hi

This patch improves `bit_cast` support for RISC-V `Wide` types. Now instead of using `std::bit_cast`, that could lead to `std::memcpy`, we use pseudo intrinsics, that will result in no real code.

Also I've added some tests for `bit_cast` to check that expected combination of types will be supported.

### DIFF
--- a/test/unit/module/core/bit_cast.cpp
+++ b/test/unit/module/core/bit_cast.cpp
@@ -10,18 +10,39 @@
 //==================================================================================================
 // Types tests
 //==================================================================================================
+
+// We use sizeof(Wide) to get real occupied size, so if we divide it to the sizeof(Scalar),
+// we get the number of Scalar elements that we need to occupy the same size.
+template<eve::arithmetic_simd_value Wide, typename Scalar>
+using same_size_of =
+    eve::wide<Scalar, eve::fixed<std::max<size_t>(1, sizeof(Wide) / sizeof(Scalar))>>;
+
 TTS_CASE_TPL("Check return types of bit_cast", eve::test::simd::all_types)
 <typename T>(tts::type<T>)
 {
   using ut_t = eve::as_integer_t<T, unsigned>;
   using it_t = eve::as_integer_t<T, signed>;
-  using v_t  = eve::element_type_t<T>;
-  using u8_t = eve::as_wide_t<std::int8_t, eve::fixed<eve::cardinal_v<T> * sizeof(v_t)>>;
+  using u8_t  = same_size_of<T, std::int8_t>;
+  using u16_t = same_size_of<T, std::int16_t>;
+  using u32_t = same_size_of<T, std::int32_t>;
+  using u64_t = same_size_of<T, std::int64_t>;
+  using f32_t = same_size_of<T, float>;
+  using f64_t = same_size_of<T, double>;
   // regular
   TTS_EXPR_IS(eve::bit_cast(T(), eve::as<it_t>()), it_t);
   TTS_EXPR_IS(eve::bit_cast(T(), eve::as<ut_t>()), ut_t);
   TTS_EXPR_IS(eve::bit_cast(T(), eve::as<T>()), T);
   TTS_EXPR_IS(eve::bit_cast(T(), eve::as<u8_t>()), u8_t);
+  if constexpr( sizeof(T) == sizeof(u16_t) )
+    TTS_EXPR_IS(eve::bit_cast(T(), eve::as<u16_t>()), u16_t);
+  if constexpr( sizeof(T) == sizeof(u32_t) )
+    TTS_EXPR_IS(eve::bit_cast(T(), eve::as<u32_t>()), u32_t);
+  if constexpr( sizeof(T) == sizeof(f32_t) )
+    TTS_EXPR_IS(eve::bit_cast(T(), eve::as<f32_t>()), f32_t);
+  if constexpr( sizeof(T) == sizeof(u64_t) )
+    TTS_EXPR_IS(eve::bit_cast(T(), eve::as<u64_t>()), u64_t);
+  if constexpr( sizeof(T) == sizeof(f64_t) )
+    TTS_EXPR_IS(eve::bit_cast(T(), eve::as<f64_t>()), f64_t);
 };
 
 //==================================================================================================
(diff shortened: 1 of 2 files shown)


include/eve/detail/function/simd/riscv/bit_cast.hpp
```diff
@@ -0,0 +1,263 @@
...
+// unsigned -> unsigned char
+// int -> char
+// lmul can not be changed.
+// doc. line:
+// Reinterpret between different SEW under the same LMUL
+template<callable_options O, scalar_value T, typename N, scalar_value U, typename M>
+EVE_FORCEINLINE wide<U, M>
+bit_cast_(EVE_REQUIRES(rvv_), const O&, wide<T, N> x, as<wide<U, M>> const&) noexcept
+requires rvv_abi<abi_t<T, N>> && rvv_abi<abi_t<U, M>> && (sizeof(T) != sizeof(U))
+         && (have_common_category<wide<T, N>, wide<U, M>>(category::int_,
```
jfalcou, line 132: what is the motivation for this particular requires ?

jfalcou (commented):
I am not sure it can be reused there but, in SVE we decided to bit_cast inputs to the proper integer no matter what, then convert this integer to the output type. It heleped a lot having a simpler code. What do you think about this ?

DenisYaroshevskiy:
I'm not sure I want this. compiler should do this one. It guranteed does it on other arches. What's up here.

include/eve/detail/category.hpp
```diff
@@ -133,4 +133,13 @@ namespace eve::detail
...
   }
+
+  template<typename W, typename V, typename... Cat>
+  EVE_FORCEINLINE consteval bool
+  have_common_category(Cat... tst)
+  {
+    auto w_cat = categorize<W>();
+    auto v_cat = categorize<V>();
+    return ((match(w_cat, tst) && match(v_cat, tst)) || ...);
+  }
```
DenisYaroshevskiy, line 144: You have one usage. Just inline. It's not a good name.

include/eve/detail/function/simd/riscv/bit_cast.hpp
```diff
@@ -0,0 +1,263 @@
...
+#pragma once
+
+#include <eve/arch/riscv/rvv_utils.hpp>
+#include <eve/as.hpp>
+#include <eve/concept/simd.hpp>
+#include <eve/detail/category.hpp>
+#include <eve/detail/is_native.hpp>
+#include <eve/traits/as_integer.hpp>
+
+namespace eve::detail
```
DenisYaroshevskiy, line 17: You copy paste code some code from simd_cast. That's not ok.

DenisYaroshevskiy:
If you can come up with example where bit_cast didn't compile well  in optimized mode- let's do a bug report but then yeah - we can put custom code somewhere. Otherwise - let's not.

ita-sc:
> If you can come up with example where bit_cast didn't compile well in optimized mode- let's do a bug report but then yeah - we can put custom code somewhere. Otherwise - let's not.

I've checked asm output with this patch, and with standard `std::bit_cast`, and it seems they are almost the same (I'll try to minimize the rest case), so dropping for now changes in EVE core code. Patch now consist only in changes in `bit_cast` test and turning this test in RISCV CI

include/eve/detail/category.hpp
ita-sc: Removed, as it seems `std::bit_cast` works fine

include/eve/detail/function/simd/riscv/bit_cast.hpp
ita-sc: Well, original idea was to have the exact match with RISC-V intrinsics documentation. For `vreinterpret` intrinsics we have 3 main categories:
1) the one that change SEW, but not change type characteristics (so you can cast `int` to `char`, but not `char` to `float`)
2) that change type characteristics, but not SEW (casts from `float` to `int`)
3) casts from mast to m1 vector registers - they should go to `simd_cast` (well, for now I use them as `rvv_simd_cast`, so not available in normal `simd_cast`). The reason to go to `simd_cast` is because `sizeof()` for mask types defined quite bad, so it is not correspond to real layout of mask register.

Anyway, for now removed this

test/unit/module/core/bit_cast.cpp
```diff
@@ -21,7 +21,33 @@ TTS_CASE_TPL("Check return types of bit_cast", eve::test::simd::all_types)
...
-  TTS_EXPR_IS(eve::bit_cast(T(), eve::as<u8_t>()), u8_t);
+  // for casts between different sizes of scalar underlying types,
+  // we have no guarantee that underlying types for wide are the same.
+  if constexpr( sizeof(T) == sizeof(u8_t) ) TTS_EXPR_IS(eve::bit_cast(T(), eve::as<u8_t>()), u8_t);
+  if constexpr( sizeof(v_t) >= 2 )
+  {
+    using u16_t = eve::as_wide_t<std::int16_t, eve::fixed<eve::cardinal_v<T> * sizeof(v_t) / 2>>;
+    if constexpr( sizeof(T) == sizeof(u16_t) )
+      TTS_EXPR_IS(eve::bit_cast(T(), eve::as<u16_t>()), u16_t);
+  }
```
DenisYaroshevskiy, line 32: make a using I guess.

```
template <typename T, typename Scalar>
using same_size_of = eve::wide<Scalar, 
     std::max(1, eve::fixed< T::size() * sizeof(eve::element_type_t<T>) / sizeof(Scalar))>>;
```

The the bunch of if constexprs go away and you test up scaling too.

ita-sc: This will not work for RISC-V:
Suppose we have input `wide<int, 1>` and want to cast it to `wide` with `char`. For cardinal calculations we get `std::max(1, eve::fixed< T::size() * sizeof(eve::element_type_t<T>) / sizeof(Scalar))= std::max(1, 1* 4/1) = 4`.  In result we get `wide<std::int8_t, 4>`, for vlen=128 it will occupy 1/4 of physical register in intrinsics interface. However, the original `wide<int, 1>` will use 1/2 of physical register.

It seems this can be partially achieved by using `sizeof(wide<int, 1>)` instead of  `T::size() * sizeof(eve::element_type_t<T>)`, but we still have an error when we try to have `bit_cast` , for example, from `wide<std::int8_t, 4>` to any `wide` with `double`. `wide<std::int8_t, 4>` will occupy 1/4 of physical register, but any `wide` with double will use at least use 1 physical register, so `bit_cast` will be unavailable anyway.

I can remove additional type checks, and keep only `u8_t`, I think for this case we can check without `if constexpr` stuff.

What do you think?

DenisYaroshevskiy: You still keep the if constexpr check for sizes matching. But you remove some

ita-sc: Ok, got it. Seems done

DenisYaroshevskiy (approved):

PR #1963, opened 2024-09-13
 avoiding some ub ops in trunc next_kernel

jtlap:


include/eve/module/core/regular/impl/trunc.hpp
```diff
@@ -12,7 +12,10 @@
 #include <eve/module/core/regular/abs.hpp>
 #include <eve/module/core/regular/convert.hpp>
 #include <eve/module/core/regular/if_else.hpp>
-#include <eve/module/core/regular/is_less_equal.hpp>
+#include <eve/module/core/regular/is_not_less.hpp>
```
jfalcou, line 15: 3 of those are useless, care to remove ?

PR #1965, opened 2024-09-15
Ub abs minus

jtlap:
standard abs and - are ub on valmin of integral values it is no more the case for eve::abs and eve::minus

### DIFF
--- a/include/eve/module/core/regular/abs.hpp
+++ b/include/eve/module/core/regular/abs.hpp
@@ -66,8 +66,8 @@ namespace eve
 //!      integral, `abs[saturated](valmin(as<T>{}))` returns `eve:valmax(as<T>{}))`
 //!
 //!   @note The absolute value of `x` is always representable except for
-//!    The minimum value of integral signed values. In that case contrarily to standard the result is not
-//!    undefined behaviour, but just incorrect (in this case eve::abs(valmin)` returns `valmin` !).
+//!    The minimum value of integral signed values. In that case, contrarily to the standard, the result is not
+//!     undefined behaviour, but just incorrect. In this case, `eve::abs(valmin)` returns `valmin`.
 //!
 //!   @warning
 //!   `abs` is also a standard library function name and there possibly exists a C macro version which may be called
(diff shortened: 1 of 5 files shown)


include/eve/module/core/regular/abs.hpp
```diff
@@ -65,7 +65,9 @@ namespace eve
 //!   3. The saturated absolute value of `x`. More specifically, for signed
 //!      integral, `abs[saturated](valmin(as<T>{}))` returns `eve:valmax(as<T>{}))`
 //!
-//!   @note The absolute value of `x` is always representable except fo. The minimum value of integral signed values.
+//!   @note The absolute value of `x` is always representable except for
+//!    The minimum value of integral signed values. In that case contrarily to standard the result is not
```
jfalcou, line 69: Spelling:

```
//! The minimum value of integral signed values. In that case, contrarily to the standard, the result is not
//!  undefined behaviour, but just incorrect. In this case, `eve::abs(valmin)` returns `valmin`.
```

include/eve/module/core/regular/impl/abs.hpp
```diff
@@ -27,20 +27,22 @@ namespace eve::detail
...
     {
-      if      constexpr( signed_integral_scalar_value<T>  ) return  ( (v == valmin(eve::as(v)))
-                                                                    ? valmax(eve::as(v)) : eve::abs(v)
-                                                                    );
-      else if constexpr( signed_integral_simd_value<T>    ) return if_else( v == valmin(eve::as(v))
-                                                                          , valmax(eve::as(v)), eve::abs(v)
-                                                                          );
-      else                                                  return eve::abs(v);
+      if constexpr( signed_integral_scalar_value<T>  ){
+        if(v == valmin(eve::as(v))) return valmax(eve::as(v));
```
jfalcou, line 31: Mask should be good too as we saw earlier ?

include/eve/module/core/regular/abs.hpp
jtlap: done

include/eve/module/core/regular/impl/abs.hpp
jtlap: non j'ai essayé

jfalcou (approved):

PR #1966, opened 2024-09-15
Add RISC-V support for simd_cast

ita-sc:
Hi

This patches add extends support for `simd_cast` for RISC-V. Also I'm fixing `combine` and `to_logical` functions for RISC-V and turn on wide api test, to be sure that `combine` and `slice` works fine.

Cast from logical to wide still prohibited.

### DIFF
--- a/.github/workflows/unit.yml
+++ b/.github/workflows/unit.yml
@@ -147,9 +147,9 @@ jobs:
           mkdir build && cd build
           cmake .. -G Ninja -DEVE_OPTIONS="${{ matrix.cfg.opts }}" -DCMAKE_TOOLCHAIN_FILE=../cmake/toolchain/${{ matrix.cfg.comp }}.${{ matrix.cfg.arch }}.cmake
       - name: Compiling Unit Tests
-        run:  cd build && ninja unit.meta.exe unit.arch.exe  unit.core.bit_cast.exe -j 5
+        run:  cd build && ninja unit.meta.exe unit.arch.exe  unit.core.bit_cast.exe unit.api.regular.wide.exe unit.core.simd_cast.exe -j 5
       - name: Running Unit Tests
-        run:  cd build && ctest --output-on-failure -j 4  -R "^unit.meta.*.exe|^unit.arch.*.exe|unit.core.bit_cast.exe"
+        run:  cd build && ctest --output-on-failure -j 4  -R "^unit.meta.*.exe|^unit.arch.*.exe|unit.core.bit_cast.exe|unit.api.regular.wide.exe|unit.core.simd_cast.exe"
 
   ##################################################################################################
   ## Mac OS X Targets
--- a/cmake/toolchain/clang.riscv.cmake
+++ b/cmake/toolchain/clang.riscv.cmake
@@ -9,5 +9,5 @@ set(CMAKE_SYSTEM_PROCESSOR  riscv64)
 set(CMAKE_C_COMPILER    /opt/llvm/bin/clang)
 set(CMAKE_CXX_COMPILER  /opt/llvm/bin/clang++)
 
-set(CMAKE_CXX_FLAGS "-march=rv64gcv --static --target=riscv64-unknown-linux-gnu ${EVE_OPTIONS}" )
+set(CMAKE_CXX_FLAGS "-march=rv64gcv --static -fbracket-depth=512 --target=riscv64-unknown-linux-gnu ${EVE_OPTIONS}" )
 set(CMAKE_CROSSCOMPILING_CMD ${PROJECT_SOURCE_DIR}/cmake/toolchain/run_rvv128.sh )
--- a/include/eve/detail/function/simd/riscv/combine.hpp
+++ b/include/eve/detail/function/simd/riscv/combine.hpp
@@ -8,20 +8,17 @@
 #pragma once
 
 #include <eve/arch.hpp>
-#include <eve/arch/riscv/rvv_common_masks.hpp>
 #include <eve/detail/abi.hpp>
-#include <eve/module/core/regular/if_else.hpp>
 
 namespace eve::detail
 {
 template<typename T, typename N>
-EVE_FORCEINLINE auto
-combine(rvv_ const&, wide<T, N> l, wide<T, N> h) noexcept
+EVE_FORCEINLINE wide<T, typename N::combined_type>
+                combine(rvv_ const&, wide<T, N> l, wide<T, N> h) noexcept
 requires rvv_abi<abi_t<T, N>>
 {
   using that_t = wide<T, typename N::combined_type>;
 
-  constexpr size_t combined_vl = N::combined_type::value;
   if constexpr( eve::has_aggregated_abi_v<that_t> )
   {
     that_t that;
@@ -33,10 +30,11 @@ requires rvv_abi<abi_t<T, N>>
     auto           wider_l        = simd_cast(l, as<that_t> {});
     auto           wider_h        = simd_cast(h, as<that_t> {});
     constexpr auto shift_size     = N::value;
+    constexpr size_t combined_vl    = N::combined_type::value;
     that_t         wider_h_placed = __riscv_vslideup(wider_h, wider_h, shift_size, combined_vl);
-    auto           mask_all_ones  = rvv_true<T, N>();
-    auto           wider_mask     = simd_cast(mask_all_ones, as<logical<that_t>> {});
-    return if_else(wider_mask, wider_l, wider_h_placed);
+    // TODO: can be optimized when simd_cast will support conversions wide<->logical
+    logical<that_t> mask([shift_size](auto i, auto) { return i < shift_size; });
+    return __riscv_vmerge(wider_h_placed, wider_l, mask, combined_vl);
   }
 }
 
@@ -45,9 +43,7 @@ EVE_FORCEINLINE auto
 combine(rvv_ const&, logical<wide<T, N>> l, logical<wide<T, N>> h) noexcept
 requires rvv_abi<abi_t<T, N>>
 {
-  constexpr size_t combined_vl = N::combined_type::value;
-  using that_t                 = logical<wide<T, typename N::combined_type>>;
-
+  using that_t = logical<wide<T, typename N::combined_type>>;
   if constexpr( eve::has_aggregated_abi_v<that_t> )
   {
     that_t that;
--- a/include/eve/module/core/regular/impl/simd/riscv/simd_cast.hpp
+++ b/include/eve/module/core/regular/impl/simd/riscv/simd_cast.hpp
@@ -7,6 +7,9 @@
 //==================================================================================================
 #pragma once
 
+#include <eve/arch/riscv/rvv_utils.hpp>
+#include <eve/arch/wide.hpp>
+
 namespace eve::detail
 {
 // Logical-wide conversions.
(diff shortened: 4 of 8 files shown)


include/eve/arch/riscv/rvv_utils.hpp
```diff
@@ -36,5 +36,8 @@ constexpr auto rvv_logical_ratio_v = []
 
 template<plain_scalar_value T>
 using rvv_m1_wide = wide<T, fixed<__riscv_v_fixed_vlen / 8 / sizeof(T)>>;
+
+// represents a wide that occupies the same size, but formed with int8_t.
+template<arithmetic_simd_value In> using rvv_i8_wide = wide<std::int8_t, fixed<sizeof(In)>>;
```
jfalcou, line 41: Ate we sure about this ? Fixed<N> awaits a number of elements. Are tou wanting cardinal_t<In> instead ?

ita-sc: Yep. I want to have a wide with the same size (that means the same lmul) that was given as template parameter. Here I use that `sizeof(std::int8_t)==1`, so `sizeof(In)` will be the number of `std::int8_t`, that I need to have the same size, as in input.

It is possible that some of these elements will be undefined. For example:
`wide<int, 1>` will occupy 1/2 physical vector register, so for vlen = 128 `rvv_i8_wide` will be `wide<std::int8_t, 8>`, and after `bit_cast` to new wide, I will have half of wide undefined.

This is needed as if I use `wide<std::int8_t, 4>`, that has that same `sizeof(scalar_type)*cardinal`, in intrinsics interface it will occupy 1/4 of physical register, so `bit_cast` will be impossible, but I need to use as underlying type `std::int8_t` (or `std::uint8_t`) as for this types full range of lmul values are available in intrinsics interface.

jfalcou: Would you mind explaining that in detail in a comment cause, I know for sure, later on, someone will see that and think it is an error as it is rather peculiar.

test/unit/module/core/simd_cast.cpp
```diff
@@ -78,35 +77,50 @@ simd_cast_test_one_way()
...
+  }
 }
 
 template<typename T, typename U>
 void
 simd_cast_test()
 {
-  // saving compile time a bit
-  if constexpr( T::size() * 2 <= U::size() )
+  // for risc-v memory consumption is too big,
```
ita-sc, line 105: Note: still a quite a lot of memory is used.

To decrease consumption more, I can remove these types checks for RISC-V: 

https://github.com/jfalcou/eve/blob/7e784587c274c57c63505d74e8b68380c83bff60/test/unit/module/core/simd_cast.cpp#L129C9-L129C84

include/eve/arch/riscv/rvv_utils.hpp
ita-sc: Added

test/unit/module/core/simd_cast.cpp
ita-sc: It occurs we anyway have a large memory consumption. Changed to test only float types with expected and fundamental cardinals. Hope this will be enough.

ita-sc: > Changed to test only float types with expected and fundamental cardinals. Hope this will be enough.

Changed to check only expected and fundamental cardinals with casts to `unsigned char` wides. We will test lmul extend/lower functionality while casting to `wide<std::uint8_t, sizeof(T) * 2>`, and same sizes while casting to `wide<std::uint8_t, sizeof(T)>`.

test/unit/module/core/simd_cast.cpp
```diff
@@ -102,12 +103,23 @@ template<typename T, typename U>
...
 {
-  // saving compile time a bit
-  if constexpr( T::size() * 2 <= U::size() )
+  if constexpr( eve::current_api == eve::rvv )
   {
-    simd_cast_test_one_way<T, U>();
-    simd_cast_test_one_way<U, T>();
+    // for RISC-V we used decrease number of checks greatly, so
+    // default condition will not work properly
+    if constexpr( sizeof(T) * 2 <= sizeof(U) )
```
DenisYaroshevskiy, line 110: This check for compile time reduction doesn't strike me as good.

You want at least up to sizeof(T) * 2 <= sizeof(U) * 2. 

ALso - what's with "default condition will not work properly"

Finally - both if else are the same.

test/unit/module/core/simd_cast.cpp
```diff
@@ -117,30 +129,71 @@ simd_cast_logical_test()
...
-        simd_cast_test<T, eve::wide<std::uint8_t, eve::fixed<1 << Idx::value>>>();
-        simd_cast_test<T, eve::wide<std::uint16_t, eve::fixed<1 << Idx::value>>>();
-        simd_cast_test<T, eve::wide<std::uint32_t, eve::fixed<1 << Idx::value>>>();
-        simd_cast_test<T, eve::wide<std::uint64_t, eve::fixed<1 << Idx::value>>>();
-      });
+  // for RISC-V we have too big memory consumption (for vlen=128):
+  // as input we can have wide<std::int8_t, 128>,
+  // so common code will try to create wide<std::uint64_t, 128>, but on physical registers we can
+  // create wide<std::uint64_t, 16>.
+  // Other cases will have the same problem, so greatly reduce testing for RISC-V
```
DenisYaroshevskiy, line 148: We should test wide<std::uint64_t, 32>, not even 16.
We should test up to max len * 2. 

If we don't want to test those - this is not the right place to do it.

test/unit/module/core/simd_cast.cpp
```diff
@@ -117,30 +129,71 @@ simd_cast_logical_test()
...
 }
 
+template<typename Wide>
+consteval bool
+rvv_check_enable()
+{
+  using scalar_type = typename Wide::value_type;
+  return (eve::fundamental_cardinal_v<scalar_type> == eve::cardinal_v<Wide>
+          || eve::expected_cardinal_v<scalar_type> == eve::cardinal_v<Wide>);
+}
```
DenisYaroshevskiy, line 139: This check is way to restrictive. 

I don't think the suggested in this pull request approach to deal with rvv testing being problematic, is a good approach. Yeah we need to solve it somehow - but you disable like almost everything.

test/unit/module/core/simd_cast.cpp
```diff
@@ -117,30 +129,71 @@ simd_cast_logical_test()
...
-        simd_cast_logical_test<T, eve::wide<std::uint32_t, eve::fixed<1 << Idx::value>>>();
-        simd_cast_logical_test<T, eve::wide<std::uint64_t, eve::fixed<1 << Idx::value>>>();
-      });
+  // The same as for plain wide types test.
+  if constexpr( eve::current_api >= eve::rvv )
+  {
+    if constexpr( rvv_check_enable<T>() )
+    {
+      // sizeof(T) will be equal the number of std::uint8_t that will be needed to have the same
+      // size.
```
DenisYaroshevskiy, line 181: No - simd cast works for any size.

include/eve/arch/riscv/rvv_utils.hpp
```diff
@@ -36,5 +36,10 @@ constexpr auto rvv_logical_ratio_v = []
 
 template<plain_scalar_value T>
 using rvv_m1_wide = wide<T, fixed<__riscv_v_fixed_vlen / 8 / sizeof(T)>>;
+
+// Represents a wide that occupies the same size, but formed with int8_t.
+// Here we use that sizeof(std::int8_t) == 1, so sizeof(In) will represent the number
+// of std::int8_t, that are needed to have the same size.
+template<arithmetic_simd_value In> using rvv_i8_wide = wide<std::int8_t, fixed<sizeof(In)>>;
```
DenisYaroshevskiy, line 43: For chars, rvv is not special.
You can always cast to `wide<std::uint8_t, eve::fixed<sizeof(T) * N::value>`

I'm ok with having some typedef for `wide_bytes<T>` or smth, but it's not a rvv specific thing.
@jfalcou  - where would we put one.

include/eve/detail/function/simd/riscv/combine.hpp
```diff
@@ -33,10 +30,11 @@ requires rvv_abi<abi_t<T, N>>
     auto           wider_l        = simd_cast(l, as<that_t> {});
     auto           wider_h        = simd_cast(h, as<that_t> {});
     constexpr auto shift_size     = N::value;
+    constexpr size_t combined_vl    = N::combined_type::value;
     that_t         wider_h_placed = __riscv_vslideup(wider_h, wider_h, shift_size, combined_vl);
-    auto           mask_all_ones  = rvv_true<T, N>();
-    auto           wider_mask     = simd_cast(mask_all_ones, as<logical<that_t>> {});
-    return if_else(wider_mask, wider_l, wider_h_placed);
+    // TODO: can be optimized when simd_cast will support conversions wide<->logical
```
DenisYaroshevskiy, line 35: I'd suggest to do it here, since you are working on simd cast.

DenisYaroshevskiy:
Here is what I suggest with tests.

* all tests below 256 bytes.
* 16 doubles, 32 doubles, 16 uint64s, 32 uint64s

The rest is left untested.

test/unit/module/core/simd_cast.cpp
ita-sc: > ALso - what's with "default condition will not work properly"

By default  `if constexpr( T::size() * 2 <= U::size() )` is used, but it seems `wide::size()` return cardinal number:

https://github.com/jfalcou/eve/blob/ed5347d230d34a7f59a9d6ca32257603c10b57ab/include/eve/arch/cpu/wide.hpp#L377

Here I'm checking conversions between same sizes wide, and the cases when we need get size twice bigger or twice smaller, but cardinal sizes for `float` and `std::int8_t` will be too different

ita-sc:
> Here is what I suggest with tests

Yep, I will try. By the way, what memory consumption for test will be ok for EVE CI?

DenisYaroshevskiy:
No idea. We run eve ci on boxes we own, it should be not too bad as far as the limits go.

include/eve/arch/riscv/rvv_utils.hpp
jfalcou: We don't a specific place for those. So I would say in wide.hpp after the main definition ?

ita-sc: Done

include/eve/detail/function/simd/riscv/combine.hpp
ita-sc: It seems this patch will be too big.
Moreover, I'd like to try out another approach to mask initialization. My previous idea of  using separate functions in `rvv_utils` to generate `logical` seems a little bit overcomplicated.

I think this could be generalized with a lambda as a template parameter to create a correct `logical`, though I'm still figuring out how to make it pretty enough.

I'd like for this patch to focus on enabling basic support for `simd_cast`, like it is used in other architectures. This will unblock other prepared patches for RISC-V, while I'm preparing support for casts between `logical` and `wide` for RISC-V.

If you insist, I'll add support with separate functions in `rvv_utils`, like it was done before.

jfalcou:
FYI, my if_else merge was incorrect and I pushed a correction, it should fix the fails.

.github/workflows/unit.yml
```diff
@@ -147,9 +147,9 @@ jobs:
           mkdir build && cd build
           cmake .. -G Ninja -DEVE_OPTIONS="${{ matrix.cfg.opts }}" -DCMAKE_TOOLCHAIN_FILE=../cmake/toolchain/${{ matrix.cfg.comp }}.${{ matrix.cfg.arch }}.cmake
       - name: Compiling Unit Tests
-        run:  cd build && ninja unit.meta.exe unit.arch.exe  unit.core.bit_cast.exe -j 5
+        run:  cd build && ninja unit.meta.exe unit.arch.exe  unit.core.bit_cast.exe unit.api.regular.wide.exe unit.core.simd_cast.exe -j 2
```
DenisYaroshevskiy, line 150: I don't like this. Why is this happening. We agreed to reduce the test cases.

include/eve/arch/cpu/wide.hpp
```diff
@@ -997,6 +997,11 @@ namespace eve
   //! @}
   //================================================================================================
 
+  // Represents a wide that occupies the same size, but formed with int8_t.
+  // Here we use that sizeof(std::int8_t) == 1, so sizeof(In) will represent the number
+  // of std::int8_t, that are needed to have the same size.
+  template<arithmetic_simd_value In> using wide_bytes = wide<std::int8_t, fixed<sizeof(In)>>;
```
DenisYaroshevskiy, line 1003: I am not  convinced we want this typedef, especially in the public api.

include/eve/module/core/regular/impl/simd/riscv/simd_cast.hpp
```diff
@@ -79,21 +82,63 @@ EVE_FORCEINLINE logical<wide<U, M>>
...
+  else if constexpr( out_lmul == 8 ) return __riscv_vlmul_ext_i8m8(a);
+}
+
+template<callable_options O, arithmetic_simd_value In, arithmetic_simd_value Out>
+EVE_FORCEINLINE Out
+simd_cast_(EVE_REQUIRES(rvv_), const O&, In x, as<Out> const& tgt) noexcept
+requires(rvv_abi<typename In::abi_type> && rvv_abi<typename Out::abi_type>)
 {
-  if constexpr( is_aggregated_v<abi_t<T, N>> || is_aggregated_v<abi_t<U, M>> )
+  constexpr auto size_cmp = sizeof(In) <=> sizeof(Out);
```
DenisYaroshevskiy, line 128: just inline.

test/unit/module/core/simd_cast.cpp
DenisYaroshevskiy: Any logical can be simd cast to any other logical. 
We shouldn't remove the tests here. 

If there is too many cases, we should just globally test less.
I think I mentioned how I suggested to reduce the global test cases somewhere.

ita-sc: > Any logical can be simd cast to any other logical

I'm checking logical's `simd_cast` availability in `simd_cast_logical_test()`, so for `logical` and` wide` the same combinations of types are tested.

> I think I mentioned how I suggested to reduce the global test cases somewhere.

All the tests you have suggested are enabled (check `test_enable` template function - it enables tests for `wide`):

> all tests below 256 bytes

` if constexpr( sizeof(T) <= size_limit && sizeof(U) <= size_limit ) return true;`

> 16 doubles, 32 doubles, 16 uint64s, 32 uint64s

`test_enable_cardinals` enables testing for all types with expected cardinal (including 16 doubles, 16 uint64s).
Furthermore, testing is enabled for 32 doubles, 32 int64, 32 int64 (that have `cardinal` == `expected_cardinal` * 2).

include/eve/arch/cpu/wide.hpp
ita-sc: I can return it back to `rvv_utils`.
I need this typedef for RISC-V as only `int8` (or `uint8`) support full range of lmuls, so only wides with `int8` can be trunc or extend to any needed lmul.

include/eve/module/core/regular/impl/simd/riscv/simd_cast.hpp
ita-sc: You mean to use `sizeof(T) == sizeof(Out)` and `sizeof(T) < sizeof(Out)`?

.github/workflows/unit.yml
ita-sc: You mean why `-j 2`? Although the test on `simd.cast` before the change required a lot of memory (It seems locally about 60 GB), now it requires about 6 GB. Other tests need memory too. It seems `unit.api.regular.wide.exe` needs about 6-7 GB, and a test from `unit.arch.exe` also needs similar amount of memory.  I guess if we do not want to drop a lot of test cases, it's better to reduce build concurrency.

ita-sc:
Gentle ping

test/unit/module/core/simd_cast.cpp
DenisYaroshevskiy: You keep not hearing me.

simd_cast is very small test.
If we are having troubles going through all of the cases here and you have to do things like:

```
    constexpr auto size_limit = 256;
    if constexpr( sizeof(T) <= size_limit && sizeof(U) <= size_limit ) return true;
    else return test_enable_cardinals<T>();
```

then we are going to have troubles in every test.

I means we need to reduce the overall number of tests drammaticaly. 

So - we should reduce the amount of test cases requested for rvv. 
The `simd::all_types` and the other one too should change.

I suggested: 

all tests below 256 bytes.
16 doubles, 32 doubles, 16 uint64s, 32 uint64s
The rest is left untested.

ita-sc: Oh, yep, I thought to change `simd_cast` only. I will update review.
In this context, I have a question:
32 uint64s, 32 doubles - these types are not the part of `eve::test::simd::all_types`, as they are aggregated (for riscv rvv with vlen=128). You mean that I need to modify test that check aggregated, or better to modify `simd::all_types` to have `expected_cardinal` and `expected_cardinal/2`?

DenisYaroshevskiy:
In case the comment get's lost - we shouldn't filter the types for this simple test. Because if in this simple test it's a problem, it's going to be a problem for every test. More details in the big coment.

test/unit/module/core/simd_cast.cpp
DenisYaroshevskiy: Thye should include aggregated. 
Some modification here is needed for rvv:
https://github.com/jfalcou/eve/blob/031592288699acb36cb833fa23510eb028765dcf/test/test.hpp#L176

@jfalcou please  help

DenisYaroshevskiy: expected_cardinals for rvv is enourmous, I don't think testing it, except for int64 and double is something we should do. Just too slow.

ita-sc:
Updated review. Now I reduce globally types in `eve::test::simd::all_types` and similar for these rules:
* all wide types with size below or equal 512 bits (for vlen=128 they use lmul=4)
* for double, int64_t, uint64_t also check expected cardinals (for vlen=128 their size will be 8 * 128 = 1024) and combined types with cardinals == 2* expected cardinals.

Note: you requested to check all types with size bellow 256 bytes, but for vlen=128 all `wide` will be smaller: max size for "physical" register will be 128*8 / 8 = 128 bytes, so we have nothing for filtering.

DenisYaroshevskiy:
Aggregation! We have aggregation! It's important to check support for emulation of 2 registers togerther

DenisYaroshevskiy:
Eve emulates bigger than native registers if you ask for it.

include/eve/module/core/regular/impl/simd/riscv/simd_cast.hpp
DenisYaroshevskiy: yeah

include/eve/arch/cpu/wide.hpp
DenisYaroshevskiy: I'm so confused by this statement. I thought you cast to u8 and from u8 to write less code. Is that not the reason? Is there some type that is represented with the wrong type.

I' just inline this one. Especially because I'd say that bytes are unsigned.

test/tts/tts.hpp
```diff
@@ -432,6 +432,19 @@ namespace tts
...
+    template<typename... Ls> static constexpr types<Ls...> tuple_to_types(const std::tuple<Ls...>&);
+    template<typename T>
+    static constexpr std::conditional_t<Pred<T>::value, std::tuple<T>, std::tuple<>> filter_type();
+
+    template<typename... Ls> static constexpr auto filter_impl(types<Ls...>)
+    {
+      return tuple_to_types(std::tuple_cat(filter_type<Ls>()...));
+    }
+    using type = decltype(filter_impl(Type {}));
+  };
```
DenisYaroshevskiy, line 447: I appreciate you doing this, shame it wasn't there.
Should be done upstream

@jfalcou - can you implement filter? or what do you think?

DenisYaroshevskiy (commented):
inline wide of bytes and I'd prefer my other nit addressed.

Other that that I think I'm good.

I would like @jfalcou to answer what's up with tts. Maybe we merge as is, and he then overrides.

include/eve/module/core/regular/impl/simd/riscv/simd_cast.hpp
ita-sc: Done

include/eve/arch/cpu/wide.hpp
ita-sc: Done

test/tts/tts.hpp
jfalcou: I can it to main tts yes

DenisYaroshevskiy: then we let this implementation be here for now

DenisYaroshevskiy (approved):

DenisYaroshevskiy:
will merge after ci

cmake/toolchain/clang.riscv.cmake
```diff
@@ -9,5 +9,5 @@ set(CMAKE_SYSTEM_PROCESSOR  riscv64)
 set(CMAKE_C_COMPILER    /opt/llvm/bin/clang)
 set(CMAKE_CXX_COMPILER  /opt/llvm/bin/clang++)
 
-set(CMAKE_CXX_FLAGS "-march=rv64gcv --static --target=riscv64-unknown-linux-gnu ${EVE_OPTIONS}" )
+set(CMAKE_CXX_FLAGS "-march=rv64gcv --static -fbracket-depth=512 --target=riscv64-unknown-linux-gnu ${EVE_OPTIONS}" )
```
jfalcou, line 12: what does it control ?

ita-sc: This is the depth limit for nested parentheses, brackets, and braces. By default it is set to 256, but as we have a lot of inline, we reach this limit for simd_cast test.

PR #1973, opened 2024-09-20
Attempt at themed CI targets re-organisation

jfalcou:


jfalcou:
Parking this until we have more machines

PR #1991, opened 2024-10-04
Narrowing widening

jtlap:
first try on add


jtlap:
narrowing was not useful widen replace this Pr

PR #1995, opened 2024-10-09
removed reference to eve::detail::map from the unit tests

SadiinsoSnowfall:
Removed reference to `eve::detail::map` from the unit tests.
This is in preparation for a larger PR that will sunset the current `eve::detail::map` function in favour of a new non-api-compatible implementation that reduces the number of template instantiations.

The new `tts::map` replacement should be easier for the compiler to type-check and is drop-in for the most part.
This new function return type is based on its first parameter which greatly simplify the return type computation, at the cost of having to pass wide parameters first. This only breaks one test, `unit.polynomial.abel.exe`, that was fixed by reordering some parameters inside a call to `map`.

### DIFF
--- a/include/eve/traits/cardinal.hpp
+++ b/include/eve/traits/cardinal.hpp
@@ -65,8 +65,8 @@ namespace eve
   };
 
   template<typename Type>
-  using cardinal_t = typename cardinal<Type>::type;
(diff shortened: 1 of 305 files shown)


DenisYaroshevskiy:
I meant - would it be nice to replace this one as well? Or is there some functionality new map can't do

PR #2003, opened 2024-10-12
Improve the as_register machinery and cleanup arch

### LINKED ISSUE #1844: [FEATURE] Simplify as_register for better compiel time and extensibility
POC:  https://godbolt.org/z/h9cvMcdWY

```c++
#include <eve/eve.hpp>

namespace eve
{
  namespace detail 
  {
    struct ignore_t {};
  }

  template<typename T, typename N>
  consteval auto find_register_type(as<T>, N, eve::emulated_)
  {
    return std::array<T, N::value>{};
  }

  template<typename T, typename N>
  consteval auto find_register_type(as<T>, N, eve::aggregated_)
  {
    return detail::blob<T,N>{};
  }

  template<typename T, typename N>
  requires(sizeof(T)*N::value <= 64)
  consteval auto find_register_type(as<T>, N, eve::x86_abi auto)
  {
    constexpr auto width = sizeof(T)*N::value;

    if constexpr(width == 64)
    {
            if constexpr(std::same_as<T,double> ) return __m512d{};
      else  if constexpr(std::same_as<T,float > ) return __m512{};
      else  if constexpr(std::integral<T>       ) return __m512i{};
    } 
    else if constexpr(width == 32)
    {
            if constexpr(std::same_as<T,double> ) return __m256d{};
      else  if constexpr(std::same_as<T,float > ) return __m256{};
      else  if constexpr(std::integral<T>       ) return __m256i{};
    } 
    else if constexpr(width <= 16 )
    {
            if constexpr(std::same_as<T,double> )  return  __m128d{};
      else  if constexpr(std::same_as<T,float > )  return  __m128{};
      else  if constexpr(std::integral<T>       )  return  __m128i{};
    }
    else return detail::ignore_t{};
  }

  template<typename N>
  requires(sizeof(_Float16)*N::value <= 64)
  consteval auto find_register_type(as<_Float16>, N, eve::x86_abi auto)
  {
    if constexpr(current_api < eve::avx512) return __m128i{};
    else                                    return __m128h{};
  }

  template<typename T, typename N, typename ABI>
  consteval auto as_register2(as<T> t, N n, ABI abi)
  {
    static_assert( !std::same_as< decltype(find_register_type(t,n,abi))
                                , detail::ignore_t
                                >
                 , "[EVE] - Type is not usable in a SIMD register"
                 );
    return find_register_type(t,n,abi);
  }

  template<typename T, typename N, typename ABI>
  using as_register2_t = decltype(as_register2(as<T>{},N{}, ABI{}));
}

static_assert( std::same_as<eve::as_register2_t<_Float16,eve::fixed<4>,eve::x86_256_>,__m128i>);
```

SadiinsoSnowfall:
should close #1844


include/eve/arch/cpu/spec.hpp
```diff
@@ -9,11 +9,9 @@
 
 #include <eve/detail/spy.hpp>
 
-#if !defined(EVE_ABI_DETECTED)
+#if !defined(EVE_CURRENT_ABI)
 #  define EVE_CURRENT_ABI ::eve::emulated_
 #  define EVE_CURRENT_API ::eve::undefined_simd_
-#  define EVE_STRICT_EMULATION
```
jfalcou, line 15: Are we sure `EVE_STRICT_EMULATION` is not referenced anywhere else ?

examples/algorithms/using_existing/inclusive_scan_par_unseq.cpp
```diff
@@ -308,7 +308,7 @@ TTS_CASE("inclusive_scan")
             T{3});
 };
 
-#if !defined(EVE_HW_POWERPC) // calling std::async crashes in our quemu run.
```
jfalcou, line 311: I am not sure about removing those HW macros. We need to talk about it.

jfalcou (approved):

PR #2008, opened 2024-10-15
Support compress for RISC-V

ita-sc:
Hi

This patch adds support for `compress` for RISC-V.

Also I've added some needed support for `store`, `count_true` and minor support for `convert` implemented. For now unit tests for `convert` can not be added, as we have some problems with saturated arithmetics here.

### DIFF
--- a/.github/workflows/unit.yml
+++ b/.github/workflows/unit.yml
@@ -147,9 +147,11 @@ jobs:
           mkdir build && cd build
           cmake .. -G Ninja -DEVE_OPTIONS="${{ matrix.cfg.opts }}" -DCMAKE_TOOLCHAIN_FILE=../cmake/toolchain/${{ matrix.cfg.comp }}.${{ matrix.cfg.arch }}.cmake
       - name: Compiling Unit Tests
-        run:  cd build && ninja unit.meta.exe unit.arch.exe  unit.core.bit_cast.exe unit.api.regular.wide.exe unit.core.simd_cast.exe -j 5
+        run:  cd build && ninja unit.meta.exe unit.arch.exe  unit.core.bit_cast.exe unit.api.regular.wide.exe unit.core.simd_cast.exe unit.core.count_true.exe unit.memory.store.exe unit.memory.store_equivalent.exe -j 5
+      - name: Compiling Big Unit Tests
+        run:  cd build && ninja unit.api.compress.exe -j 2
       - name: Running Unit Tests
-        run:  cd build && ctest --output-on-failure -j 4  -R "^unit.meta.*.exe|^unit.arch.*.exe|unit.core.bit_cast.exe|unit.api.regular.wide.exe|unit.core.simd_cast.exe"
+        run:  cd build && ctest --output-on-failure -j 4  -R "^unit.meta.*.exe|^unit.arch.*.exe|unit.core.bit_cast.exe|unit.api.regular.wide.exe|unit.core.simd_cast.exe|^unit.api.compress.*|unit.core.count_true.exe|unit.memory.store.*"
 
   ##################################################################################################
   ## Mac OS X Targets
--- a/include/eve/module/core/compress/simd/riscv/compress.hpp
+++ b/include/eve/module/core/compress/simd/riscv/compress.hpp
@@ -1,0 +1,27 @@
+//==================================================================================================
+/*
+  EVE - Expressive Vector Engine
+  Copyright : EVE Project Contributors
+  SPDX-License-Identifier: BSL-1.0
+*/
+//==================================================================================================
+#pragma once
+#include <eve/conditional.hpp>
+namespace eve::detail
+{
+
+template<relative_conditional_expr C, typename T, typename N, typename U>
+EVE_FORCEINLINE auto
+compress_(EVE_SUPPORTS(rvv_), C c, wide<T, N> v, logical<wide<U, N>> mask) noexcept
+requires rvv_abi<abi_t<T, N>>
+{
+  auto                c_mask        = expand_mask(c, as<wide<T, N>> {});
+  auto                formated_mask = simd_cast(mask, as<logical<wide<T, N>>> {});
+  logical<wide<T, N>> real_mask     = __riscv_vmand(c_mask, formated_mask, N::value);
+  wide<T, N>          init {0};
+  wide<T, N>          compressed      = __riscv_vcompress_tu(init, v, real_mask, N::value);
+  auto                new_element_num = count_true(real_mask);
+  kumi::tuple         cur {compressed, new_element_num};
+  return kumi::tuple<decltype(cur)> {cur};
+}
+}
--- a/include/eve/module/core/regular/impl/simd/riscv/count_true.hpp
+++ b/include/eve/module/core/regular/impl/simd/riscv/count_true.hpp
@@ -1,0 +1,48 @@
+//==================================================================================================
+/*
+  EVE - Expressive Vector Engine
+  Copyright : EVE Project Contributors
+  SPDX-License-Identifier: BSL-1.0
+*/
+//==================================================================================================
+#pragma once
+
+#include <eve/concept/conditional.hpp>
+#include <eve/concept/value.hpp>
+#include <eve/detail/implementation.hpp>
+
+namespace eve::detail
+{
+template<scalar_value T, typename N, relative_conditional_expr C>
+EVE_FORCEINLINE std::ptrdiff_t
+                count_true_(EVE_SUPPORTS(cpu_), C cond, logical<wide<T, N>> v) noexcept
+requires rvv_abi<abi_t<T, N>>
+{
+  if constexpr( C::is_complete )
+  {
+    if constexpr( !C::is_inverted ) return 0;
+    else return __riscv_vcpop(v, N::value);
+  }
+  else
+  {
+    auto const m = cond.mask(as<wide<T, N>> {});
+    return __riscv_vcpop(m, v, N::value);
+  }
+}
+
+template<scalar_value T, typename N>
+EVE_FORCEINLINE std::ptrdiff_t
+                count_true_(EVE_SUPPORTS(rvv_), logical<wide<T, N>> v) noexcept
+requires rvv_abi<abi_t<T, N>>
+{
+  return count_true[ignore_none](v);
+}
+
+template<scalar_value T, typename N>
+EVE_FORCEINLINE std::ptrdiff_t
+                count_true_(EVE_SUPPORTS(rvv_), top_bits<logical<wide<T, N>>> v) noexcept
+requires rvv_abi<abi_t<T, N>>
+{
+  return count_true[ignore_none](v.storage);
+}
+}
--- a/include/eve/module/core/regular/impl/simd/riscv/store.hpp
+++ b/include/eve/module/core/regular/impl/simd/riscv/store.hpp
@@ -1,0 +1,77 @@
+//==================================================================================================
+/*
+  EVE - Expressive Vector Engine
+  Copyright : EVE Project Contributors
+  SPDX-License-Identifier: BSL-1.0
+*/
+//==================================================================================================
+#pragma once
+#include <eve/arch/riscv/rvv_common_masks.hpp>
+#include <eve/concept/value.hpp>
+#include <eve/detail/category.hpp>
+#include <eve/detail/implementation.hpp>
+#include <eve/module/core/regular/unalign.hpp>
+
+namespace eve::detail
+{
+
+template<arithmetic_scalar_value T, typename N, simd_compatible_ptr<wide<T, N>> Ptr>
+EVE_FORCEINLINE void
+riscv_store(logical<wide<T, N>> mask, wide<T, N> v, Ptr p)
+requires rvv_abi<abi_t<T, N>>
+{
+  constexpr auto c = categorize<wide<T, N>>();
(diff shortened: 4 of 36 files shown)


include/eve/module/core/regular/impl/simd/riscv/store.hpp
```diff
@@ -0,0 +1,70 @@
...
+EVE_FORCEINLINE void
+store_(EVE_SUPPORTS(rvv_), C const& cond, wide<T, N> const& v, Ptr ptr) noexcept
+requires rvv_abi<abi_t<T, N>> && (!has_store_equivalent<wide<T, N>, Ptr>)
+{
+  if constexpr( C::is_complete || C::has_alternative || N() != expected_cardinal_v<T> )
+  {
+    store_(EVE_RETARGET(cpu_), cond, v, ptr);
+  }
+  else riscv_store(cond.mask(as<wide<T, N>> {}), v, unalign(ptr));
+}
```
DenisYaroshevskiy, line 68: N() != expected cardinal strikes me as weird. expected cardinal is enormous. why can't you do a masked store for small register

test/unit/api/compress/compress_copy_test.hpp
```diff
@@ -9,6 +9,29 @@
 
 #include "test.hpp"
 
+template<typename Type> struct rvv_compress_filter
```
DenisYaroshevskiy, line 12: Why is rvv compress filter different from any other rvv filter. This doesn't seem right

test/unit/api/compress/compress_copy_test.hpp
```diff
@@ -151,7 +174,8 @@ compress_copy_tst_mask(eve::as<T> tgt, L m, auto algo)
       for( int j = 0; j != T::size() - i; ++j ) { for_ignore_tests(eve::ignore_extrema(i, j)); }
     }
   }
-  else if constexpr( eve::has_bundle_abi_v<T> && eve::current_api >= eve::sve )
+  else if constexpr( eve::has_bundle_abi_v<T> && eve::current_api >= eve::sve
+                     && eve::current_api >= eve::rvv )
```
DenisYaroshevskiy, line 178: && doesn't seem right. this is always false.

ita-sc: Yep, fixed.

test/unit/api/compress/compress_copy_test.hpp
ita-sc: Actually, this is the main reason why I start with compress  (and next patch will be about shuffles). They require a lot of memory to compile even with global reduction for testcases. So, for now I've created an additional filter for these tests, though it is still not clear how to make it better.

include/eve/module/core/regular/impl/simd/riscv/store.hpp
ita-sc: Yep, removed. Also tried to make some optimizations for store. Take a look please.
Note: I have some doubts that the case of `C::is_complete && !C::is_inverted` is good (I just do nothing in this case), though it seems like the tests pass.

test/unit/api/compress/compress_wide.cpp
```diff
@@ -1,9 +1,12 @@
...
 
-TTS_CASE_WITH( "Check compress behavior"
-        , eve::test::simd::all_types
-        , tts::generate(tts::ramp(1))
-        )
+#ifdef SPY_SIMD_IS_RISCV_FIXED_RVV
+using types = tts::filter<rvv_compress_filter, eve::test::simd::all_types>::type;
+#else
+using types = eve::test::simd::all_types;
+#endif
```
DenisYaroshevskiy, line 7: I'd suggest: 

put this in comoress tests

call it simd_types_for_compress

ita-sc: Done

ita-sc:
Gentle ping

include/eve/module/core/compress/simd/riscv/compress.hpp
```diff
@@ -0,0 +1,27 @@
...
+
+template<relative_conditional_expr C, typename T, typename N, typename U>
+EVE_FORCEINLINE auto
+compress_(EVE_SUPPORTS(rvv_), C c, wide<T, N> v, logical<wide<U, N>> mask) noexcept
+requires rvv_abi<abi_t<T, N>>
+{
+  auto                c_mask          = expand_mask(c, as<wide<T, N>> {});
+  auto                formated_mask   = simd_cast(mask, as<logical<wide<T, N>>> {});
+  logical<wide<T, N>> real_mask       = __riscv_vmand(c_mask, formated_mask, N::value);
+  wide<T, N>          init            = make(as<wide<T, N>> {}, static_cast<T>(0));
```
DenisYaroshevskiy, line 21: `wide<T, N> init(0)` or `eve::zero(eve::as(v))`

include/eve/module/core/regular/impl/simd/riscv/count_true.hpp
```diff
@@ -0,0 +1,39 @@
...
+
+#include <eve/concept/conditional.hpp>
+#include <eve/concept/value.hpp>
+#include <eve/detail/implementation.hpp>
+
+namespace eve::detail
+{
+template<scalar_value T, typename N, relative_conditional_expr C>
+EVE_FORCEINLINE std::ptrdiff_t
+                count_true_(EVE_SUPPORTS(cpu_), C cond, logical<wide<T, N>> v) noexcept
```
DenisYaroshevskiy, line 18: You should check for `C::is_complete` and not call the masked versions for those.

(is_complete is ignore_all, ignore_none)
ignore_all - is_inverted == false
ignore_none - is_inverted == true

include/eve/module/core/regular/impl/simd/riscv/count_true.hpp
```diff
@@ -0,0 +1,39 @@
...
+{
+  return __riscv_vcpop(v, N::value);
+}
+template<scalar_value T, typename N>
+EVE_FORCEINLINE std::ptrdiff_t
+                count_true_(EVE_SUPPORTS(rvv_), top_bits<logical<wide<T, N>>> v) noexcept
+requires rvv_abi<abi_t<T, N>>
+{
+  return __riscv_vcpop(v.storage, N::value);
+}
```
DenisYaroshevskiy, line 38: I'd prefere if for both of these you'd call the ignore version and handle masked not masked there.

include/eve/module/core/regular/impl/simd/riscv/store.hpp
DenisYaroshevskiy: Yeah - that terminology is weird. !C::is_inverted means everything is ignored.

include/eve/module/core/regular/impl/simd/riscv/store.hpp
```diff
@@ -0,0 +1,77 @@
...
+         relative_conditional_expr       C,
+         simd_compatible_ptr<wide<T, N>> Ptr>
+EVE_FORCEINLINE void
+store_(EVE_SUPPORTS(rvv_), C const& cond, wide<T, N> const& v, Ptr ptr) noexcept
+requires rvv_abi<abi_t<T, N>> && (!has_store_equivalent<wide<T, N>, Ptr>)
+{
+  auto p = unalign(ptr);
+  if constexpr( C::is_complete )
+  {
+    if constexpr( !C::is_inverted ) return;
```
DenisYaroshevskiy, line 66: !is_inverted but has alternative should write the thing.

https://github.com/jfalcou/eve/blob/c2dacd782c29177ffc4b8ef86a900d6bc6389e67/include/eve/module/core/regular/impl/store.hpp#L119

Did you enable the tests for store?

test/unit/api/compress/compress_test_filter.hpp
```diff
@@ -0,0 +1,45 @@
...
+    constexpr auto cardinal          = eve::cardinal_v<Type>;
+    constexpr auto expected_cardinal = eve::expected_cardinal_v<scalar_t>;
+    // and double
+    if( sizeof(scalar_t) != 8 || !std::is_floating_point_v<scalar_t> ) return false;
+    // for expected cardinal (lmul==8)
+    if( expected_cardinal == cardinal ) return true;
+    // for combined type
+    if( 2 * expected_cardinal == cardinal ) return true;
+    return false;
+  }
```
DenisYaroshevskiy, line 28: I find this function way too difficult to parse through. What's bit_size / 8.

sizeof(Type) <= bit_size / 8 should be true for all types. What are you trying to do. 

If you explain in plain english what you want, we can do figure out how to write it.

DenisYaroshevskiy (commented):
Sorry for slow responses.

Most of this is ~fine - some small things.

The reduction in test cases is where we have most problems.

include/eve/module/core/regular/impl/simd/riscv/store.hpp
jfalcou: we shoudl probably have proper concept for that instead of those mysterious naming scheme

.github/workflows/unit.yml
```diff
@@ -147,9 +147,9 @@ jobs:
           mkdir build && cd build
           cmake .. -G Ninja -DEVE_OPTIONS="${{ matrix.cfg.opts }}" -DCMAKE_TOOLCHAIN_FILE=../cmake/toolchain/${{ matrix.cfg.comp }}.${{ matrix.cfg.arch }}.cmake
       - name: Compiling Unit Tests
-        run:  cd build && ninja unit.meta.exe unit.arch.exe  unit.core.bit_cast.exe unit.api.regular.wide.exe unit.core.simd_cast.exe -j 5
+        run:  cd build && ninja unit.meta.exe unit.arch.exe  unit.core.bit_cast.exe unit.api.regular.wide.exe unit.core.simd_cast.exe unit.api.compress.exe unit.core.count_true.exe unit.memory.store.exe unit.memory.store_equivalent.exe -j 3
```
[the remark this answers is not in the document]
jfalcou: You can duplicate the section and have the non compress test run at 5 and the other at previous # of jobs

include/eve/module/core/regular/impl/simd/riscv/count_true.hpp
ita-sc: You mean call there `eve::count_true[eve::ignore_none](...)`? Done

test/unit/api/compress/compress_test_filter.hpp
DenisYaroshevskiy: maximum register of 16 bytes is not good enough. It means you never test any multiple bigger than 1. 

What is your problem? compiler running out of memory?

DenisYaroshevskiy: We already have a very large filtering on types.
Because not all of the interesting types are in the pool, it is very hard to judge what's tested. 

What if we remove just some of the signs?

Smth like:

```
return eve::detail::is_one_of<eve::element_type_t<T>>(
 types<std::int8_t, std::uint16_t, float, double>{}
);
```

is one of function is defined here: https://github.com/jfalcou/eve/blob/11c8d8c8ee2c9b42fa3a25613d6f5bc1efece991/include/eve/detail/meta.hpp#L75

I feel that in order to understand what you are doing here otherwise, you need to go into test and see what's removed there.

.github/workflows/unit.yml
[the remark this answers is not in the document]
DenisYaroshevskiy: This is definetly something to have a look at, if you have troubles with memory.

test/unit/api/compress/compress_test_filter.hpp
ita-sc: You mean in compress filter try to specify exact types, instead of reducing fully? Yep, let's try. You suggest to check `std::int8_t, std::uint16_t, float, double` only?

.github/workflows/unit.yml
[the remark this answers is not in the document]
ita-sc: Yep, done

test/unit/api/compress/compress_test_filter.hpp
DenisYaroshevskiy: > `exact types, instead of reducing fully`? 

What do you mean by reducing fully? I'm just looking for some filtering that make more sense.

ita-sc: Please take a look now. Did I understand you right?

test/tts/tts.hpp
```diff
@@ -445,6 +445,21 @@ namespace tts
...
+  is_one_of(types<Ts...> const&) noexcept
+  {
+    bool found[] = {std::is_same_v<T, Ts>...};
+
+    return [&]<std::size_t... I>(std::index_sequence<I...> const&)
+    {
+      return (false || ... || found[I]);
+    }
+    (std::make_index_sequence<sizeof...(Ts)> {});
+  }
```
DenisYaroshevskiy, line 461: Yhis shouldn't be in tts

Just use https://github.com/jfalcou/eve/blob/11c8d8c8ee2c9b42fa3a25613d6f5bc1efece991/include/eve/detail/meta.hpp#L75

A;so you don't need the nested lambda

DenisYaroshevskiy (commented):
Please remove your  is one of and gtg

test/tts/tts.hpp
ita-sc: Thanks, fixed.

DenisYaroshevskiy (approved):

PR #2010, opened 2024-10-19
default_behavior map rework

SadiinsoSnowfall:
- Rework the internals of default_behavior to allow for more flexibility in some callables (ex. `logical_xxx`).
- Remove the EVE_DISPATCH_CALL macro (this combined with the previous changed lead to consistant single digit improvements in compile time).
- Updated all uses of `<callable>.behavior` and `eve::detail::map`
- Some general cleanups

### DIFF
--- a/include/eve/traits/overload/default_behaviors.hpp
+++ b/include/eve/traits/overload/default_behaviors.hpp
@@ -106,12 +106,12 @@ namespace eve
     constexpr EVE_FORCEINLINE auto adapt_call(as<R> pt, auto a, O const& o, T x, Ts const&... xs) const
     {
       constexpr bool has_implementation         = requires{ func_t::deferred_call(a, o, x, xs...); };
-      constexpr bool supports_map_no_conversion = requires{ map(pt, this->derived(), x, xs...); };
+      constexpr bool supports_map_no_conversion = requires{ detail::map(pt, this->derived(), x, xs...); };
       constexpr bool any_emulated               = (has_emulated_abi_v<T> || ... || has_emulated_abi_v<Ts>);
       constexpr bool any_aggregated             = (has_aggregated_abi_v<T> || ... || has_aggregated_abi_v<Ts>);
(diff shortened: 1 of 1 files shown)


include/eve/as.hpp
```diff
@@ -31,6 +31,14 @@ namespace eve
...
-    explicit constexpr as(T const&)  noexcept {}
+    explicit constexpr as(T const&) noexcept {}
+  };
+
+  template<>
+  struct as<void>
+  {
+    using type = void;
+
+    constexpr as() noexcept {}
```
DenisYaroshevskiy, line 42: surprised that this is needed

include/eve/detail/diff_div.hpp
```diff
@@ -33,8 +33,8 @@ namespace eve::detail
                                        ) noexcept
   {
     auto derivexp = (sizeof(element_type_t<T>) == 4) ? -12 :-26;
-    auto e = ldexp(one(as(x)), derivexp);
-    auto h = eve::max(e*ldexp(one(as(x)), exponent(x)), e);
+    auto e = ldexp(one(as{x}), derivexp);
+    auto h = eve::max(e*ldexp(one(as{x}), exponent(x)), e);
```
DenisYaroshevskiy, line 37: why do you change these. I like the () better here

DenisYaroshevskiy (commented):
so many just formatting and stylistic changes mixed within. why?

DenisYaroshevskiy:
why remove eve from eve::as everywhere
what's the point
especially mixed with all the other stuff. 900 files pr is unreviewable.

DenisYaroshevskiy:
Big sweeping changes need to be done in stages. Robots running separately, as much as possibly.
If there is a point to them at all, because I am not sure there is much need for removing the eve::.

DenisYaroshevskiy:
sorry that it will take some work to unravel.

PR #2025, opened 2024-11-21
Cleanup useless/non-constexpr branchs

SadiinsoSnowfall:


jfalcou:
I am OK with the x86 fixes.
The shuffle ones, aren't those function only use at constexpr anyway ? Is the correct fix to make them co steal instead ? Same for top_bits::is_cheap().

DenisYaroshevskiy:
I am against replaicing constexpr with consteval:
1. there is no reason.
2. Now I can't log in them for testing

include/eve/module/core/named_shuffles/swap_adjacent.hpp
```diff
@@ -65,52 +65,52 @@ struct swap_adjacent_t
   }
 
   template<simd_value T, std::ptrdiff_t G>
-  static constexpr std::ptrdiff_t level(eve::as<T> tgt, eve::fixed<G> g)
```
DenisYaroshevskiy, line 68: these tests in `level` were deliberately not if constexpr where posssible. 
main consideration is that if constexpr is probably slower to compile.
it also, for example, does not to do else branches.

include/eve/module/core/named_shuffles/reverse.hpp
```diff
@@ -55,79 +55,84 @@ namespace eve
...
     if constexpr( eve::has_aggregated_abi_v<T> )
     {
       if constexpr( G == T::size() / 2 ) return 0;
-      using half_t = decltype(T {}.slice(lower_));
-      return level(as<half_t> {}, g);
+      else
+      {
+        using half_t = decltype(T {}.slice(lower_));
+        return level(as<half_t> {}, g);
+      }
```
DenisYaroshevskiy, line 73: I'm not sure thids change improves things. maybe to not compile level(half)

include/eve/module/core/named_shuffles/broadcast_lane.hpp
```diff
@@ -80,47 +80,47 @@ struct broadcast_lane_t
...
-      if (current_api == avx) return 4;
-      if (g_size >= 8) return 2;
-      if (g_size >= 4) return 3;
-      if (g_size >= 2 && current_api == avx512) return 3;
-      return 4;
+      if constexpr (g_size >= 16) return 2;
+      if constexpr (current_api == avx) return 4;
+      if constexpr (g_size >= 8) return 2;
+      if constexpr (g_size >= 4) return 3;
+      if constexpr (g_size >= 2 && current_api == avx512) return 3;
```
DenisYaroshevskiy, line 114: you see - now you push things to comile time switching instead of just evaluating. I don't know how this works under the hood, but it seems worse.

include/eve/arch/cpu/top_bits.hpp
```diff
@@ -106,32 +106,32 @@ namespace detail
...
 
-    EVE_FORCEINLINE static constexpr std::ptrdiff_t bits_per_element_impl()
+    EVE_FORCEINLINE static consteval std::ptrdiff_t bits_per_element_impl()
     {
       if constexpr ( is_aggregated ) return top_bits<half_logical>::bits_per_element;
       else                           return decltype(movemask(logical_type{}).second){}();
     }
 
-    static constexpr bool is_cheap_impl()
+    static consteval bool is_cheap_impl()
```
DenisYaroshevskiy, line 115: same story here - we don't want the ifs to be constexpr when t doesn't have to. And I don't like consteval for no reason.

DenisYaroshevskiy (commented):
Please put it back - constexpr functions are better like they were

DenisYaroshevskiy:
> a good part of these branchs were already if-constexpr.

Where I had to.

>  these branchs depends on template parameters, wouldn't marking them as constexpr help with the compile times?

this is my thinking, which may be wrong. I do not know for sure. I work with some people I can ask - so I will when I have the chance.

When you do an `if constexpr` - the condition has to evaluate during compilation of that constexpr function.
When you do an `if` it evaluates during the execution of that constexpr function.

I suspect that if constexpr mechanism is similar to some template stuff and might be heavier on the compiler.
Wether or not that's true - I don't actually know.

We can park this until I confirm.

DenisYaroshevskiy:
> My reasoning is as follows: since these branches can be marked as constexpr, the compiler will likely eliminate them at compile time

everything here is compile time! We want less compile time.

jfalcou (approved):

PR #2040, opened 2025-01-08
Fix typo in compiler flag in setup docs

maxnoe:


DenisYaroshevskiy:
Thanks!

maxnoe:
Just in case I come across other things: is there a something setup in this repo to skip the extensive CI test for such simple docs changes?

PR #2044, opened 2025-01-09
Moved swap_pair into its callable

SadiinsoSnowfall:
Also fixed bit_swap_pair et fixed some tests for the swap_pair function family

### DIFF
--- a/include/eve/module/core/regular/swap_pairs.hpp
+++ b/include/eve/module/core/regular/swap_pairs.hpp
@@ -23,7 +23,6 @@ namespace eve
     EVE_CALLABLE_OBJECT(swap_pairs_t, swap_pairs_);
   };
 
-  //TODO DOC
   //================================================================================================
   //! @addtogroup core_bitops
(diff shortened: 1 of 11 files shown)


include/eve/module/core/regular/swap_pairs.hpp
```diff
@@ -11,47 +11,58 @@
...
+    template<simd_value T, std::ptrdiff_t I0, std::ptrdiff_t I1>
+    EVE_FORCEINLINE T operator()(T x, index_t<I0> i0, index_t<I1> i1) const noexcept
+    {
+      return EVE_DISPATCH_CALL(x, i0, i1);
+    }
+
+    EVE_CALLABLE_OBJECT(swap_pairs_t, swap_pairs_);
+  };
+
+  //TODO DOC
```
jfalcou, line 26: can we just fix the doc then ?

jfalcou (changes requested):

Issue #1604, opened 2023-05-17
[BUG] lousy unsigned number comparison on intel


DenisYaroshevskiy:
THere is some things with min and what not

Correct solution is:

```
eve::min(a, b) == a;
```

PR #2048, opened 2025-01-16
Move self_less to the is_less callable

SadiinsoSnowfall:


include/eve/concept/element_type.hpp
```diff
@@ -0,0 +1,20 @@
...
+#include <eve/concept/simd.hpp>
+#include <eve/traits/element_type.hpp>
+
+namespace eve
+{
+  template<typename T, typename... Ts>
+  concept same_element_type = (std::same_as<element_type_t<T>, element_type_t<Ts>> || ...);
+
+  template<typename T, typename... Ts>
+  concept same_element_type_or_scalar = (scalar_value<T> || ... || scalar_value<Ts>) || same_element_type<T, Ts...>;
```
DenisYaroshevskiy, line 19: This one is definitely not what you want. 
And the concept file name `element_type` is not ideal, you won't be able to guess where it is.

Let's think. 

On the surface i want to say: `same_or_scalar`. 

And then your call looks like: 

```
template <arithmetic_value T, arithmetic_value U>
  requires same_or_scalar<T, U>
```

But I think there is also some merit in `compatible_arithmetic_values`.
Then, in the future, when we talk about translations, we can add those here as well. 

What do you think?

include/eve/concept/value.hpp
```diff
@@ -151,6 +151,25 @@ namespace eve
...
+  //! `eve::arithmetic_simd_value` or `eve::arithmetic_scalar_value`.
+  //!
+  //! @groupheader{Examples}
+  //! - `int`
+  //! - `eve::wide<double>`
+  //! - `kumi::tuple<double, int>`
+  //! - `eve::wide<kumi::tuple<double, int>>`
+  //================================================================================================
+  template <typename T>
+  concept arithmetic_value = arithmetic_simd_value<T> || arithmetic_scalar_value<T>;
```
DenisYaroshevskiy, line 168: The pair wise `compatible_arithmetic_values` would go here.

include/eve/module/core/regular/impl/is_less.hpp
```diff
@@ -0,0 +1,42 @@
...
+  Copyright : EVE Project Contributors
+  SPDX-License-Identifier: BSL-1.0
+*/
+//==================================================================================================
+#pragma once
+
+namespace eve::detail
+{
+  template<typename T, typename U, callable_options O>
+  EVE_FORCEINLINE constexpr common_logical_t<T, U> is_less_(EVE_REQUIRES(cpu_), O const& o, T a, U b) noexcept
```
DenisYaroshevskiy, line 13: I'm concerned about implicit coinvercions. 

I think that our implicit conversions might prove problematic here.

Let's restrict this to the same element_type at least.

include/eve/module/core/regular/impl/is_less.hpp
```diff
@@ -0,0 +1,42 @@
...
+  {
+    if constexpr (O::contains(definitely))
```
DenisYaroshevskiy, line 15: This seems like handling of integers won't be correct.
I'm not sure what this option does but for integrals, we should probably just drop it and restart.

include/eve/module/core/regular/impl/simd/arm/sve/is_less.hpp
```diff
@@ -0,0 +1,22 @@
...
+namespace eve::detail
+{
+  template<callable_options O, typename T, typename N>
+  EVE_FORCEINLINE logical<wide<T, N>> is_less_(EVE_REQUIRES(sve_), O const& opts, wide<T, N> a, wide<T, N> b) noexcept
+    requires sve_abi<abi_t<T, N>>
+  {
+    if constexpr (O::contains(definitely)) return is_less.behavior(cpu_{}, opts, a, b);
+    else                                   return svcmplt(sve_true<T>(), a, b);
+  }
+}
```
DenisYaroshevskiy, line 22: please set up the new line end of file thing.

include/eve/module/core/regular/impl/simd/riscv/is_less.hpp
```diff
@@ -0,0 +1,33 @@
...
+    {
+      return is_less.behavior(cpu_{}, opts, a, b);
+    }
+    else
+    {
+      constexpr auto c = categorize<wide<T, N>>();
+
+      if      constexpr (match(c, category::int_))   return __riscv_vmslt(a, b, N::value);
+      else if constexpr (match(c, category::uint_))  return __riscv_vmsltu(a, b, N::value);
+      else if constexpr (match(c, category::float_)) return __riscv_vmflt(a, b, N::value);
```
DenisYaroshevskiy, line 29: O_o this works? On sve this probably would require an explicit cast to logical due to runtime register sizes.

include/eve/module/core/regular/impl/simd/x86/is_less.hpp
```diff
@@ -19,58 +19,140 @@
...
+          else if constexpr (use_avx2 && c == category::uint8x32)  return unsigned_cmp(a, b);
+          else if constexpr (c == category::int64x2)               return map(lt, a, b);
+          else if constexpr (c == category::int32x4)               return _mm_cmplt_epi32(a, b);
+          else if constexpr (c == category::int16x8)               return _mm_cmplt_epi16(a, b);
+          else if constexpr (c == category::int8x16)               return _mm_cmplt_epi8(a, b);
+          else if constexpr (c == category::uint64x2)              return map(lt, a, b);
+          else if constexpr (c == category::uint32x4)              return unsigned_cmp(a, b);
+          else if constexpr (c == category::uint16x8)              return unsigned_cmp(a, b);
+          else if constexpr (c == category::uint8x16)              return unsigned_cmp(a, b);
+          else                                                     return aggregate(lt, a, b);
```
DenisYaroshevskiy, line 102: Is this aggregate doing anything?

include/eve/module/core/regular/impl/simd/x86/is_less.hpp
```diff
@@ -19,58 +19,140 @@
...
+          else if constexpr (use_avx2 && c == category::int8x32)   return _mm256_cmpgt_epi8(b, a);
+          else if constexpr (use_avx2 && c == category::uint8x32)  return unsigned_cmp(a, b);
+          else if constexpr (c == category::int64x2)               return map(lt, a, b);
+          else if constexpr (c == category::int32x4)               return _mm_cmplt_epi32(a, b);
+          else if constexpr (c == category::int16x8)               return _mm_cmplt_epi16(a, b);
+          else if constexpr (c == category::int8x16)               return _mm_cmplt_epi8(a, b);
+          else if constexpr (c == category::uint64x2)              return map(lt, a, b);
+          else if constexpr (c == category::uint32x4)              return unsigned_cmp(a, b);
+          else if constexpr (c == category::uint16x8)              return unsigned_cmp(a, b);
+          else if constexpr (c == category::uint8x16)              return unsigned_cmp(a, b);
```
DenisYaroshevskiy, line 101: Can you please check the codegen for unsigned_cmp?
We had this issue: https://github.com/jfalcou/eve/issues/1604

include/eve/module/core/regular/impl/is_less.hpp
```diff
@@ -0,0 +1,42 @@
...
+        else                                                   return is_less(element_type_t<U>{a}, b);
+      }
+      else
+      {
+        if      constexpr (simd_value<U>)                      return map(is_less, a, b);
+        else if constexpr (std::same_as<element_type_t<T>, U>) return is_less(a, T{b});
+        else                                                   return is_less(a, element_type_t<T>{b});
+      }
+    }
+  }
```
DenisYaroshevskiy, line 41: Just for me - where is the mask handled in generaic case?
I see handling in avx512 but that's it.

include/eve/module/core/regular/impl/is_less.hpp
jfalcou: we need to support `is_less(wide<float>, int)`

include/eve/module/core/regular/impl/is_less.hpp
jfalcou: It is handled in the callable

include/eve/module/core/regular/impl/is_less.hpp
DenisYaroshevskiy: The implicit conversions should be done by elementwise callable, not here. 
I am concerned that this might get called instead of being prior converted.

include/eve/module/core/regular/impl/is_less.hpp
```diff
@@ -0,0 +1,50 @@
...
+#pragma once
+
+namespace eve::detail
+{
+  template<typename T, typename U, callable_options O>
+  EVE_FORCEINLINE constexpr common_logical_t<T, U> is_less_(EVE_REQUIRES(cpu_), O const& o, T a, U b) noexcept
+  {
+    if constexpr (product_type<T> && product_type<U>)
+    {
+      return convert(kumi::to_tuple(a) < kumi::to_tuple(b), as_element<common_logical_t<T, U>>());
```
DenisYaroshevskiy, line 17: I bet this gives us an extra conversion some case. We should think about this more

include/eve/module/core/regular/impl/is_less.hpp
```diff
@@ -0,0 +1,50 @@
...
+    }
+    else if constexpr (O::contains(definitely))
+    {
+      static_assert(floating_value<T> && floating_value<U>, "[EVE] eve::is_less[definitely] only accepts floating point values.");
```
DenisYaroshevskiy, line 21: Meh. What if you float, int?
Also - probably this needs to be in the interface. I dunno, let's think about this.

include/eve/module/core/regular/impl/simd/x86/is_less.hpp
```diff
@@ -19,58 +19,184 @@
...
+          constexpr auto use_avx2   = current_api >= avx2;
+          constexpr auto use_sse4_1 = current_api >= sse4_1;
+          constexpr auto lt         = []<typename E>(E ev, E fv) { return as_logical_t<E>(ev < fv); };
+
+          [[maybe_unused]] auto unsigned_cmp = [](auto lhs, auto rhs)
+          {
+            using l_t     = logical<wide<T, N>>;
+            auto const sm = signmask(as<as_integer_t<wide<T, N>, signed>>{});
+            return bit_cast((bit_cast(lhs, as(sm)) - sm) < (bit_cast(rhs, as(sm)) - sm), as<l_t>{});
+          };
```
DenisYaroshevskiy, line 85: i thought you fix this. Is there no max/min?

include/eve/module/core/regular/impl/simd/x86/is_less.hpp
```diff
@@ -19,58 +19,184 @@
...
+
+          if      constexpr (use_avx2 && c == category::int64x4)   return _mm256_cmpgt_epi64(b, a);
+          else if constexpr (use_avx2 && c == category::uint64x4)  return unsigned_cmp(a, b);
+          else if constexpr (use_avx2 && c == category::int32x8)   return _mm256_cmpgt_epi32(b, a);
+          else if constexpr (use_avx2 && c == category::uint32x8)
+          {
+            return _mm256_xor_si256(_mm256_cmpeq_epi32(_mm256_min_epu32(a, b), b), _mm256_cmpeq_epi32(b, b));
```
DenisYaroshevskiy, line 92: No - just equal max. I'm not 100 what this is. Comparing self to self for equality.

include/eve/module/core/regular/impl/is_less.hpp
DenisYaroshevskiy: Like always false? Or what does it mean?

include/eve/module/core/regular/impl/simd/x86/is_less.hpp
DenisYaroshevskiy: I see - sucks

> The comparison with self is there to generate a vector of 1s

You don't need to teach the compilers how to generate that, don't be clever.
Also you have andnot - you don't need this.
Also also - please write this as `a != max(a, b)` - srsly. 
Here - compiler will do everything: https://godbolt.org/z/hMbzeo1MW

DenisYaroshevskiy: I don't think that codegen is a problem. I'd trust the compiler. There are ~no algorithms that can survive without having a zero somewhere.

include/eve/detail/logical_tie_breaker.hpp
```diff
@@ -0,0 +1,84 @@
...
+    {
+      if constexpr (integral_value<ea_t>)        return A{};
+      else                                       return B{};
+    }
+    // both types have the same size, signedness and integral-ness, they are functionally the same.
+    else
+    {
+      // handles the (long, long long) and (unsigned long, unsigned long long) cases
+      if constexpr (std::same_as<decltype(ea_t{} + eb_t{}), ea_t>) return A{};
+      else                                                         return B{};
```
DenisYaroshevskiy, line 57: this will go away with translation, I hope

include/eve/detail/logical_tie_breaker.hpp
```diff
@@ -0,0 +1,84 @@
...
+  struct logical_tie_breaker<A, B>
+  {
+    using type = decltype(logical_tie_breaker_impl<typename as_logical<A>::type, typename as_logical<B>::type>());
+  };
+
+  template <typename T>
+  struct logical_tie_breaker<T>
+  {
+    using type = typename as_logical<T>::type;
+  };
```
DenisYaroshevskiy, line 80: Use a fold expression. See common_type for example

include/eve/module/core/regular/impl/is_less.hpp
```diff
@@ -0,0 +1,48 @@
...
+  SPDX-License-Identifier: BSL-1.0
+*/
+//==================================================================================================
+#pragma once
+
+namespace eve::detail
+{
+  template<typename T, typename U, callable_options O>
+  EVE_FORCEINLINE constexpr common_logical_t<T, U> is_less_(EVE_REQUIRES(cpu_), O const& o, T a, U b) noexcept
+  {
```
DenisYaroshevskiy, line 14: I continue not liking that we accept different types here, instead of dealing with them at the callable level. I bet there are bugs hiding in this.

include/eve/module/core/regular/impl/is_less.hpp
DenisYaroshevskiy: noone answered me

include/eve/module/core/regular/impl/simd/x86/is_less.hpp
```diff
@@ -19,58 +19,151 @@
...
+          else if constexpr (use_avx2 && c == category::int16x16)  return _mm256_cmpgt_epi16(b, a);
+          else if constexpr (use_avx2 && c == category::uint16x16) return eve::min(a, b) != b;
+          else if constexpr (use_avx2 && c == category::int8x32)   return _mm256_cmpgt_epi8(b, a);
+          else if constexpr (use_avx2 && c == category::uint8x32)  return eve::min(a, b) != b;
+          else if constexpr (c == category::int32x4)               return _mm_cmplt_epi32(a, b);
+          else if constexpr (c == category::int16x8)               return _mm_cmplt_epi16(a, b);
+          else if constexpr (c == category::int8x16)               return _mm_cmplt_epi8(a, b);
+          else if constexpr (c == category::uint32x4)
+          {
+            if constexpr (use_sse4_1) return eve::min(a, b) != b;
```
DenisYaroshevskiy, line 100: I am surprised you didn't need to forward anything. I'm pretty sure max/min rely on less.
Can you check how that works?

include/eve/module/core/regular/is_less.hpp
```diff
@@ -22,16 +23,23 @@
...
   template<typename Options>
-  struct is_less_t : strict_elementwise_callable<is_less_t, Options, definitely_option>
+  struct is_less_t : elementwise_callable<is_less_t, Options, definitely_option>
   {
-    template<value T,  value U>
-    requires(eve::same_lanes_or_scalar<T, U>)
-    constexpr EVE_FORCEINLINE common_logical_t<T,U> operator()(T a, U b) const
+    template<floating_value T, floating_value U>
+    constexpr EVE_FORCEINLINE common_logical_t<T, U> operator()(T a, U b) const
+      requires (compatible_arithmetic_values<T, U> && Options::contains(definitely))
```
DenisYaroshevskiy, line 30: I'd just static assert in the operator().
We had some examples of that somewhere.

Also - what about the tuple types? @jfalcou

include/eve/detail/logical_tie_breaker.hpp
DenisYaroshevskiy: Wait - I don't hink you need variadic at all. Just don't have a variadic. There is already a variadic.

include/eve/module/core/regular/impl/is_less.hpp
jfalcou: The general masking (ie not using AVX512/SVE masking) is done in elementwise_callable that detect a non trivial mask, etxract it and call perform_mask.

See : https://github.com/jfalcou/eve/blob/71d23596c5d5ba1f008f1f729388d952aa46d6ce/include/eve/traits/overload/default_behaviors.hpp#L131

This has been the case since ages.

include/eve/module/core/regular/is_less.hpp
jfalcou: The question is : 
 - do we want SFINAEing out people use is_less[tolerance](int,int)
 - or do we want an hard error ?

DenisYaroshevskiy: Depends on what you want for tuple

include/eve/module/core/regular/impl/is_less.hpp
DenisYaroshevskiy:
```
          return detail::mask_op(cond, detail::return_2nd, x0, f(x0,xs...));
```

Yeah - ok - this doesn't seem right for is_less. I don't even understand how it works.

jfalcou: Good question I will check 
I am a bit wary about masking predicates anyway

include/eve/module/core/regular/is_less.hpp
jfalcou: Well we need a global protocol. Is using options when they aren't supposed too sfinae or hard error.

DenisYaroshevskiy: I think the answer will come from what you want for tuple. Can you look at that first please

DenisYaroshevskiy:
How to forward declare the common_logical: https://godbolt.org/z/v3T1Pofbn

include/eve/module/core/regular/is_less.hpp
```diff
@@ -22,16 +23,23 @@
...
     {
       //      static_assert( valid_tolerance<common_value_t<T, U>, Options>::value, "[eve::is_less] simd tolerance requires at least one simd parameter." );
       return EVE_DISPATCH_CALL(a, b);
     }
 
+    template<arithmetic_value T, arithmetic_value U>
+    constexpr EVE_FORCEINLINE common_logical_t<T, U> operator()(T a, U b) const
+      requires (compatible_arithmetic_values<T, U> && !Options::contains(definitely))
```
DenisYaroshevskiy, line 38: decided to static assert for now

include/eve/module/core/regular/is_less.hpp
```diff
@@ -22,11 +23,11 @@
...
   template<typename Options>
-  struct is_less_t : strict_elementwise_callable<is_less_t, Options, definitely_option>
+  struct is_less_t : elementwise_callable<is_less_t, Options, definitely_option>
   {
-    template<value T,  value U>
-    requires(eve::same_lanes_or_scalar<T, U>)
-    constexpr EVE_FORCEINLINE common_logical_t<T,U> operator()(T a, U b) const
+    template<arithmetic_value T, arithmetic_value U>
+    constexpr EVE_FORCEINLINE common_logical_t<T, U> operator()(T a, U b) const
+      requires (compatible_arithmetic_values<T, U>)
```
DenisYaroshevskiy, line 30: put the static assert in the interface please

DenisYaroshevskiy (approved):

jfalcou (approved):

PR #2054, opened 2025-01-27
Move self_leq to the is_less_equal callable

SadiinsoSnowfall:


jfalcou (commented):
I think this looks good, shoudl we make the non trivial map lambda forceinline ?

DenisYaroshevskiy:
This is draft still, right? Let's wait for @SadiinsoSnowfall  to be ready.

DenisYaroshevskiy (approved):

PR #2061, opened 2025-02-06
Made replace_ignored a new format callable

SadiinsoSnowfall:


include/eve/module/core/regular/replace.hpp
```diff
@@ -15,56 +15,54 @@
...
-//!
-//!   Same as if_else.
-//!
-//================================================================================================
-
-EVE_MAKE_CALLABLE(replace_ignored_, replace_ignored);
+  template<typename Options>
+  struct replace_ignored_t : callable<replace_ignored_t, Options>
+  {
+    template<simd_value Wide, conditional_expr Ignore, typename Other>
```
DenisYaroshevskiy, line 21: No to typename. In the new format we always specify the interfaces fully.

Now that I think about it - we should probably accept `eve::zero` and other things like that here. Can you maybe make that work?

include/eve/module/core/regular/replace.hpp
```diff
@@ -15,56 +15,54 @@
...
+    constexpr EVE_FORCEINLINE Wide operator()(Wide x, Ignore ignore, Other with) const noexcept
+    {
+      return eve::if_else(ignore, x, with);
+    }
+  };
 
-namespace detail
-{
+  // TODO DOC is it to be in user interface ?
```
DenisYaroshevskiy, line 28: you can remove this, it's good for interface.

include/eve/module/core/regular/replace.hpp
```diff
@@ -15,56 +15,54 @@
...
+  //!   A convinience wrapper around if_else.
+  //!
+  //!   @groupheader{Callable Signatures}
+  //!
+  //!   @code
+  //!   namespace eve
+  //!   {
+  //!      template< eve::value T, eve::conditional_expr Ignore, eve::value Other >
+  //!      auto replace_ignored(T x, Ignore ignore, Other with )
+  //!        -> decltype(eve::if_else(ignore, x, with))
```
DenisYaroshevskiy, line 49: result type is T.

jfalcou (approved):

PR #2066, opened 2025-02-12
Enable SVE2 support

jfalcou:


DenisYaroshevskiy (approved):
I'm OK with this but FYI - 128 bits seems like the most important one. We should be testing that.

jfalcou:
It's probably because we test 512 which is too large. I can go down to SVE 256 and SVE2 128 for now

DenisYaroshevskiy:
sve512 is a thing a bit. 512 sometimes exposes problems. I dunno, up to you. When I finally send the old computer, hopefully the CI gets better

PR #2078, opened 2025-03-13
Fix and streamline callable options combinations

jfalcou:
+ Fixed a bug where passing a options pack was always overwriting pre-existing options of the callable.
 + Fixed the order in which same key options are kept : last one wins in all case now.
 + Added tests for those situations.


DenisYaroshevskiy (commented):
Is it important that I understand this change?

jfalcou:
> Is it important that I understand this change?

This fixes the inconsistency we noticed and discussed yesterday. Now it make the removal of options work when we will do it.

PR #2079, opened 2025-03-13
Fixed x86 strict lower/upper impl

SadiinsoSnowfall:
The combination of callable options `strict`+`lower/upper`+`condition` of `sub`, `mul`, `div` and `add` was not handled properly on x86 AVX512 resulting in subtle bug / precision loss in some operations.


DenisYaroshevskiy:
Add a comment on what's broken in the description?

jfalcou (approved):

PR #2081, opened 2025-03-17
Drop the condition from the callable options in elementwise masked mode

SadiinsoSnowfall:
[first commit]: I expect this to fail on x86 AVX512 and ARM SVE (because the calls to `behavior` do not yet relay the condition.

### DIFF
--- a/test/unit/internals/options.cpp
+++ b/test/unit/internals/options.cpp
@@ -16,6 +16,12 @@ enum class tests
 , has_ignore_all
 };
 
+enum class OverloadType
+{
+  regular = 1
+, masked
+};
+
 namespace eve
 {
   template<typename Options>
@@ -65,13 +71,15 @@ namespace eve::detail
   template<eve::callable_options O, eve::conditional_expr C>
   auto ew_drop_func_(EVE_REQUIRES(cpu_), C const&, O const&, eve::wide<int>)
   {
-    return !O::contains(condition_key);
+    TTS_EXPECT(!O::contains(condition_key));
+    return OverloadType::masked;
   }
 
   template<eve::callable_options O>
   auto ew_drop_func_(EVE_REQUIRES(cpu_), O const&, eve::wide<int>)
   {
-    return !O::contains(condition_key);
+    TTS_EXPECT(!O::contains(condition_key));
+    return OverloadType::regular;
   }
 }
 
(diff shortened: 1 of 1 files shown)


jfalcou (approved):

include/eve/traits/overload/supports.hpp
```diff
@@ -63,6 +63,13 @@ namespace eve
       auto    dropped =  rbr::drop(kws, *this);
       return  kumi::tuple{value, options<decltype(dropped)>{dropped}};
     }
+
+    template <rbr::concepts::settings S>
+    constexpr EVE_FORCEINLINE auto operator&&(options<S> const& opt) const noexcept
+    {
+      auto new_opts = rbr::merge(opt, *this);
+      return options<decltype(new_opts)>{new_opts};
+    }
```
DenisYaroshevskiy, line 72: I am not wild about this overload. operator&& and other operators should only be overloaded if the meaning is super obvious.

What does this do?

include/eve/traits/overload/impl/strict_elementwise.hpp
```diff
@@ -49,9 +49,44 @@ namespace eve
...
+        auto [cond, rmv_cond] = opts.extract(condition_key);
+        using cond_t =  decltype(cond);
+        [[maybe_unused]] Func<decltype(rmv_cond)> const f{rmv_cond};
+
+        // Check that the mask and the value are of same kind if simd
+        constexpr bool compatible_mask = !(simd_value<decltype(cond.mask(as(x0)))> && (scalar_value<T> && ... && scalar_value<Ts>));
+        static_assert(compatible_mask, "[EVE] - Scalar values can't be masked by SIMD logicals.");
+
+        // Shush any other cascading errors
+        if constexpr(!compatible_mask) return ignore{};
```
DenisYaroshevskiy, line 62: you just static asserted this. Why if constexpr

include/eve/traits/overload/impl/conditional.hpp
```diff
@@ -19,50 +19,22 @@ namespace eve
...
-        if constexpr (match_option<condition_key, O, ignore_none_>)
-        {
-          return base_t::behavior(arch, opts, x0, xs...);
-        }
+        // Check that the mask and the value are of same kind if simd
+        constexpr bool compatible_mask = !(simd_value<decltype(opts[condition_key].mask(as(x0)))> && (scalar_value<T> && ... && scalar_value<Ts>));
+        static_assert(compatible_mask, "[EVE] - Scalar values can't be masked by SIMD logicals.");
+
+        // Shush any other cascading errors
+        if constexpr(!compatible_mask) return ignore{};
```
DenisYaroshevskiy, line 34: same question as below

DenisYaroshevskiy (commented):
Where are new tests?

include/eve/traits/overload/impl/strict_elementwise.hpp
DenisYaroshevskiy: Joel says for error messages. Ok then

PR #2090, opened 2025-04-10
Added ignore_extrema tests and debug asserts on relative_conditional_expr usage

SadiinsoSnowfall:
Adds tests for `ignore_extrema`.

Also add debug asserts on `relative_conditional_expr` construction and usage where applicable.
The goal of these asserts is to prevent users from constructing an invalid condition or using a condition in an invalid way.
The current behaviour in such cases is undefined and can lead to [unexpected results](https://godbolt.org/z/KcYE7oqYq).

### DIFF
--- a/include/eve/conditional.hpp
+++ b/include/eve/conditional.hpp
@@ -259,7 +259,7 @@ namespace eve
     }
 
     template<typename T>
-    EVE_FORCEINLINE constexpr void check_mask(eve::as<T> tgt) const
+    EVE_FORCEINLINE constexpr void check_mask(eve::as<T>) const
     {
       EVE_ASSERT(count_ <= T::size(), "[eve::keep_first] Invalid count");
     }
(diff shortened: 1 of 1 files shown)


test/unit/api/regular/conditional.cpp
```diff
@@ -418,7 +419,9 @@ TTS_CASE_TPL( "ignore_first/last behavior", eve::test::simd::all_types)
       if(fi+li <= type::size())
       {
         TTS_EQUAL( (ignore_first(fi) && ignore_last(li)).mask(as<type>()).bits(), mref.bits() );
+        TTS_EQUAL( (ignore_extrema(fi, li)).mask(as<type>()).bits(), mref.bits() );
         TTS_EQUAL( (if_else(ignore_first(fi) && ignore_last(li),value, type(69))), ref);
+        TTS_EQUAL( (if_else(ignore_extrema(fi, li), value, type(69))), ref);
```
DenisYaroshevskiy, line 424: the idea is that operator&& returns the ignore-extrema but ok

DenisYaroshevskiy (approved):

PR #2095, opened 2025-04-29
Made sum a new format callable

SadiinsoSnowfall:


test/unit/module/core/reduction_test.hpp
```diff
@@ -0,0 +1,232 @@
...
+
+    test_case<TruthFn>(callable, rhs1, runner);
+    test_case<TruthFn>(callable, rhs2, runner);
+    test_case<TruthFn>(callable, rhs3, runner);
+    test_case<TruthFn>(callable, rhs4, runner);
+  }
+}
+
+template<typename TruthFn, typename Callable, typename T>
+void arithmetic_test_case(Callable callable, T v, double expected_ulp = 0.0)
```
DenisYaroshevskiy, line 196: I feel like the word `reduction` is missing somewhere.

Why is just having the word test case enough?

include/eve/module/core/regular/impl/reduce.hpp
```diff
@@ -48,8 +48,12 @@ template<scalar_value T, typename N, typename Callable>
 EVE_FORCEINLINE auto
 reduce_(EVE_SUPPORTS(cpu_), splat_type const& s, wide<T, N> v, Callable f) noexcept
 {
-  auto op = find_reduction(f, s);
-  return op(s, v);
+  if constexpr (std::same_as<Callable, callable_add_>) return eve::detail::sum[splat](v);
```
DenisYaroshevskiy, line 51: find reduction can return sum.
I don't understand this change.

include/eve/detail/function/sum.hpp
```diff
@@ -10,17 +10,78 @@
...
+  //!     * `c`: [Conditional expression](@ref eve::conditional_expr) masking the operation.
+  //!     * `m`: [Logical value](@ref eve::logical_value) masking the operation.
+  //!
+  //!   **Return value**
+  //!
+  //!     1. The sum of all values in the input. Scalar values are returned as is.
+  //!     2. The sum of all values in the input splatted across every lane of said input.
+  //!     3. Same as the above but the masked lanes are ignored during the operation.
+  //================================================================================================
+  inline constexpr auto sum = functor<sum_t>;
```
DenisYaroshevskiy, line 81: It's in detail. I don't think it needs to be added to the doxygen. (this won't do anything anyways)

include/eve/detail/function/simd/common/sum.hpp
```diff
@@ -18,61 +18,70 @@
 
 namespace eve::detail
 {
-  template<arithmetic_scalar_value T, typename N>
-  EVE_FORCEINLINE auto sum_ ( EVE_SUPPORTS(cpu_)
-                            , splat_type const&, wide<T,N> const &v
+  template<callable_options O, arithmetic_scalar_value T, typename N>
```
DenisYaroshevskiy, line 21: What happened in this file?

include/eve/detail/function/simd/arm/sve/sum.hpp
```diff
@@ -12,42 +12,36 @@
...
-EVE_FORCEINLINE auto
-sum_(EVE_SUPPORTS(sve_), C const& cond, wide<T, N> v) noexcept -> T
-requires sve_abi<abi_t<T, N>>
-{
-  // Fix mask to not touch garbage lanes if it is not an ignore_*
-  auto m = cond.mask(as(v));
-  if constexpr(!relative_conditional_expr<C>) m = remove_garbage(m);
-  return svaddv(m, v);
-}
+  template<callable_options O, typename T, typename N>
```
DenisYaroshevskiy, line 17: Is this a full rewrite? I think we talked about this I just don't remember.

include/eve/detail/function/simd/arm/neon/sum.hpp
```diff
@@ -27,76 +27,77 @@ namespace eve::detail
     else  if constexpr( c== category::int8x8    ) return vpadd_s8(v,v);
   }
 
-  template<typename T, typename N>
-  EVE_FORCEINLINE wide<T,N> sum_( EVE_SUPPORTS(neon128_), splat_type const&, wide<T,N> v) noexcept
-  requires arm_abi<abi_t<T, N>>
+  template<callable_options O, typename T, typename N>
```
DenisYaroshevskiy, line 30: Are you mixing formatting and functional changes again?

include/eve/module/core/regular/impl/reduce.hpp
SadiinsoSnowfall: This is just temporary until all reductions callables are reworked.
`find_reduction` returns an object which is expected to be called with a `splat_type` parameter. We cannot do that anymore as it has been changed to a callable option.

include/eve/detail/function/simd/common/sum.hpp
SadiinsoSnowfall: Because the splatting mode and mask are now passed as callable options and not as function parameters anymore, I merged everything into two functions (splat & no-splat).
I also changed the code of the non-splat-wide-emulated `sum` impl to use our `for_` construct instead of an IILF + fold.

include/eve/detail/function/simd/arm/sve/sum.hpp
```diff
@@ -12,42 +12,36 @@
...
+    requires sve_abi<abi_t<T, N>>
+  {
+    if constexpr (O::contains(splat2))
+    {
+      // Attempts to force the use of a broadcasting mov
+      return wide<T, N>{ wide<T>{ sum[opts.drop(splat2)].retarget(current_api, v) } };
+    }
+    else
+    {
+      // s/uaddv is expensive, for small lanes count this approach is faster
```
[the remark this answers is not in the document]
SadiinsoSnowfall: Of course the best course of actions would be to benchmark that on a real neoverse / SVE-enabled CPU, but I don't have access to one.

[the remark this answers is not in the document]
SadiinsoSnowfall: Ok, it should be better now.

PR #2101, opened 2025-05-14
Fixed possible UB in SVE count_true and first_true

SadiinsoSnowfall:


include/eve/module/core/regular/impl/simd/arm/sve/count_true.hpp
```diff
@@ -17,7 +17,12 @@ namespace eve::detail
   EVE_FORCEINLINE std::ptrdiff_t count_true_(EVE_REQUIRES(sve_), O const& opts, logical<wide<T,N>> v) noexcept
     requires sve_abi<abi_t<T, N>>
   {
-    auto const m = expand_mask(opts[condition_key], as<wide<T, N>>{});
+    using C = rbr::result::fetch_t<condition_key, O>;
+    const auto cx = opts[condition_key];
+
+    logical<wide<T, N>> m;
+    if constexpr (relative_conditional_expr<C>) m = sve_true(cx, as(m));
+    else                                        m = remove_garbage(expand_mask(cx, as(m)));
```
DenisYaroshevskiy, line 25: I'm not sure what the problem was, can you elablorate.
I would not expect expand mask to produce garbage.

include/eve/module/core/regular/impl/simd/arm/sve/first_true.hpp
```diff
@@ -22,35 +22,44 @@ namespace eve::detail
...
+
+        // Compute the condition mask only if necessary, this gives slightly better codegen.
+        // This also masks the inactive lanes of the input.
+        if constexpr (!C::is_complete || !is_ec)
+        {
+          if      constexpr (relative_conditional_expr<C>) c_m = sve_true(cx, as(m));
+          else if constexpr (is_ec)                        c_m = expand_mask(cx, as<L>{});
+          // merge everything right away in this case to eliminate one and instruction
+          else                                             c_m = m && remove_garbage(expand_mask(cx, as<L>{}));
+        }
```
DenisYaroshevskiy, line 56: I appreciate that you written this in an older pr, but this took me a while to navigate, so let's fix it.

>Compute the condition mask only if necessary,

Not helpful comment. I guess you mean don't call `expand_mask`.

> merge everything right away in this case to eliminate one and instruction

I don't get it. m && here definetly seems wrong - you want to and them after svptest_any§

> remove_garbage

remove garbage here is clearly wrong. Because if you have no conditional, you forgot to remove garbage from the `m` and garbage just slips through.

include/eve/module/core/regular/impl/simd/arm/sve/first_true.hpp
```diff
@@ -20,44 +20,50 @@ namespace eve::detail
     auto cx = opts[condition_key];
 
     if constexpr (C::is_complete && !C::is_inverted) return std::nullopt;
+    else if constexpr (has_aggregated_abi_v<L>)
+    {
+      if constexpr (!C::is_complete)
+      {
+        if constexpr (relative_conditional_expr<C>) m = m && sve_true(cx, as(m));
+        else                                        m = m && expand_mask(cx, as<L>{});
+      }
```
DenisYaroshevskiy, line 29: I think we should teach expand mask for relative conditional to call sve_true. Ottherwise this is annoying

include/eve/module/core/regular/impl/simd/arm/sve/first_true.hpp
```diff
@@ -20,44 +20,50 @@ namespace eve::detail
...
 
-        return hi_res;
-      }
-      else
-      {
-        // Merging the two masks after the ptest makes this branch appear earlier in the resulting assembly.
-        if (!svptest_any(c_m, m)) return std::nullopt;
+      // Unconditional mask merge because we need to mask the inactive lanes of there are any.
+      // no-op in the has_inactive_lanes case.
+      m = m && c_m;
```
DenisYaroshevskiy, line 61: I don't get this. Why not: `svbrkb_z(c_m, m);`

DenisYaroshevskiy: Can you actually run tests and post the failing test please?

PR #2102, opened 2025-05-15
Add conditional any PPC impl

SadiinsoSnowfall:


include/eve/module/core/regular/impl/simd/ppc/any.hpp
```diff
@@ -10,27 +10,44 @@
...
   {
-    logical<wide<T>> mm = [](auto i, auto) { return i < N::value; };
-    m &= bit_cast(mm, as<logical<wide<T, N>>>()).bits();
+    auto m = simd_cast(v0, as<logical<wide<T>>>{}).bits();
+
+    if constexpr (N::value != expected_cardinal_v<T, ppc_>)
+    {
+      logical<wide<T>> mm = [](auto i, auto) { return i < N::value; };
+      m &= mm.bits();
+    }
```
DenisYaroshevskiy, line 42: I believe this is "remove_garbage"

include/eve/module/core/regular/impl/simd/ppc/any.hpp
```diff
@@ -10,27 +10,44 @@
...
+
+    if constexpr (!match_option<condition_key, O, ignore_none_>)
+    {
+      const auto cm = simd_cast(expand_mask(opts[condition_key], as<wide<T, N>>{}), as<logical<wide<T>>>{});
+      m &= cm.bits();
+    }
```
DenisYaroshevskiy, line 48: Hmm. Hmm. Hmm.

Here is what I don't like.

* If we have ignore none - we should remove garbage from the main register.
* If we have relative conditional - expand_mask from that could for free generate False everywhere where garbage might go
* If we have a non-relative conditional - expand_mask won't be able to zero out false elements for free.

^ We need some helper to handle all of this stuff.

DenisYaroshevskiy (approved):

include/eve/module/core/regular/impl/simd/ppc/any.hpp
```diff
@@ -10,27 +10,41 @@
...
+    else
+    {
+      if constexpr (match_option<condition_key, O, ignore_none_>)
+      {
+        m = remove_garbage(m);
+      }
+      else 
+      {
+        m &= expand_mask_no_garbage(opts[condition_key], as(v)).bits();
+      }
```
DenisYaroshevskiy, line 45: Meh, but w/e

PR #2106, opened 2025-05-28
Made load a new format callable

SadiinsoSnowfall:


test/unit/memory/load/tuple.cpp
```diff
@@ -59,8 +59,8 @@ TTS_CASE_TPL( "Check load behavior with soa_ptr", eve::test::scalar::all_types)
   TTS_EQUAL(constructed                                , reference   );
   TTS_EQUAL(eve::load(src)                             , reference   );
   TTS_EQUAL(eve::load(src, eve::lane<8>)               , reference8  );
-  TTS_EQUAL(eve::unsafe(eve::load)(src)                , reference   );
-  TTS_EQUAL(eve::unsafe(eve::load)(src, eve::lane<8>)  , reference8  );
+  TTS_EQUAL(eve::load[eve::unsafe](src)                , reference   );
```
DenisYaroshevskiy, line 62: Why here unsafe but in other place you have to use `eve::load[eve::unsafe2]`

DenisYaroshevskiy (commented):
algo changes look OK. THe load itself - didn't look yet.

include/eve/detail/function/load.hpp
```diff
@@ -10,12 +10,113 @@
...
+  struct load_t : relative_conditional_callable<load_t, Options, unsafe_option>
+  {
+    template<detail::data_source Ptr>
+    EVE_FORCEINLINE wide_value_type_t<Ptr> operator()(Ptr ptr) const noexcept
+    {
+      return EVE_DISPATCH_CALL(ptr, as<wide_value_type_t<Ptr>>{});
+    }
 
-  template<> struct supports_optimized_conversion<tag::load_> : std::true_type {};
+    template<detail::data_source Ptr, std::ptrdiff_t N>
```
DenisYaroshevskiy, line 26: concepts using detail is ofc not awesome, but this is not for you.

include/eve/detail/function/load.hpp
```diff
@@ -10,12 +10,113 @@
...
+      return EVE_DISPATCH_CALL(b, e, tgt);
+    }
+
+    EVE_CALLABLE_OBJECT(load_t, load_);
+  };
+
+  //================================================================================================
+  //! @addtogroup simd
+  //! @{
+  //!   @var load
```
DenisYaroshevskiy, line 59: @jfalcou  - docs in detail. Will they be rendered?

include/eve/module/core/regular/load.hpp
```diff
@@ -6,95 +6,4 @@
 //==================================================================================================
 #pragma once
 
-#include <eve/arch.hpp>
```
DenisYaroshevskiy, line 9: Maybe actual load should be in this file instead of detail.

include/eve/detail/function/simd/common/load.hpp
```diff
@@ -19,135 +21,37 @@
...
   {
     Wide that;
 
-    auto cast = []<typename Ptr, typename Sub>(Ptr p, as<Sub>)
+    auto cast = []<typename DS, typename Sub>(DS p, as<Sub>)
     {
       using type = element_type_t<Wide>;
       using a_p = eve::aligned_ptr<const type, cardinal_t<Sub>>;
-      if constexpr (std::is_pointer_v<Ptr>) return p;
+      if constexpr (std::is_pointer_v<DS>) return p;
```
DenisYaroshevskiy, line 54: formatting

include/eve/detail/function/simd/common/load.hpp
```diff
@@ -164,105 +68,189 @@ namespace eve::detail
...
+  EVE_FORCEINLINE Wide load_(EVE_REQUIRES(cpu_), O const&, It src, [[maybe_unused]] It end, as<Wide> tgt) noexcept
   {
-    return load(ignore_none, safe, as<Wide>{}, b, e);
+    if constexpr (logical_simd_value<Wide> && !Wide::abi_type::is_wide_logical)
+    {
+      return to_logical(load(src, end, as<as_arithmetic_t<Wide>>{}));
+    }
+    else
+    {
+      return piecewise_load(src, tgt);
```
DenisYaroshevskiy, line 83: For my sanity - where is the assert that there is enough space?

include/eve/detail/function/simd/common/load.hpp
```diff
@@ -164,105 +68,189 @@ namespace eve::detail
...
-    return load(ignore_none, safe, as<as_wide_t<base>>{}, b, e);
-  }
 
-  template<std::input_iterator Iterator, typename Wide>
-  EVE_FORCEINLINE auto load_(EVE_SUPPORTS(cpu_), as<Wide> const&, Iterator b, Iterator e) noexcept
+  template<callable_options O, std::input_iterator It, typename Wide>
+  EVE_FORCEINLINE Wide load_(EVE_REQUIRES(cpu_), O const&, It src, [[maybe_unused]] It end, as<Wide> tgt) noexcept
   {
-    return load(ignore_none, safe, as<Wide>{}, b, e);
+    if constexpr (logical_simd_value<Wide> && !Wide::abi_type::is_wide_logical)
```
DenisYaroshevskiy, line 77: There is some confusion here. 

This code is duplicated between this overload and the next overload.
Should be unified for sure.

include/eve/detail/function/simd/common/load.hpp
```diff
@@ -164,105 +68,189 @@ namespace eve::detail
...
-  template<typename T, typename N, data_source Ptr>
-  EVE_FORCEINLINE wide<T, N> load_( EVE_SUPPORTS(cpu_), ignore_none_ const&, safe_type const&
-                                  , eve::as<wide<T, N>> const &tgt, Ptr ptr
-                                  ) noexcept
-  requires(simd_compatible_ptr<Ptr,wide<T, N>> && std::same_as<abi_t<T, N>, aggregated_>)
-  {
-    return aggregate_load(tgt,ptr);
+    if constexpr (!std::is_pointer_v<DS>)
+    {
+      constexpr bool is_aligned_enough = c_t() * sizeof(e_t) >= DS::alignment();
```
DenisYaroshevskiy, line 108: I'm so confused to what;'s happening.

DS is data_source. We use that in the API of load. Here we call ::alignmetn on it. We obv can't do that for pointers. What's happening.

include/eve/module/core/regular/load.hpp
jfalcou: It is this way cause of our adage of having the basics of wide, etc, not depending on the module/core. When everything settles on this front, we can revisit that. Also, the doc should be here.

include/eve/detail/function/load.hpp
SadiinsoSnowfall: Fixed

include/eve/detail/function/simd/common/load.hpp
jfalcou: It was. We can add an assert in load(it,it) I think.

include/eve/detail/function/simd/common/load.hpp
SadiinsoSnowfall: Fixed

PR #2115, opened 2025-06-10
Remove PCH CMake prologue so warning can get through in CI all the time

jfalcou:


cmake/toolchain/gcc.x86.cmake
```diff
@@ -3,7 +3,7 @@
 ##  Copyright : EVE Project Contributors
 ##  SPDX-License-Identifier: BSL-1.0
 ##==================================================================================================
-set(CMAKE_C_COMPILER    gcc  )
-set(CMAKE_CXX_COMPILER  g++  )
+set(CMAKE_C_COMPILER    gcc-14  )
+set(CMAKE_CXX_COMPILER  g++-14  )
```
DenisYaroshevskiy, line 7: Are we shipping many gccs?

include/eve/arch/cpu/top_bits.hpp
```diff
@@ -263,9 +263,9 @@ namespace detail
     {
       if constexpr ( !is_aggregated )
       {
-        storage_type bit_mask = detail::set_lower_n_bits<storage_type>(bits_per_element) << (i * bits_per_element);
-        if ( x ) storage |= bit_mask;
-        else     storage &= ~bit_mask;
+        storage_type masked_bits = detail::set_lower_n_bits<storage_type>(bits_per_element) << (i * bits_per_element);
```
DenisYaroshevskiy, line 266: ith_element_mask

include/eve/arch/riscv/top_bits.hpp
```diff
@@ -98,8 +98,8 @@ struct top_bits<Logical>
     {
       // if static_bit_size < 8, we have some dirty bits in resulted unsigned,
       // so cleanning them.
-      u_type mask = detail::set_lower_n_bits<u_type>(static_bits_size);
-      return raw_value & mask;
+      u_type mask = raw_value & detail::set_lower_n_bits<u_type>(static_bits_size);
```
DenisYaroshevskiy, line 101: Is it about a cast? What's this change?

DenisYaroshevskiy (commented):
WHOAH! YOU ACTUALLY DID IT!
This is massive!

Couple of nits

include/eve/arch/riscv/top_bits.hpp
jfalcou: The old code returned an int because `&` promotes. It made a warning because we compared the int to an unsigned in the test. As we kinda want this function to return an unsigned, I moved the part aroudn to prevent the promotion to get out

cmake/toolchain/gcc.x86.cmake
jfalcou: I think I only have 14 in my current CI image, and it is installed this way. I can later fix the image to use the default GCC selector indeed.

DenisYaroshevskiy (approved):

PR #2134, opened 2025-07-27
shuffle v2: free masking

DenisYaroshevskiy:
Making masking generic. Now for avx512/sve/rvv we just generate a shuffle + blend with 0 and rely on the compiler to make blend with 0 to go away. 

The driver test could've been better but let's rely on the integration testing to extra verify this one later.
I only removed P::has_zeroes from avx512 and not everywhere as well.

The plan is, after this, to start white box  testing individual patterns in l2/l3.

### DIFF
--- a/include/eve/detail/shuffle_v2/simd/common/shuffle_v2_driver.hpp
+++ b/include/eve/detail/shuffle_v2/simd/common/shuffle_v2_driver.hpp
@@ -10,8 +10,6 @@
 #include <eve/detail/shuffle_v2/simplify_plain_shuffle.hpp>
 #include <eve/module/core/regular/if_else.hpp>
 
-#include <iostream>
-
 namespace eve::detail
 {
 
--- a/include/eve/detail/shuffle_v2/simd/common/shuffle_v2_driver.hpp
+++ b/include/eve/detail/shuffle_v2/simd/common/shuffle_v2_driver.hpp
@@ -8,6 +8,9 @@
 #pragma once
 
 #include <eve/detail/shuffle_v2/simplify_plain_shuffle.hpp>
+#include <eve/module/core/regular/if_else.hpp>
+
+#include <iostream>
 
 namespace eve::detail
 {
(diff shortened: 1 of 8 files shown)


include/eve/detail/shuffle_v2/simd/common/shuffle_v2_driver.hpp
```diff
@@ -8,6 +8,9 @@
 #pragma once
 
 #include <eve/detail/shuffle_v2/simplify_plain_shuffle.hpp>
+#include <eve/module/core/regular/if_else.hpp>
+
+#include <iostream>
```
jfalcou, line 13: spurrious `<iostream>`

include/eve/detail/shuffle_v2/simd/x86/shuffle_l2.hpp
```diff
@@ -257,6 +228,7 @@ shuffle_l2_x86_u64x2(P p, fixed<G> g, wide<T, N> x)
   {
     constexpr int mm = idxm::x86_mm_shuffle_4(P::idxs);
 
+    // TODO(2135): this is accidentally important.
```
jfalcou, line 231: Can you give some context here?

include/eve/detail/shuffle_v2/simd/x86/shuffle_l2.hpp
```diff
@@ -285,37 +257,14 @@ shuffle_l2_alignr_epi32_self(P p, fixed<G> g, wide<T, N> x)
     if constexpr( P::reg_size == 32 )
     {
       if constexpr( P::g_size >= 8 ) return _mm256_alignr_epi64(x, x, shift_epi64);
-      else return _mm256_alignr_epi64(x, x, shift_epi32);
+      else return _mm256_alignr_epi32(x, x, shift_epi32);
```
jfalcou, line 260: Why the change from 64 to 32?

test/unit/api/regular/shuffle_v2/shuffle_v2_driver.cpp
```diff
@@ -312,6 +311,124 @@ TTS_CASE_TPL("arm-v7, emulate double", tts::types<double>)
...
+  }
+  else
+  {
+    constexpr bool free_masking = eve::current_api >= eve::avx512 || eve::current_api >= eve::sve
+                                  || eve::current_api >= eve::rvv;
+
+    const T               arithmetic_in {[](int i, int) { return i + 1; }};
+    const eve::logical<T> logical_in([](int i, int) { return i % 3 == 1; });
+
+    (void)free_masking;
```
jfalcou, line 331: I would prefer we use `[[maybe_unused]]`.

jfalcou (approved):

PR #2141, opened 2025-08-11
Add variadic version of combine

SadiinsoSnowfall:


include/eve/concept/combinable.hpp
```diff
@@ -0,0 +1,29 @@
...
+
+#include <eve/concept/vectorized.hpp>
+
+#include <type_traits>
+#include <concepts>
+#include <bit>
+
+namespace eve
+{
+  template <typename W0, typename W1, typename... Ws>
```
jfalcou, line 18: this would be better with some documentation :D

include/eve/module/core/regular/impl/simd_cast.hpp
```diff
@@ -34,7 +34,7 @@ simd_cast_(EVE_REQUIRES(cpu_), O const&, T x, as<Target> tgt) noexcept
 
     return res;
   }
-  else if constexpr( sizeof(T) < sizeof(Target) ) { return simd_cast(eve::combine(x, T{}), tgt); }
+  else if constexpr( sizeof(T) < sizeof(Target) ) { return simd_cast(combine(current_api, x, T{}), tgt); }
```
jfalcou, line 37: why ?

test/unit/api/regular/combine.cpp
```diff
@@ -0,0 +1,53 @@
...
+  SPDX-License-Identifier: BSL-1.0
+**/
+//==================================================================================================
+#include "test.hpp"
+
+
+TTS_CASE_TPL("combine", eve::test::scalar::all_types)
+<typename T>(tts::type<T>)
+{
+  constexpr std::size_t max_size_bit = 7; // 128 bytes max
```
jfalcou, line 14: drop a small comment to explain the logic of the test
Also, should we also do that with logical ?

jfalcou (changes requested):

include/eve/module/core/regular/impl/simd_cast.hpp
SadiinsoSnowfall: Some impls included the `detail/function` header but used the callable version defined in `core/regular` (because the former included the later, which is not the case anymore).
Do you think it would be better to use the callable directly when possible ?

test/unit/api/regular/combine.cpp
SadiinsoSnowfall: should be better now

PR #2155, opened 2025-09-01
Fix some translatable edge cases

SadiinsoSnowfall:
Fix the following translatable edge cases :
- fix `combine`: didn't work well with translatable values in some cases
- fix `slice`: `std::array<T, N>` is now translatable by default to `std::array<U, N>` if `T` is translatable into `U`.
- fix `translate_ptr`: no longer discard qualifiers in some cases. Added a new `copy_qualifiers_t` helper type that will also be used for the undergoing fp16 work (specifically when translating `[qals] _Float16*` to `[quals] __fp16*` on ARM.

\+ added some tests

### DIFF
--- a/include/eve/detail/ptr.hpp
+++ b/include/eve/detail/ptr.hpp
@@ -1,0 +1,27 @@
+//==================================================================================================
+/*
+  EVE - Expressive Vector Engine
+  Copyright : EVE Project Contributors
+  SPDX-License-Identifier: BSL-1.0
+*/
+//==================================================================================================
+#pragma once
+
+#include <type_traits>
+
+namespace eve::detail
+{
+  template<typename Src, typename Tgt>
+  struct copy_qualifiers
+  {
+  private:
+      using R = std::remove_pointer_t<Tgt>;
+      using U1 = std::conditional_t<std::is_const<R>::value, std::add_const_t<Src>, Src>;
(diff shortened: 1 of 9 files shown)


include/eve/detail/ptr.hpp
```diff
@@ -0,0 +1,27 @@
...
+  private:
+      using R = std::remove_pointer_t<Tgt>;
+      using U1 = std::conditional_t<std::is_const<R>::value, std::add_const_t<Src>, Src>;
+      using U2 = std::conditional_t<std::is_volatile<R>::value, std::add_volatile_t<U1>, U1>;
+  public:
+      using type = U2;
+  };
+
+  template<typename Src, typename Tgt>
+  using copy_qualifiers_t = typename copy_qualifiers<Src, Tgt>::type;
```
DenisYaroshevskiy, line 26: this should be somewhere in traits.
@jfalcou 

Also - we don't seem to have a better implementation of conditional than std, we should!

DenisYaroshevskiy (commented):
I was just passing by, it looks fine to me but up to @jfalcou

include/eve/detail/ptr.hpp
SadiinsoSnowfall: Done

jfalcou (approved):

PR #2182, opened 2025-09-19
Allow callables to advertise a neutral element used in masked reductions

SadiinsoSnowfall:


SadiinsoSnowfall:
missclicked on the AI BS button...

PR #2207, opened 2025-10-27
Disallow fixed<0>

SadiinsoSnowfall:
`fixed<0>` is currently allowed to be constructed and used. This yields to undefined and [observably compiler-dependant behavior](https://godbolt.org/z/PzeqzPqeq).


include/eve/module/core/regular/sort.hpp
```diff
@@ -131,7 +129,7 @@ namespace eve
     constexpr EVE_FORCEINLINE
     T make_bitonic(T x, Less less, fixed<G>) noexcept
     {
-      if constexpr (G < 1)
+      if constexpr (G == 1)
```
DenisYaroshevskiy: Up to you but maybe just replace fixed with index? I think it's cuter

PR #2210, opened 2025-10-31
add average

jtlap:


jtlap:
i added the simplification of all fonctions using widen as an option, but this only works with g++ and clang seems to still have  a problem with upgrade_if_t

PR #2216, opened 2025-11-02
[FP16] min/max/minmax/abs/is_nan base support

SadiinsoSnowfall:


include/eve/arch/cpu/wide.hpp
```diff
@@ -175,7 +175,7 @@ namespace eve
 
     //! Constructs a eve::wide by splatting a scalar value in all lanes
     template<typename S>
-    requires std::constructible_from<Type,S>
+    requires (std::constructible_from<Type, S> || (std::floating_point<S> && std::same_as<Type, eve::float16_t>))
```
jfalcou: we need some eve::constructible_from then.

PR #2222, opened 2025-11-05
[FP16] absmin/absmax support

SadiinsoSnowfall:


include/eve/arch/float16.hpp
```diff
@@ -272,6 +272,34 @@ namespace eve
...
     };
+
+    template <typename T>
+    constexpr EVE_FORCEINLINE float16_t operator+(T const& lhs, float16_t const& rhs) noexcept
+      requires (std::integral<T> || std::floating_point<T>)
+    {
+      return float16_t{ static_cast<float>(lhs) + static_cast<float>(rhs) };
+    }
+
+    template <typename T>
```
jfalcou: whay are not these operators symmetric ?

PR #2246, opened 2025-11-15
Feat/fp16/fma family

jtlap:
lower upper is not tested yet on fp16

### DIFF
--- a/test/unit/module/core/fma.cpp
+++ b/test/unit/module/core/fma.cpp
@@ -85,7 +85,7 @@ TTS_CASE_WITH("Check behavior of fma[promote] on all types",
               eve::test::simd::ieee_reals,
               tts::generate(tts::randoms(-1000, 1000),
                             tts::randoms(-1000, 1000),
-                            tts::randoms(-1000, 1000)) //wf16 TODO
+                            tts::randoms(-1000, 1000))
              )
   <typename T>(T const& a0, T const& a1, T const& a2 )
 {


include/eve/module/core/constant/ieee_constant.hpp
```diff
@@ -70,6 +70,7 @@ namespace eve
 
   namespace _
   {
+    struct real16 { constexpr real16(auto v) : value(v) {} eve::float16_t value; };
```
jfalcou: Why changing this file here while there is another branch for that ?

test/unit/module/core/fsm.cpp
```diff
@@ -86,7 +86,7 @@ TTS_CASE_WITH("Check behavior of fsm lower upper on all types",
                             tts::randoms(-1000, 1000),
                             tts::randoms(-1000, 1000))
              )
-  <typename T>(T const& a0, T const& a1, T const& a2 )
+  <typename T>(T const& a0, T const& a1, T const& a2 ) //TO DO wf16
```
jfalcou: why those comments ?

test/unit/module/core/fms.cpp
```diff
@@ -86,7 +86,7 @@ TTS_CASE_WITH("Check behavior of fms[promote] on all types",
                             tts::randoms(-1000, 1000),
                             tts::randoms(-1000, 1000))
              )
-  <typename T>(T const& a0, T const& a1, T const& a2 )
+  <typename T>(T const& a0, T const& a1, T const& a2 )  //TO wf16
```
jfalcou: why the comment again

jfalcou (changes requested):

include/eve/module/core/constant/ieee_constant.hpp
jtlap: sans ça j'avais des pb de compil

PR #2259, opened 2025-11-30
Shuffle l3.cpp

DenisYaroshevskiy:
x86 and current arm neon should be tested.
A lot of this is LLM assisted - I did make it add the logs and see that the tests actually call things.

Before submitting I rereviewed again and did the logging validation too.


DenisYaroshevskiy (commented):
It should be vqtbx1q_u8 instead of vqtbx1_u8 seems like

Issue #2103, opened 2025-05-21
[BUG] requirements for comparisons


DenisYaroshevskiy:
currently it's same_lane_or_scalar. That's not enough, we should be better at requiring what we need.

PR #2284, opened 2026-01-07
fixing requirements for comparisons operators

Nxirda:
This PR adresses #2103

### DIFF
--- a/include/eve/module/core/regular/is_greater.hpp
+++ b/include/eve/module/core/regular/is_greater.hpp
@@ -21,7 +21,7 @@ namespace eve
   {
     template<arithmetic_value T, arithmetic_value U>
     constexpr EVE_FORCEINLINE common_logical_t<T, U> operator()(T a, U b) const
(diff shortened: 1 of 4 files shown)


include/eve/module/core/regular/is_greater.hpp
```diff
@@ -20,7 +21,7 @@ namespace eve
   {
     template<arithmetic_value T, arithmetic_value U>
     constexpr EVE_FORCEINLINE common_logical_t<T, U> operator()(T a, U b) const
-      requires (compatible_arithmetic_values<T, U>)
+      requires ( compatible_arithmetic_values<T, U> && eve::greater_comparable<element_type_t<T>,element_type_t<U>> )
```
jfalcou: Some comparators use `compatible_arithmetic_values` , some use `eve::same_lanes_or_scalar<T, U> `. We need to investigate which is requried actually

jfalcou (changes requested):

PR #2323, opened 2026-03-29
Add support for compilation on Windows with MSVC

Nxirda:
MSVC has a hard time doing EBO due to ABI compatibility, simply putting the empty base class first in the wide inheritance tree makes it correct. There is a proper explanation [there](https://learn.microsoft.com/en-us/cpp/cpp/empty-bases?view=msvc-170).

An other issue that will need to be tackled is the difference between Iterators, Ranges and raw pointers. Some of the code assumes Iterators to be contiguous iterators when they may not. 

An other point is that MSVC iterators are not convertible to pointers meaning code such as `auto x = std::array{1,2,3}; auto* it = x.begin()` will not work. Either the `.data()` can be used otherwise it will need proper handling.

There seems to be some issues related to conversions, at this stage, related MSVC warnings are disabled. `-Wconversion` should perhaps be considered ?

Left to do are `algorithm`, `memory` and `examples/doc` targets.

There seems to be issues with ClangCL. Errors are mainly coming from callables using the old interface, it will be easier to tackle down when does use the new system.


DenisYaroshevskiy:
Cool! Are you trying like all of it at once? Maybe some targets separately first?

Nxirda:
> Cool! Are you trying like all of it at once? Maybe some targets separately first?

For now, I’m trying to ensure that every target, apart from the algorithms, can at least compile, as this shouldn’t require too many changes to the code. (All of them having very similar issues)

DenisYaroshevskiy:
> > Cool! Are you trying like all of it at once? Maybe some targets separately first?
> 
> For now, I’m trying to ensure that every target, apart from the algorithms, can at least compile, as this shouldn’t require too many changes to the code. (All of them having very similar issues)

it used to not compile much more severely.
Why not algorithms? They are extra bad?

Nxirda:
> > > Cool! Are you trying like all of it at once? Maybe some targets separately first?
> > 
> > 
> > For now, I’m trying to ensure that every target, apart from the algorithms, can at least compile, as this shouldn’t require too many changes to the code. (All of them having very similar issues)
> 
> it used to not compile much more severely. Why not algorithms? They are extra bad?

>... MSVC iterators are not convertible to pointers meaning code such as auto x = std::array{1,2,3}; auto* it = x.begin() will not compile. 

This is the main issue with the algorithms as far as i could tell. Some of them are fines but the ones that are not passing will require either to handle iterators separately from the rest or move everything to use `.data()` and `.size()`. So not to sure on how to address it at this point.

DenisYaroshevskiy:
> Will require either to handle iterators separately from the rest or move everything to use .data() and .size(). So not to sure on how to address it at this point.

I'm not 100% on what you mean here. Algorithms should work on iterators fine, if that's what you mean.

I'm absolutely not trying to steer you to do algorithms in the same pr though, it's big enough

include/eve/arch/x86/predef.hpp
```diff
@@ -41,7 +41,7 @@
         &&  defined(__AVX512DQ__) && defined(__AVX512VL__)  \
         &&  defined(__AVX512DQ__) && defined(__AVX512VL__)  )
 #    undef SPY_SIMD_IS_X86_AVX512
-#    undef SPY_SIMD_DETECTED ::spy::_::simd_version::avx512_
```
jfalcou, line 44: why ?

jfalcou (approved):

DenisYaroshevskiy:
Single threaded compilation sounds awful. Why do you need this? Are you changing only msvc

Nxirda:
> Single threaded compilation sounds awful. Why do you need this? Are you changing only msvc

There is an issue with `bit_ternary` on MSVC that is causing the compiler to run out of heap space. This is only a test to see if reducing parallelism fixes the issue. If it does I'll handle the case properly with a dedicated step. This will not affect other targets.

.github/workflows/unit.yml
```diff
@@ -375,7 +375,7 @@ jobs:
           system_profiler SPHardwareDataType | grep -E "(Memory|Cores)"
           df -h /System/Volumes/Data
       - name: Fetch current branch
-        uses: actions/checkout@v6
+        uses: actions/checkout@v4
```
jfalcou: fix

jfalcou (changes requested):

PR #2329, opened 2026-04-09
experiment: replace eve::detail with eve::_

SadiinsoSnowfall:


DenisYaroshevskiy:
_ is a reserved symbol in 26

jfalcou:
It's context-dependent and only affects the variable side of the language.

https://godbolt.org/z/xrf9Yxh94

Kumi et al. uses that for a long time and have no issue in C++26

DenisYaroshevskiy:
I know that it works. I'm just ify about it. It might affect parsing speed maybe.
I dunno, I'd use smth else but I don't have strong feelings about it.

SadiinsoSnowfall:
`$` is available if we want our code to look like old-style JS.

jfalcou:
> I know that it works. I'm just ify about it. It might affect parsing speed maybe. I dunno, I'd use smth else but I don't have strong feelings about it.

Smart ISO people told me this one use case was OK. So ...
_ or $, yoru choice :D

DenisYaroshevskiy:
Ok - let it be _ if you like it. I'd maybe go with one letter but this is fine

jfalcou:
<joke>
Let's have `namespace eve::λ`
</joke>

PR #2330, opened 2026-04-09
Fix find_broadcast_group where Size < Cardinal

SadiinsoSnowfall:


jfalcou (approved):

jfalcou (changes requested):
hmm, care to see if the buggy pattern is actually tested ?

include/eve/module/core/regular/impl/broadcast_group.hpp
```diff
@@ -25,7 +25,7 @@ find_broadcast_group(std::array<std::ptrdiff_t, Size> const& pattern)
     std::ptrdiff_t group = -1, index = -1;
   };
 
-  for( std::ptrdiff_t g = Cardinal; g > 0; g /= 2 )
+  for( std::ptrdiff_t g = Size; g > 0; g /= 2 )
```
DenisYaroshevskiy, line 28: The tests are here https://github.com/jfalcou/eve/blob/981e4f3d7ec335618161d400f48faa4ad7676840/test/unit/internals/optimize_pattern.cpp#L21

DenisYaroshevskiy:
What bothers me - is that there is a generic fallback that should always work. And it didn't. There is a more fundamental problem I think.

SadiinsoSnowfall:
The problem isn't with the generic fallback, it was simply `find_broadcast_group` returning invalid broadcast group shuffle.

jfalcou (approved):

PR #2339, opened 2026-04-27
Prevint / nextint

SadiinsoSnowfall:


jfalcou (approved):
Maybe have slightly more explanation in the docs

Issue #2023, opened 2024-11-19
[FEATURE] eve::algo::count_if


DenisYaroshevskiy:
We don't have the algorithm, seems like

PR #2341, opened 2026-05-12
Feat/count_if

su-mt:
[Issue](https://github.com/jfalcou/eve/issues/2023)

### DIFF
--- a/include/eve/module/algo.hpp
+++ b/include/eve/module/algo.hpp
@@ -28,6 +28,7 @@
 #include <eve/module/algo/algo/container/soa_vector.hpp>
 #include <eve/module/algo/algo/copy.hpp>
 #include <eve/module/algo/algo/copy_if.hpp>
+#include <eve/module/algo/algo/count_if.hpp>
 #include <eve/module/algo/algo/equal.hpp>
 #include <eve/module/algo/algo/fill.hpp>
 #include <eve/module/algo/algo/find_last.hpp>
--- a/include/eve/module/algo/algo/count_if.hpp
+++ b/include/eve/module/algo/algo/count_if.hpp
@@ -1,0 +1,115 @@
+//==================================================================================================
+/*
+  EVE - Expressive Vector Engine
+  Copyright : EVE Project Contributors
+  SPDX-License-Identifier: BSL-1.0
+*/
+//=================
+
+#pragma once
+
+#include <eve/module/algo/algo/common_forceinline_lambdas.hpp>
+#include <eve/module/algo/algo/traits.hpp>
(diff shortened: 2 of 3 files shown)


su-mt:
Hi! Thanks for the cool library. I tried implementing count_if and would be happy to contribute it

DenisYaroshevskiy:
Lovely, let me have a look!

DenisYaroshevskiy:
Thank you for you time!
This is not an amazing count if, unfortunately, one can do a much better count if.
Let me explain.

eve::count_true is not a terribly fast function. It has to do like a register wide reduction, which is meh.

The trick to a good count_if is knowing that on most platforms, true is -1 and false is 0.

Here is a limited implementation with transform reduce:

```
std::ptrdiff_t good_but_limited_count_bigger_than(std::span<const int> s,
                                                  int x) {
    return -eve::algo::transform_reduce(
        s, [&](eve::wide<int> xs) { return (xs > x).bits(); }, 0);
}
```
https://godbolt.org/z/bd8vP9vY4

## Why limited? 

Well - there is an issue of count overflowing, especially for small types, like chars or shorts.
We have one algorithm that does something like this and that's `min_element_1_pass_` and it's not an easy one

https://github.com/jfalcou/eve/blob/26046f89204e4aa791a1a6230dfb6336f466f671/include/eve/module/algo/algo/min_element.hpp#L18

The secret sauce there is `for_each_iteration_with_fixed_overflow_` https://github.com/jfalcou/eve/blob/26046f89204e4aa791a1a6230dfb6336f466f671/include/eve/module/algo/algo/for_each_iteration_fixed_overflow.hpp#L165
That has a callback for every time an overflow is happening.

## What to do?

If it's more than you wanted - that's completely understandable, we appreciate you taking your time.
If you still want to do this - go for it. 

If you want to start smaller to get here - try implementing `max_element_1_pass_` 
It will require less innovation also still - quite hard
https://github.com/jfalcou/eve/blob/26046f89204e4aa791a1a6230dfb6336f466f671/include/eve/module/algo/algo/max_element.hpp#L14

You can also look at some other missing algorithms, but I suspect will still require some serious thinking.

Finally we have some "good first issue" tasks if you want 
https://github.com/jfalcou/eve/issues?q=is%3Aissue%20state%3Aopen%20label%3A%22good%20first%20issue%22

NOTE: please carefully review the LLM code, there are some sloppy things in your current pr.

su-mt:
Thank you very much for the detailed explanation.

I did not initially realize that the problem was this involved.

I will take a closer look at `min_element_1_pass_`  and the “good first issue” tasks to better understand the codebase and the overall approach.

Thank you again for your time and guidance.

PR #2345, opened 2026-05-28
Experiment: switch to using NTTP for wide sizes

SadiinsoSnowfall:


include/eve/detail/function/simd/common/movemask.hpp
```diff
@@ -17,18 +17,18 @@ namespace eve::_
...
 
-  template<typename T, typename N>
+  template<typename T, size_type N>
   EVE_FORCEINLINE auto movemask( eve::logical<eve::wide<T, N>> const& v ) noexcept
   {
     static_assert ( !has_aggregated_abi_v<wide<T, N>>, "movemask is an internal function that does not make sense for aggregated" );
 
     std::uint64_t res = 0;
-    for (std::uint64_t i = 0; i != (std::uint64_t)N(); ++i) {
+    for (std::uint64_t i = 0; i != (std::uint64_t) N; ++i) {
```
jfalcou: `static_cast` maybe?

jfalcou (changes requested):

jfalcou:
we are blocked because we choose size as the new name and we can't have size_t as a traits
Can we say count is the proper name ? and we end up with count_t ?

Nxirda:
> we are blocked because we choose size as the new name and we can't have size_t as a traits Can we say count is the proper name ? and we end up with count_t ?

FWIW the standard now simply calls it `width`. The width of a simd register reads naturally, I think it is reasonable to follow the standard. your call.

jfalcou:
Sounds good to me

Nxirda:
This PR introduces a few breaking changes :

- cardinal* have been renamed to width*, traits and overall naming are impacted.
- eve::wide now takes an NTTP instead of a fixed<N> to describe the register width

On a sidenote :

- The overall style is not homogeneous, we sometimes use `T::size()` directly which we should not.
- The usage of `lanes` vs `fixed` in function seems quite unclear and pretty arbitrary, that might be an easy clean up angle.

include/eve/arch/cpu/as_register.hpp
```diff
@@ -111,29 +110,38 @@ namespace eve
       template<typename Func>
       EVE_FORCEINLINE decltype(auto) apply(Func f) const  { return kumi::apply(f,*this); }
     };
+
+    template<typename T>
+    inline constexpr bool is_blob_impl = false;
```
Nxirda, line 115: We have a few of these new traits that will need to be refactored properly. With the new `wide` construction it is harder to use the old `is_instance_of` trait as we're mixing NTTPs and Types in template signatures.

jfalcou (approved):

PR #2355, opened 2026-06-13
Fix #2353 - Improve popcount code generation on X86

### LINKED ISSUE #2353: Population count in a loop
First of all, thanks for this library!

I'm playing around with translating this function to EVE:

```c++
template <size_t K>
auto count(const std::array<uint64_t, K>& words_) -> size_t {
    size_t c = 0;
    for (size_t i = 0; i < words_.size(); ++i)
        c += static_cast<size_t>(std::popcount(words_[i]));
    return c;
}
```

and checking on godbolt the generated code: https://godbolt.org/z/xsf3GvPxz As you can see `K` is pretty small (4), the compiler manages to generate vectorized code already. 

The EVE-based code surprised me though and it's entirely possible that what I wrote is extremely bad! The compiler generates many, many more instructions. A cursory benchmark also shows that the EVE-based version is quite a lot worse. 

Again, I'm pretty sure the code I wrote with EVE is likely not the correct way to use the library, but I wanted to bring this to your attention.

jfalcou:
As per #2353 bug report, this improves X86 popcoutn codegen : 
  + Use a LUT based approach for pre-AVX512 implementation
  + Redierct to AVX512 popcnt intrinsic whenever possible
  + Use slice/combine trick for case like AVX or pre BITALG/PCNT AVX512

### DIFF
--- a/include/eve/module/core/regular/impl/simd/x86/popcount.hpp
+++ b/include/eve/module/core/regular/impl/simd/x86/popcount.hpp
@@ -15,7 +15,7 @@
 
 namespace eve::_
 {
-  template<unsigned_scalar_value T, typename N, callable_options O>
+ template<unsigned_scalar_value T, typename N, callable_options O>
   EVE_FORCEINLINE auto popcount_(EVE_REQUIRES(sse2_), O const& o, wide<T, N> x) noexcept
   requires std::same_as<abi_t<T, N>, x86_128_>
   {
@@ -37,8 +37,12 @@ namespace eve::_
 
       __m128i count16 = _mm_add_epi16(count8, _mm_srli_epi16(count8, 8));
 
-      if constexpr (sizeof(T) == 2) return wide<T,N>{count16};
-      if constexpr (sizeof(T) == 4) return wide<T,N>{_mm_add_epi32(count16, _mm_srli_epi32(count16, 16))};
+      if constexpr (sizeof(T) == 2) return wide<T,N>{_mm_and_si128(count16, _mm_set1_epi16(0x00FF))};
+      if constexpr (sizeof(T) == 4)
+      {
+        __m128i count32 = _mm_add_epi32(count16, _mm_srli_epi32(count16, 16));
+        return wide<T,N>{_mm_and_si128(count32, _mm_set1_epi32(0x000000FF))};
+      }
     }
     else
     {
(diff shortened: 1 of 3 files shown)


DenisYaroshevskiy:
Ooh. Fun. So - pop count chars and then sum together?

jfalcou:
You lookup 4 bits into their numbers of bits, same for hi bits then sum properly depending on size. We use the funky add 8-> 16 or 8->64 to complète larger types.

DenisYaroshevskiy:
Gotcha yeah.

I recommend the following:

1. Do blocks: popcount u8, sum to u16, sum to u32 so we do not have to recurse. It's not too bad to recurse but I think it'd be cute.
2. The hsum thing: this is not the best hsum code. Remember we used to fo the same in reduce with a plus.
We need some details helper function

jfalcou:
> Gotcha yeah.
> 
> I recommend the following:
> 
> 1. Do blocks: popcount u8, sum to u16, sum to u32 so we do not have to recurse. It's not too bad to recurse but I think it'd be cute

Yes will do

> 2. The hsum thing: this is not the best hsum code. Remember we used to fo the same in reduce with a plus.

There is no hsum. Popcpount is not a reduction.

jfalcou:
Current performance is 

`u8` : 3 cycles
`u16`: 6 cycles
`u32`: 9 cycles
`u64`: 4 cycles

As per LLVM-MCA : https://godbolt.org/z/17fahWWrT

include/eve/module/core/regular/impl/simd/x86/popcount.hpp
```diff
@@ -12,78 +12,90 @@
...
 
-    xx -= bit_shr(xx, 1) & pattern_2bit; // put
-    xx  = (xx & pattern_4bit) + (bit_shr(xx, 2) & pattern_4bit);
-    xx  = (xx + bit_shr(xx, 4)) & pattern_16bit; // put count of each 8 bits into those 8 bits
+      if constexpr (sizeof(T) == 1) return wide<T,N>{count8};
+      if constexpr (sizeof(T) == 8) return wide<T,N>{_mm_sad_epu8(count8, _mm_setzero_si128())};
 
-    return bit_cast(xx,as(in));
-  };
+      __m128i count16 = _mm_add_epi16(count8, _mm_srli_epi16(count8, 8));
```
DenisYaroshevskiy: can we have a comment about why this works. I'm surpised the top part doesn't need a mask

jfalcou: Because it does ! I fixed that

DenisYaroshevskiy (approved):

PR #2363, opened 2026-07-14
Add compile-time subscript indices

### LINKED ISSUE #2362: [FEATURE] index_t based insert and extract functions
Currently the `insert` and `extract` function, defined in `detail/function/subscript.hpp` both take a runtime index value as a parameter. It may be possible to provide better implementation on some architectures if the index is known at compile time (taking example on the bit shift family of functions).

Proposal : 
Implement new overloads of `insert` and `extract` that take an `index_t` as their index type.

aouxwoux:
Closes #2362.

Adds `index_t` overloads to the subscript dispatch points and forwards them through `wide::get`/`set` and `logical<wide>::get`/`set`. This keeps the runtime-index overloads unchanged while allowing architecture-specific implementations to recognize a compile-time lane index.

The added API test covers numeric and logical wides at lanes 0 and 1.

Validation:
- `git diff --check`

I could not run the unit binary locally because this host has no C++ compiler or CMake installation. The PR unit workflow builds `unit.api` across the supported toolchains.

### DIFF
--- a/include/eve/arch/cpu/logical_wide.hpp
+++ b/include/eve/arch/cpu/logical_wide.hpp
@@ -391,8 +391,20 @@ namespace eve
       _::insert(*this, i, v);
     }
 
+    template<std::ptrdiff_t I>
+    EVE_FORCEINLINE void set(index_t<I> i, std::convertible_to<bool> auto v) noexcept
+    {
+      _::insert(*this, i, v);
+    }
+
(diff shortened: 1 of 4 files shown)


DenisYaroshevskiy:
Hi

Creating pull requests with ai while not putting effort into them results in nonsense prs like this that waste maintainers time.

Please reconsider doing this.

PR #2364, opened 2026-07-14
Feature : Make algorithms compile on MSVC

Nxirda:
This PR aims at finalizing EVE support of MSVC.

- Main changes are in tests, MSVC iterators are not convertible to pointers -> rely on `.data()` and `.size()` instead.

- some file names have been changed : MSVC imposes an internal limit on the number of characters per file path. This limit is reached when building the tests targets.

- For_each_selected requires some changes due to MSVC STL iterators beeing checked iterators, thus the arithmetic issued asserts


test/unit/module/algo/views/backward_eve_iterator.cpp
```diff
@@ -65,9 +65,12 @@ TTS_CASE_TPL("Check backward_iterator", algo_test::selected_types)
       algo_test::writeable_readable_iterator(eve::views::backward(u_l), values, replace);
     }
   };
-
-  run_test(data.begin(), data.end());
-  run_test(data.cbegin(), data.cend());
+  auto *f = data.data();
+  auto *l = f + data.size();
+  auto const* cf = f;
+  auto const* cl = l;
```
DenisYaroshevskiy, line 71: const auto

jfalcou: we're an East Coast workshop, man ;)

DenisYaroshevskiy (commented):
the algo part is fine, dunno about the rest

test/unit/module/algo/views/backward_eve_iterator.cpp
DenisYaroshevskiy: We are? oh, I didn't know.

jfalcou:
Jtlap approved th emath, so we're good

PR #2365, opened 2026-08-02
add compile-time indexes to insert and extract

### LINKED ISSUE #2362: [FEATURE] index_t based insert and extract functions

aouxwoux:
## what changed

insert and extract now accept eve::index_t for compile-time lane indexes. the common implementation forwards to the existing runtime path for now, while leaving room for architecture-specific implementations.

## why

a compile-time lane index gives the implementation enough information to specialize this operation on architectures where that helps.

## checks

- git diff --check passes.
- the local machine does not have cmake or a c++ compiler installed, so repository ci will do the native validation.

fixes #2362

### DIFF
--- a/include/eve/detail/function/simd/common/subscript.hpp
+++ b/include/eve/detail/function/simd/common/subscript.hpp
@@ -49,6 +49,12 @@ namespace eve::_
     }
   }
 
+  template<callable_options O, typename Wide, std::ptrdiff_t I>
+  EVE_FORCEINLINE auto extract_(EVE_REQUIRES(cpu_), O const&, Wide p, index_t<I>) noexcept
+  {
+    return extract(p, static_cast<std::size_t>(I));
+  }
+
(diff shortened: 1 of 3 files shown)


jfalcou:
* This doesn't fix the whole issue: no use of constant-based intrinsic in any architecture
* this a fucking copypasta from some random ass LLM.

You spent 0 effort on this PR, I am giving 0 effort to it.

Don't post AI SLOP again or you're banned.

jfalcou:
OH SORRRY
you already did that 3 weeks ago.
Ban hammer time

PR #2373, opened 2026-08-27
Use forceinlining wisely

### LINKED ISSUE #1501: [FEATURE] Consider flatten instead of aggressive force_inline
As explained here: https://awesomekling.github.io/Smarter-C++-inlining-with-attribute-flatten/

I wonder how it would be better than our FORCE INLINE ALL THE THING strategy as of now.

Nxirda:
Change the implementation to rely on flattening attribute instead of agressive forceinline based on kumi experiments.
This should fix #1501.

TLDR : 
EVE_FORCEINLINE -> EVE_ABI, now extends to 
- `[[gnu::flatten, gnu::always_inline, gnu::artificial]]` on clang and gcc 
- `[[msvc::flatten, msvc::forceinline]]`

Functions in namespace eve::_ should not be marked with `EVE_ABI`.


DenisYaroshevskiy:
Explain about ::_ please. Those functions really ought to be force inlined

DenisYaroshevskiy:
It is also important for eve functions themselves to be inlined. I don't understand this change. Flatten is unlikely to be a good substitute.

Like algorithms for example, they are giant but they should be inlined, because the user provides the body to be inlined to.

Nxirda:
> Explain about ::_ please. Those functions really ought to be force inlined

The reason for this is that the functions in the public interface are not in the ‘detail’ namespace and must in any case be forced to be inline; we want to ensure this.
 
On the other hand, for internal functions, it is not necessary to force them to be inline in every case; we simply want them to be inline within the body of the function that uses them. This is what ‘flatten’ is for: we can use it to recursively inline calls to internal functions without having to mark each internal function as ‘forceinline’.

As tested in kumi this eases out pression on the compiler in the codegen phase.

DenisYaroshevskiy:
Ok, let's see, if the codegen comes bad, we'll adjust

jfalcou:
we need one target without `NO_FORCE_INLINE` now

PR #2382, opened 2026-08-30
Add a CONTRIBUTING.md

jfalcou:
EVE had no contribution guide, so nothing stated that a callable spans seven files, that an example lives under a CMake glob, or what a documentation block is expected to contain. This adds one, deliberately linking setup.md, dev_cmake.md and dev_environment.md rather than restating them, so the toolchain and option tables keep a single home.

Two sections are policy rather than mechanics. The first says we will not ask whether a patch came out of a language model — we have used one here ourselves — but that a pull request its author cannot defend, or a feature whose tests would pass against any implementation, gets closed regardless of how it was written. The second addresses the people arriving with a course or hackathon quota to fill: welcome, but volume is what a small team cannot absorb, and the good first issue label is a better use of the same fortnight.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,330 @@
+# Contributing to EVE
+
+Thank you for your interest in EVE. Bug reports, documentation fixes, a missing overload on one
+architecture and whole new functions are all welcome.
+
+By opening a pull request you agree to license your contribution under the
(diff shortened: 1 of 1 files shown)


jfalcou:
@DenisYaroshevskiy @jtlap @SadiinsoSnowfall @Nxirda Your opions are welcome about this one

PR #2383, opened 2026-08-30
Bring the Doxygen build down to zero warnings

jfalcou:
The documentation build emitted 107 warnings and a 7135-line log; it now emits none, so a new warning becomes a signal rather than noise.

Most of it is mechanical: unterminated LaTeX delimiters that silently turned the rest of a block into code, comment markers glued to the text they introduced, concepts documented with the wrong structural command, and the leftovers of an old search-and-replace that had eaten a few identifiers. The thirty-two unresolved references to product-type concepts were a different matter — they belong to kumi, so the fix is to point them there and consume kumi's tagfile, downloaded at configure time the way KYOSU already does.

Two things beyond the warnings. The Godbolt button was quietly broken for any example containing a unicode identifier, because btoa cannot encode beyond Latin-1; the state is now encoded through UTF-8 first. And the Doxygen group ids carry an eve_ prefix, which matters now that tagfiles are shared between projects and groups merge by id.

The decorators, ieee_constant and float16_t are documented along the way, the last of them through the EVE_DOXYGEN_INVOKED pattern so the two-branch definition stays readable.


jfalcou:
bonus: we CHECK the warnings stays low

PR #2388, opened 2026-09-01
Update kumi and other dependencies

Nxirda:
- In eve kumi standalone was named `tuple.hpp` which does not make sens anymore, the standalone is now `kumi.hpp`

- Spy was not handled like other dependancies

- Dependency handling was done in detail and deps. Now everything is centralized into deps.


include/eve/module/bessel/detail/kernel_bessel_ij_small.hpp
Nxirda, line 1: Well spurious includes here

include/eve/arch/ppc/tags.hpp
```diff
@@ -7,7 +7,7 @@
 //==================================================================================================
 #pragma once
 
-#include "eve/detail/spy.hpp"
+#include "eve/deps/spy.hpp"
```
jfalcou: no " in EVE include, this is an old mistake, coudl you fix it too everywhere it spawns

PR #2392, opened 2026-09-02
Remove types instantiations and refactor traits

Nxirda:
This PR aims at starting the clean up of type instantiation where it proves to be unnecessary

- Some traits should be inline variables used for detection purpose

- Gather all cardinal traits together on the model of abi

- Gather decorators to the right place and not in details

- fix tests accordingly

- down do not require a struct by itself, same for up (the details one)

- struct traits are now inline variables

- using aliases for return type deduction instead of a struct


Nxirda:
This basically depends on #2374 to have the proper infra. 
Once (if) it's merged I'll just rebase upon it.

Issue #2390, opened 2026-09-02
[FEATURE] Measure per-TU compile time and peak memory, and give the template cost a target


jfalcou:
**Is your feature request related to a problem? Please describe.**

Compile-time cost is a known concern — #2368 says the kumi update nearly doubled the compile memory
of `eve::algo::soa_vector`, and #1601 tracks debug size — but neither carries a number, because
nothing measures. Every discussion about compile cost currently rests on impressions.

Measurements taken by hand today, building the unit suite one file at a time with the flags the CI
actually uses (`-O0 -Wno-psabi -DEVE_NO_FORCEINLINE`, since the toolchain sets no `-O` and
`CMAKE_BUILD_TYPE` is never defined):

| file | peak RSS | CPU |
|---|---|---|
| `unit/module/core/add.cpp` | 2.9 GB | 42 s |
| `unit/module/core/mul.cpp` | 2.4 GB | 35 s |
| `unit/module/math/geommean.cpp` | 1.5 GB | 18 s |
| most files | 0.5 – 1.2 GB | 5 – 15 s |

**Describe the solution you'd like**

Two tiers, because the cheap measurement and the detailed one have different costs and different
audiences.

### Tier 1 — per-translation-unit time and memory, on every PR

`clang++ -fproc-stat-report=<file>` appends one line per invocation with the wall time, user time and
**peak memory of that process**. Per-process, so `-j 16` does not distort it. No JSON, negligible
overhead, one flag.

The job produces a single artifact: a table of every test target with its time and its peak RSS,
sortable by either column, **and a delta against a baseline**. The delta is the part that makes this
worth building — a sorted list of what is expensive today gets read twice and then ignored, whereas
*this file gained 400 MB since main* is actionable. It is also exactly what #2368 lacks.

### Tier 2 — template-level breakdown, local only

`-ftime-trace` writes a JSON per TU with a per-phase and per-template-instantiation breakdown, and
[ClangBuildAnalyzer](https://github.com/aras-p/ClangBuildAnalyzer) aggregates those into a ranking of
the costliest templates and functions. That ranking only means something over the whole corpus: run
on a handful of TUs it is noise.

This stays out of CI. `-ftime-trace` slows compilation down and writes about 1 MB per TU — some
700 MB over the suite, roughly 100 MB zipped, which would count against the account's artifact quota
on every run. It belongs on the machine of whoever is investigating.

A CMake target makes it one command:

```cmake
[...]

Issue #2368, opened 2026-08-13
[BUG] Compile-time memory of `eve::algo::soa_vector` nearly doubled by #2357 (kumi update)


matthijs:
**Describe the bug**
Since `5b0a96ef3` ("Update kumi to latest (#2357)", 2026-06-19), compiling a single
translation unit that instantiates `eve::algo::soa_vector<T>` for a 10-member
`eve::struct_support` product type went from **1.33 GB → 2.29 GB peak RSS (+72%)** and
**37.9s → 53.9s** wall.

In our project this is not marginal: five such instantiation units went from ~2.5–3.0 GB
to **6.5–9.7 GB** each, which takes a `-j4` build from 12 GB to 32 GB of peak concurrent
demand and OOM-kills our CI runner (32 GB, no swap). The largest single TU now peaks at
**9.68 GB**.

The product-type/`wide` layer is unaffected — this is specific to `soa_vector`.

**To Reproduce**
Self-contained, no project dependencies:

```cpp
#include <cstdint>

#include <eve/module/algo.hpp>
#include <eve/module/core.hpp>
#include <eve/traits/product_type.hpp>

struct State : eve::struct_support<State,
    double,  double,  double,  double,  double,
    float,   float,   float,
    int32_t, int32_t
>
{
  using eve_disable_ordering = void;
};

using vec_t = eve::algo::soa_vector<State>;

void exercise(vec_t& v)
{
  v.push_back(State{});
  State s = v.get(0);
  v.set(0, s);
  v.erase(v.begin());
}

std::size_t size_of(vec_t const& v) { return v.size(); }
```

Build:

```sh
clang++-22 -std=c++23 -fproc-stat-report=stats.csv -I<eve>/include -c repro.cpp -o /dev/null
```

## Bisect

`5b0a96ef3` is the first bad commit; `4f9ac7c58` is its direct parent. The step is clean —
no intermediate values across the probed range:

| Commit | Peak RSS |
|---|---:|
| `4f9ac7c58` (parent) | 1 392 432 KB (1.33 GB) |
| **`5b0a96ef3`** | **2 395 892 KB (2.29 GB)** |
| `3eb49dc75` (main, 2026-08-12) | 2 396 200 KB (2.29 GB) |

Three GOOD probes landed within 0.3 MB of each other, two BAD probes within 0.5 MB.

The commit changes 2 866 lines, essentially all in `include/eve/deps/kumi/tuple.hpp`
(206 646 → 227 292 bytes).

[...]

Issue #1601, opened 2023-05-12
[BUG] Debug size for eve functions is quite bad


DenisYaroshevskiy:
Here is `eve::any_of` being 5000 instructions in Debug: https://godbolt.org/z/e4TYv7E4e
Would be nice to see where we get the worse offenders and fix them if possible.

PR #2393, opened 2026-09-02
Fix #2390 - Measure what a translation unit costs to compile

jfalcou:
EVE's compilation cost is a tracked problem (#2368, #1601) that nothing currently measures. This adds two levels of it. `-fproc-stat-report` records peak resident memory and CPU time for every translation unit of the unit test suite, per process, so the figures do not move with `-j`. A local `time-trace` target drives ClangBuildAnalyzer over `-ftime-trace` output when the breakdown by template is wanted. Both are off by default and add nothing to an ordinary build.

The CI side runs on LEGEND at `-j 16` once the unit tests have released the runner. A push to main leaves a reference artifact, a pull request downloads it and reports the delta in its job summary. The measurement happens inside the pinned image so that two runs weeks apart stay comparable: the machine's own clang moves with its packages, and a delta read across an upgrade describes the compiler instead of the code.


Nxirda:
One point : do we want the last artifact built on main compared to the current PR or a plot for the longer run ?
We probably want the latest to see how much we diverge in the long run.
FWIW, ACTS is using CI to compare execution times of algorithms and they have a bot that when asked uploads a PDF of the plot to ensure nothing broke. Thus it is not triggered every time a PR is launched.
We should somehow try to do this for the comp-time target. If a PR is currently beeing worked, on the CI machine would not blocked.

On a sidenote : as per local experiment compile time is sensible to the stability of the machine, are we able to guarantee it ?

jfalcou:
The reference is already the latest main: a pull request looks up the last successful run of this workflow on `main` and reads its artifact, and `main` refreshes it on every push. The long run is the part that is missing, and the CSVs are kept for it, 90 days from `main` and 14 from a pull request. Worth an issue on copacabana: either the reader page gains a history view, or a second page draws one, since the machinery is shared by the whole fleet rather than specific to eve.

Stability is not guaranteed, which is why the report leans on peak memory. Two runs of the same code repeated the peak to the mebibyte and moved CPU time by 19 %. Both sides are measured in the same image, so a compiler upgrade cannot pass for a code change, and the measurement is per translation unit, so it does not depend on `-j`. It runs on a GitHub-hosted runner, never on the self-hosted machine, so it blocks nobody; `workflow_dispatch` is one line if you also want to ask for a run by hand.

Nxirda:
> The reference is already the latest main: a pull request looks up the last successful run of this workflow on `main` and reads its artifact, and `main` refreshes it on every push. The long run is the part that is missing, and the CSVs are kept for it, 90 days from `main` and 14 from a pull request. Worth an issue on copacabana: either the reader page gains a history view, or a second page draws one, since the machinery is shared by the whole fleet rather than specific to eve.
> 
> Stability is not guaranteed, which is why the report leans on peak memory. Two runs of the same code repeated the peak to the mebibyte and moved CPU time by 19 %. Both sides are measured in the same image, so a compiler upgrade cannot pass for a code change, and the measurement is per translation unit, so it does not depend on `-j`. It runs on a GitHub-hosted runner, never on the self-hosted machine, so it blocks nobody; `workflow_dispatch` is one line if you also want to ask for a run by hand.

- Sounds good to me, we would greatly benefit from the history view (just like the tests we were doing locally with @SadiinsoSnowfall script), the exact design is up to you tho.

- As per memory consumption that's a great news, CPU time moving by 19% is really high tho.  Then we might want to rely more (maybe a later step) onto metrics offered by ClangBuildAnalyzer like number of template instantiated and so on. This would add meaning to the time, for two runs that diverge a lot, if the numbers are the same we know it comes from the machine being unstable ?

DenisYaroshevskiy:
If you manage to substantially improve the compilation speed, it'll make for a good post. Or i'd like to know how at least

Issue #2397, opened 2026-09-07
[BUG] gcc16 with C++26 gives a compilation error when using eve::wide in swap_adjacent.hpp


Roy-974:
**Describe the bug**
Same issue as #2317 but with gcc16 (tested 16.1 and 16.2 on godbolt)
Compiler error is:
```swap_adjacent.hpp:101:41: error: no matching function for call to 'add_shuffle_levels(<brace-enclosed initializer list>)' [-Wtemplate-body]```
See https://godbolt.org/z/z7nqKv9br

**To Reproduce**
Including wide.hpp and compiling.

**Expected behavior**
Normal compilation

**Setup:**
Godbolt

**Additional context**

DenisYaroshevskiy:
Thank you for reporting this.

 It's span constructible from {} issue.
Should be an easy fix, although we are missing the CI.
Maybe gcc doesn't support it yet in 16 or smth.

Roy-974:
Hello,

No problem! Glad to help.

From https://en.cppreference.com/cpp/container/span/span there is a workaround for initializer list:
```return _::idxm::add_shuffle_levels({{half_l, half_l, 4}});```
does seem to work.

pipatron:
Today I tried eve for the first time and got this error message, but with clang++ 22.1.8

Building with `-std=c++23` removes the error, but I really want to use C++26 together with eve.. Is there a fix to this? I'm afraid I don't understand the discussion here about spans or so, is there anything a lowly user of the library can do?

````
$ git clone https://github.com/jfalcou/eve.git
Cloning into 'eve'...
remote: Enumerating objects: 203066, done.
remote: Counting objects: 100% (1740/1740), done.
remote: Compressing objects: 100% (831/831), done.
remote: Total 203066 (delta 1157), reused 984 (delta 902), pack-reused 201326 (from 3)
Receiving objects: 100% (203066/203066), 70.39 MiB | 20.68 MiB/s, done.
Resolving deltas: 100% (160247/160247), done.
````
````
$ cat wide.cpp 
#include <eve/wide.hpp>
int main() { return 0; }
````
````
$ clang++ --version
clang version 22.1.8 (Fedora 22.1.8-4.fc44)
Target: x86_64-redhat-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
Configuration file: /etc/clang/x86_64-redhat-linux-gnu-clang++.cfg
````
````
$ clang++ -isystem eve/include -std=c++26 wide.cpp
[...]

PR #2401, opened 2026-09-08
Fix #2397 - Name the array type at the add_shuffle_levels call site

jfalcou:
A braced list at the `add_shuffle_levels` call site resolves differently on every standard library: libstdc++ 15 sees both the `span` and the `std::array` overload and calls it ambiguous, libstdc++ 14 and 16 find neither viable. A named `std::array` is an exact match in all three, leaves the C++26 guard added for #2317 untouched, and is already how `blend.hpp` writes the same call.

The preemptive matrix gains gcc 14, whose libstdc++ has no `span` constructor from an `initializer_list`. That is the regime gcc 16 reports and the one gcc 15 hides, so the job covers it without changing the image.

### DIFF
--- a/include/eve/detail/shuffle_v2/idxm.hpp
+++ b/include/eve/detail/shuffle_v2/idxm.hpp
@@ -1283,7 +1283,10 @@ add_shuffle_levels(std::span<const std::ptrdiff_t> ls)
 constexpr auto
 add_shuffle_levels(std::integral auto a, std::integral auto... as)
 {
-  return add_shuffle_levels(std::array {static_cast<std::ptrdiff_t>(a), static_cast<std::ptrdiff_t>(as)...});
+  return add_shuffle_levels ( std::array{ static_cast<std::ptrdiff_t>(a)
+                                        , static_cast<std::ptrdiff_t>(as)...
+                                        }
+                            );
 }
 
 template<std::ptrdiff_t... ls>
--- a/include/eve/detail/shuffle_v2/idxm.hpp
+++ b/include/eve/detail/shuffle_v2/idxm.hpp
@@ -1280,14 +1280,11 @@ add_shuffle_levels(std::span<const std::ptrdiff_t> ls)
   return base + use_masks;
 }
 
-// In C++26 the {} can be used with spans, so it's not needed anymore
-#if __cplusplus <= 202302L
 constexpr auto
-add_shuffle_levels(std::array<std::ptrdiff_t, 3> ls)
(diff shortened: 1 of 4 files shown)


include/eve/module/core/named_shuffles/swap_adjacent.hpp
```diff
@@ -98,7 +98,7 @@ struct swap_adjacent_t
       std::ptrdiff_t half_l = level(eve::as<half_t> {}, g);
       // since we are adding, we need to deal with aggregation
       if( reg_size > 32 ) return half_l;
-      return _::idxm::add_shuffle_levels({half_l, half_l, 4});
+      return _::idxm::add_shuffle_levels(std::array<std::ptrdiff_t, 3> {half_l, half_l, 4});
```
DenisYaroshevskiy: I don't kike this. how did we solve it in other places? I bet we didn't specify the size

DenisYaroshevskiy: eugh https://github.com/jfalcou/eve/commit/66040ceeab9f5df52567ae6d475c0f85f7341877#diff-a9bc4319f65d472b6cb39966afe0a1293d8f8ae2b73faabbdf47a9302964908e

let's at least ctad? And remove the `if cpp version`?

jfalcou: the ifdef can go yes, and I'll try CTAD

include/eve/detail/shuffle_v2/idxm.hpp
```diff
@@ -1280,14 +1280,11 @@ add_shuffle_levels(std::span<const std::ptrdiff_t> ls)
...
 
-// In C++26 the {} can be used with spans, so it's not needed anymore
-#if __cplusplus <= 202302L
 constexpr auto
-add_shuffle_levels(std::array<std::ptrdiff_t, 3> ls)
+add_shuffle_levels(std::ptrdiff_t a, std::ptrdiff_t b, std::ptrdiff_t c)
 {
-  return add_shuffle_levels(std::span(ls));
+  return add_shuffle_levels(std::array {a, b, c});
 }
```
DenisYaroshevskiy, line 1290: Why not variadic?

DenisYaroshevskiy (approved):
variadic would be nice.

include/eve/detail/shuffle_v2/idxm.hpp
jfalcou: Didn't thought of it, will fix

jfalcou: Done

Issue #2404, opened 2026-09-12
[BUG] Rendering bug in AlgoTraits documentation on https://jfalcou.github.io/eve/group__eve__algo__traits.html


pipatron:
I tried to read the documentation on https://jfalcou.github.io/eve/index.html

After scratching my head for a while I realized some pages must be buggy. This is one example that doesn't make sense, I've tested with Firefox and even got desperate and tried in Edge(!) on my work laptop: https://jfalcou.github.io/eve/group__eve__algo__traits.html

Surely there must be text missing from this screenshot?

<img width="1168" height="1053" alt="Image" src="https://github.com/user-attachments/assets/255d670f-ac80-45da-a6bc-81a93b3b672f" />

I didn't find a specific project for the documentation website so I assume it is generated from the main project somehow, that's why I add an issue here.

jfalcou:
Some pages are not fully written yet. The algo parts are currently being worked on; I expect to push a bunch next week.

The traits documentation in the code is there and readable, however.

jfalcou:
OK, this is a doxygen bug on our side, going to patch this real quick

DenisYaroshevskiy:
If you have a specific problem you are trying to solve, I can perhaps answer.

PR #2405, opened 2026-09-13
Fix #2404 - Rewrite the algorithm options documentation

jfalcou:
The Options page lost its text to a CSS rule that hid the first paragraph of every entry once the brief moved, and its options shared one group, with no page of their own. The rule is removed, and every option is now a doxygen group with the eve_ prefix, which gives it a stable page and a tag file entry that cannot collide with another library's; the options split in three themes, the trait sets move to the developer page, and the module page and its three sub-pages get a description.


jfalcou:
The rest of the Algo doc will be refurbished the same way, then the main documentaiton too.

include/eve/module/algo/algo/concepts.hpp
```diff
@@ -11,16 +11,37 @@
 //! @addtogroup eve_algorithms
 //! @{
 //!   @defgroup eve_algos Algorithms
-//!   @brief Provides SIMD implementation of standard algorithms
+//!   @brief SIMD implementations of the standard algorithms.
+//!
+//!   Each algorithm keeps the name and the meaning of its `std` counterpart, takes ranges, and
```
DenisYaroshevskiy, line 16: did you ai the comments by any chance? This statement is incorrect. I'm pretty sure whole text is off

include/eve/module/algo/algo/set_intersection.hpp
```diff
@@ -16,7 +16,10 @@ namespace eve::algo
 
 //================================================================================================
 //! @addtogroup eve_algos
-//! return type from set_intersection
+//! @{
+//!   @struct set_intersection_result
+//!   @brief Return type of eve::algo::set_intersection: where each input stopped, and the output.
```
DenisYaroshevskiy, line 21: also not true

include/eve/module/algo.hpp
```diff
@@ -11,10 +11,23 @@
 //! @addtogroup eve_simd
 //! @{
 //!   @defgroup eve_algorithms Algorithms and Views
-//!   @brief Algorithms and Views
+//!   @brief SIMD versions of the standard range-based algorithms.
 //!
-//!   This module provides implementation of SIMD versions of some std ranges algorithms.
-//!   Take care that some prerequisite and syntaxes are not fully standard conformant
```
DenisYaroshevskiy, line 17: I don't like this text either

